PR #108, opened 2018-10-25
Textmaps crash fix

SaschaKP:
- we can't trust the textmaps, not every landtile is stretchable, so we create a stretchables hashset beforehand, and use that hashset to check if we can use the textures or not.

- solved a bug where if at end of tiledata, it could cause a crash

### DIFF
--- a/Game/Map/Tile.cs
+++ b/Game/Map/Tile.cs
@@ -33,6 +33,7 @@ namespace ClassicUO.Game.Map
 {
     public sealed class Tile : GameObject
     {
+        public static readonly HashSet<ushort> _Stretchables = new HashSet<ushort>();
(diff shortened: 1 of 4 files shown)


andreakarasho:
I know that each textmaps texture is stretchable. Can you explain better?

SaschaKP:
Not every landtile has a texture map, see image below:

image
this is the landtile for a paver, now let's search that in textures:
image
It isn't there, now, when I reach in the game a zone where there is that tile, without my fix, the game will crash

SaschaKP:
the image is misleading, anyway, the textures are for 0x420, here is the image of something similiar, but again, we can't trust the texturemaps to always be present:
image

image

PR #171, opened 2018-11-21
Feature/status gump outlands

### LINKED ISSUE #162: Outlands Custom Status Gump
**Description**
The Outlands player status gump needs to be rendered slightly differently than the regular player status gump.  For example, murder count and various timers are displayed rather than swing speed, resists, etc.

**Additional context**
Shard: Outlands

floyd501:
This patch adds the Outlands-style character status gump as an option to display in lieu of the classic and modern status gumps.  A configuration setting has been added (`status_gump_style`), which replaces the `use_old_status` setting, and allows the user to select between three options: "classic", "modern", and "outlands".  Architecturally, the previous StatusGump class has been separated into a hierarchy to support these different styles, as the rendering is quite different between each of the them.  The existing logic in StatusGump for both the "classic" and "modern" styles has not changed, only separated into different subclasses of StatusGumpBase.  The configuration setting determines the type to instantiate when the player opens the detailed character status, either through the paper doll or by double-clicking the mini status (hits/mana/stam) gump.

Outstanding issues:
- The packet format containing some information that is displayed on the Outlands-style status gump, such as murder count decay and criminal timer vs. resist values, is temporarily being reinterpreted from other player values that occupy the same portion of the packet.
- Beyond an alpha or beta, this may not make much sense as a configuration option for the end user.  Ideally, the client itself would know which gump style to use based on some communication with the server.
- Having shard-specific code in the core client may not be ideal, whether its server communication or gump rendering.  This implementation can be reworked if a different direction is desired.

Resolves #162


andreakarasho:
Implemented here: https://github.com/andreakarasho/ClassicUO/commit/2777c41151685888137824625b7dada9baa27a40

Thanks!

PR #175, opened 2018-11-26
Default settings refactor

Mentos-:
Set proper defaults in Settings class. Added functionality to GameLoop so that if a settings.json is not detected on launch, a new one with proper default values will be saved and notepad.exe will be launched to edit it


andreakarasho (approved):
Some settings still don't be implemented. Anyway thanks :)

PR #179, opened 2018-11-27
Added Error Code message popup

### LINKED ISSUE #172: Message for character already into the world
When you try to login with a different character, you should receive a message like "Another character is already logged ..... ".
At the moment ClassicUO remains stucked into the "Entering into the world" Scene.

It's for you :D @rdegelo

rdegelo:
Closes #172


SaschaKP (approved):
ok

PR #191, opened 2019-01-06
Minimum drag distance

jkachhad:
Requires a minimum distance that the mouse move to consider a drag. Resolves an issue where players may be clicking rapidly (mount/dismount), and a single pixel move during LButton Down could be considered a drag.

Currently set to 3 pixels.


jkachhad:
A discussion point prior to merge: why was DragBegin raised AFTER MouseDrag? Shouldn't it be raised prior to any dragging logic? I've changed it to this behavior in this commit and did not notice any input issues.

SaschaKP (changes requested):
please, whenever you want to do patches or submit modifications, at the moment you should refer to the "dev" branch.

Master is used only for official submissions, the dev branch is ATM the active development branch with the latest modifications inside.

jkachhad:
Rebased into ```andreakarasho:dev``` branch as requested.

SaschaKP (approved):
tested

Issue #185, opened 2018-12-02
CharacterCreation city choise gump


andreakarasho:
This gump is necessary because some server uses different city index.
@rdegelo

PR #192, opened 2019-01-07
Character Creation Starting Cities

jkachhad:
Implemented selection of a starting city. Closes https://github.com/andreakarasho/ClassicUO/issues/185.

Mirrored as close as possible to the original client. A feature I was unable to recreate is highlighting of both the selection button and text when hovering over _the text_. Both will highlight if hovering over the button.

**ClassicUO**
image

7.0.20.0
image


andreakarasho (commented):
First of all: thanks!

dont use Gump class as inheritance class if the object that you are rappresenting is a child of a Gump. Instead use the Control class.
We need to fix it, but the change is minimal.

PR #193, opened 2019-01-07
Fix drag distance

jkachhad:
The previous PR request (https://github.com/andreakarasho/ClassicUO/pull/191) did not work as intended on the dev-branch after rebase. Reverted the prior PR.

This is a much simpler fix. You can test by dragging on your character for 5 pixels (watch the debug gui mouse position!).

### DIFF
--- a/src/Game/UI/Controls/Control.cs
+++ b/src/Game/UI/Controls/Control.cs
@@ -47,6 +47,8 @@ namespace ClassicUO.Game.UI.Controls
 
     internal abstract class Control : IDrawableUI, IUpdateable, IColorable
     {
+	    private static int _minimumDistanceForDrag = 5;
+
         private static SpriteTexture _debugTexture;
         private readonly List<Control> _children;
         private bool _acceptKeyboardInput, _acceptMouseInput, _mouseIsDown;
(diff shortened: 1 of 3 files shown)


andreakarasho:
Have you checked the UIManager/DoDragControl method? 
I cannot try because im out home, but it handles the min pixel dragging distance (is 1 atm)

jkachhad:
As I can see it, that method continues to move the control after it's started to drag, I think you meant AttemptDragControl.

```
 gump.InvokeMouseOver(position);

if (_mouseDownControls[0] == gump)
     AttemptDragControl(gump, position);
```

```
1. InvokeMouseOver -> OnMouseOver -> (solution) -> InvokeDragBegin -> DragBegin.Raise -> OnMouseDragBegin
2. AttemptDragControl -> (solution) -> InvokeDragBegin -> DragBegin.Raise -> OnMouseDragBegin
```
Both paths lead to invoking Drag.Begin and conflict with each other.  Placing the solution in the second path only prevents existing gump controls from moving with a minimum distance.

The issue is that ```GameSceneInputHandler.OnMouseDragBegin()``` is invoked on path 1 and creates the ```HealthBarGump```:
```
Engine.UI.Add(currentHealthBarGump = new HealthBarGump(mobile) { X = Mouse.Position.X - (rect.Width >> 1), Y = Mouse.Position.Y - (rect.Height >> 1) });
Engine.UI.AttemptDragControl(currentHealthBarGump, Mouse.Position, false);
```
prior to any checks from ```AttemptDragControl``` on path 2.

Hope this clarifies why the HealthBarGump popping up with minimal drag. Let me know if you have other places where we can add a fix.

Maybe both paths need a minimum distance check :)

PR #195, opened 2019-01-07
UpdateCharacterList/ReceiveCharacterList

jkachhad:
There is a distinction between UpdateCharacterList (Characters) and ReceiveCharacterList (Characters, Cities, CharacterListFlags) packets. After #192, there is additional data that is expected on the packet, resulting in a crash from deleting a character.

I'm not sure if I did the CharacterSelectionGump updating (disposing and re-instantiation) in the best manner. One of the other contributors can comment.

### DIFF
--- a/src/Game/Scenes/LoginScene.cs
+++ b/src/Game/Scenes/LoginScene.cs
@@ -394,8 +394,20 @@ namespace ClassicUO.Game.Scenes
 
                     break;
                 case 0x86: // UpdateCharacterList
-                case 0xA9: // ReceiveCharacterList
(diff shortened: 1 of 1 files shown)


andreakarasho:
It's ever better to dispose/initialize background gump in the same way i did inside the loginscene/update method, cause it avoids black wholes (
Frame - dispose gump - frame - remove disposed gump - frame (black whole) - add new gump).
But idk if it happens it this case.

SaschaKP:
we could anyway test it, and if there is any problem, place a fix.

Issue #154, opened 2018-11-12
Tracking Arrow Gump does not dismiss


jaedan:
The tracking arrow gump does not dismiss when right clicked.

Use tracking on Outlands to track someone. An arrow will appear. Attempt to dismiss the arrow with right click.

andreakarasho:
here works!

PR #197, opened 2019-01-09
Quest Arrows, Skill Buttons

jkachhad:
Implements quest arrows (resolves https://github.com/andreakarasho/ClassicUO/issues/154). The implementation differs from the original clients, which I don't believe account for map Z values.

Additionally fixes invalid SkillEntry reads, the action should be read prior to skill name.


andreakarasho (approved):
Well done

PR #215, opened 2019-01-11
The topbar, when collapsed, will display a background

markdwags:
This is to mimic the behavior in the standard client

image


andreakarasho:
Tried the same thing, but was not enought to fix cause remains a transparent background clickable (you can see it if you enable the Debugging mode).

i've done the right fix  :D 

image

PR #224, opened 2019-01-19
Login Scane (login&passwd field focus), TextBox SetNextFocus

plugtopus:
**SetNextFocus** method to set next TextBox focus elem

3zfwfir

### DIFF
--- a/src/Game/UI/Controls/TextBox.cs
+++ b/src/Game/UI/Controls/TextBox.cs
@@ -32,6 +32,8 @@ namespace ClassicUO.Game.UI.Controls
     {
         public TextEntry TxEntry { get; private set; }
 
+        public TextBox SetNextFocus { get; set; }
+
         public TextBox(TextEntry txentry, bool editable)
         {
             TxEntry = txentry;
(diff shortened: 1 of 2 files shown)


andreakarasho (commented):
I think that is not the best and beautiful way to do it.
You should use parent.findcontrol<T>() and search for another control that accepts the input keyboard, then set focus to it

plugtopus:
I thought so too, but I decided that for the two fields it would be too complicated code. 
Ok, I'll fix it soon.

plugtopus:
I created **NextElement()** to search for the next TextBox item. 

> case SDL.SDL_Keycode.SDLK_TAB:
> NextElement().SetKeyboardFocus();
> break;

andreakarasho:
Trust me : 
https://github.com/3HMonkey/JuicyUO/blob/e9f26d450598ee080af98dc2ceff9be574ce4a94/JuicyUO/Core/UI/AControl.cs#L874

plugtopus:
Make a similar implementation or you duplicate it yourself?

andreakarasho:
similar implementation !
Thanks for your work :)

plugtopus:
Np :)

andreakarasho:
how is going on?

plugtopus:
Temporarily switched to another task, there is absolutely no time to return.
I can close "Pull request" and transfer to "Issues".

andreakarasho:
ok i do it :)

PR #228, opened 2019-01-22
Save & Restore "scala"

plugtopus:
- Add to Debug Gump
- ScaleZoom to settings.json
- Save & Restore with relogin
- Move def value Apply() => Default()

3zjlvmd


andreakarasho:
I haven't added it because in a lowend pc a low scale can decrease performance and cause app blocking.
A restart of ClassicUO can resolve the issue and the user will never set the scale anymore :D.

plugtopus:
But this is solved by simply deleting the "Profiles" folder. I think  user himself is to blame if he sets a scale that his computer cannot process) On my server I play 0.9 to improve the FPS. My friend uses 1.2 constantly. I think the possibility of saving is a must.

I can add checkbox "reset the scale when entering the game" by default. It will be better this way?

andreakarasho:
Yes add the checkbox please :)

plugtopus:
3zk8mdz

PR #230, opened 2019-01-22
TopbarGumpIsDisabled

plugtopus:


andreakarasho:
why topbar should be disabled ?

plugtopus:
I have many reasons:
- Many shard not support this menu (sphere server). Example, "Chat" & "Help" buttons
- Many servers love Ultima Renaissance skin without this menu. This is the same as the opportunity return old user status bar. 
- This function is in Orion
- This function is in Classic Ultama Online client (>AOS?)

Original client (5x,7x): 
3zkcky0

andreakarasho:
perfect ;)

PR #237, opened 2019-01-28
Feature/char creation outlands

floyd501:
This PR addresses character creation for Outlands clients connecting to the Outlands server.

The Outlands server still sends the original list of cities (Yew, Britain, etc.) as part of the ReceiveCharacterList (0xA9) packet.  The current behavior of `CreateCharCityGump` is to display that list of cities received over the map image using the names, positions, and descriptions sent from the server.  This doesn't quite work for Outlands because the cities (1) the cities aren't the same, and (2) they are static and already drawn onto the map itself.

The solution here essentially ignores the city info sent from the server to avoid overlaying that data on top of the Outlands map that already conveys the same information.  This may be improved in a couple of ways, especially if the Outlands server changes its behavior at some point, but it works for now.

Tested on the Outlands server and a local version of the regular ServUO server.


jkachhad:
How does it display when you connect with the original client? There shouldn't be a need to make an exception for a specific shard, otherwise we'll end up make switches almost everywhere. 

What is the OSI/Broadsword client version you are connecting with? Does the server use modified gumps.mul?

floyd501:
The original client used to connect to Outlands is modified (including Gumpart.mul) and based on 7.0.15.1, I believe.  It displays the city selection screen in the same way that's described here; namely, the city info list as provided from the server is ignored since the map gump is completely static and already contains the city locations.  Also, there's no actual city selection available to the user, not even the ability to click cities to view their descriptions.  The goal of this PR is to make that interaction the same.  An alternative could be to just skip the screen altogether (I believe the [Orion client](https://github.com/jaedan/OrionUO) does this), but that might negatively impact the player's experience.

Agree, the extra checks for "shard type" are not ideal and probably could be improved.  There was some initial discussion about this a while ago, and the consensus was to at least not encourage separate forks of this repo for shard-specific functionality.  Instead, it's currently a configuration option.  I think everyone is open to other ideas.

SaschaKP:
the gump is not shown if not present on the original client, and also the section where there is that gump is ignored and skipped completely.

So the correct behaviour is: if the image is present on the file gump.mul, then show the section, otherwise ignore it and skip it.

For example, if you go with fiddler to position 5593 on GUMPS, there isn't one on your side, as also there isn't on my side on my shard.

So, the correct pull request is to just skip and send the default map and city to the server (britain - felucca)

plugtopus:
I think a bad idea adapts to a particular shard (Which has its own client and modifications). 
If no gump with cities, it’s a good idea to skip this step.

SaschaKP:
Yes, @plugtopus, the client behaviour is that, so we should continue that way...

floyd501:
Sorry, I'm a little confused.  There is a gump with cities, but the cities are in the image itself, unlike other city selection map gumps, where the city info is overlaid.

The changes in this PR will display this during character creation (depending on `ShardType`):

create_char_outlands_cuo

Do we not want to do that?  I don't mind skipping the step, but someone went to the trouble to create the map.

SaschaKP:
In fact I misunderstood, but anyway, I believe that the server should send the cityinfo, this is something to look accurately at, if the server is sending the cityinfo, position and name on the gumps...

This is what is sent on characterlist

m_Stream.Write((byte)info.Length);

			for (int i = 0; i < info.Length; ++i)
			{
				CityInfo ci = info[i];

				m_Stream.Write((byte)i);
				m_Stream.WriteAsciiFixed(ci.City, 32);
				m_Stream.WriteAsciiFixed(ci.Building, 32);
				m_Stream.Write(ci.X);
				m_Stream.Write(ci.Y);
				m_Stream.Write(ci.Z);
				m_Stream.Write(ci.Map.MapID);
				m_Stream.Write(ci.Description);
				m_Stream.Write(0);
			}

All the data is already there, what is really needed is to adopt and calculate buttons and positioning based on that data, and also, considering that description is a cliloc, you could make all the necessary things that way.

I'll submit a patch for the no gump, direct creation, but for this, I believe we can work out something more flexible, non unilateral

floyd501:
No worries, thanks for the feedback.  I agree, the server should send the correct cities and probably would be nice to have a map gump that doesn't have the cities embedded in the image.  Unfortunately, I don't have control over that.

To be clear, `CreateCharCityGump` already handles the case of regular servers that send the correct `CityInfo` (see #192); this was just a slight modification of that for Outlands, but I understand the concern about these kinds of checks for a specific shard.

jaedan:
I'll take a look and see if I can make outlands work in a more standard way.

floyd501:
Thanks, @jaedan.  If that happens, then everything @jkachhad already added should work.

I'll close this for now.  Thank you everyone for the feedback.

jaedan:
If I can't make it more standard, then I'll apply this patch to the outlands fork I'll be maintaining in the future. I want the delta to be as small as possible, but I can carry patches like this.

andreakarasho:
When we'll have a launcher, the user can choise if it's outlands or not.

@floyd501 if you have splitted into 2 gumps by the "shard_type" setting props and it doesn't change the standard game flow i'll accept your PR .

jkachhad:
Anyways, for a comprimise on all the shards that have different city selection screens, here's how we should do it:
A) Server sends normal info -> Use normal info.
B) Server does not send any cities -> Skip the window and send creation packet
C) Server does not wish to show cities, but still display window. Server should send a single city with location 0, 0  ->  We ignore all cities with Point2D.Zero when creating map

floyd501:
@andreakarasho Yeah, nothing in this PR changes the behavior of regular shards.  That all works well thanks to @jkachhad.  What I've done is definitely janky, though, given the `CityInfo` data sent from the server.  If @jaedan is able to update the server side, what I've written could be replaced by the more robust solution described by @jkachhad.

Issue #240, opened 2019-01-28
Russian characters in login textbox


zolter:
Other utf-8 characters also create an error and crash client.

![97ef6-clip-57kb](https://user-images.githubusercontent.com/1370602/51862531-7f7c1800-2347-11e9-9d6c-2498c9e9d19e.png)

![3zrvwji](https://user-images.githubusercontent.com/1370602/51862532-7f7c1800-2347-11e9-8524-50b5148cc0ce.jpg)

![3zrw0ih](https://user-images.githubusercontent.com/1370602/51862623-acc8c600-2347-11e9-9218-a0013c1a6675.jpg)

zolter:
crash too 
i think need ignore all > 127

![3zrwmeo](https://user-images.githubusercontent.com/1370602/51863003-9707d080-2348-11e9-88dd-c941e4fbea1c.png)

andreakarasho:
Textbox should be support standard Eng characters only? Can you write rus chars on classic client textboxes?

zolter:
Of course not, the classic client just ignores them (skips) like the other >127 characters & unicode. That is why I added "TextEntry.SafeCharactersOnly()" to match the behavior of the standard client. I think this flag should be enabled by default, unless the "isUnicode" option is set.

Now when we try to enter Russian (or other regional characters), we get an exception and a ClassicUO crash. This does not match the behavior of the classic client  :)

zolter:
client_7.0.18.0 ignore all <32 && >127 characters + # 
in "account name" and "password" Textbox

andreakarasho:
the original client does it deny to put rus chars everywhere?

zolter:
No, only account name and password

rus chars correct work in options and other fields: 

![3ztinsr](https://user-images.githubusercontent.com/1370602/51920472-274c2100-23ee-11e9-908f-1ef69ffcefce.jpg)

zolter:
Fixed in #243

PR #243, opened 2019-01-28
TextEntry.SafeCharactersOnly()

zolter:
#240 
How about this option?


zolter:
Add SafeCharactersOnly to _nameTextBox (CreateCharAppearanceGump) and _textboxPassword (LoginGump). Now matches the behavior of the original client and does not cause a CUO crash.

andreakarasho:
nice

PR #250, opened 2019-01-30
Use keyboard buttons to move player

zolter:
arrow + keypad

### DIFF
--- a/src/Game/Scenes/GameSceneInputHandler.cs
+++ b/src/Game/Scenes/GameSceneInputHandler.cs
@@ -52,6 +52,26 @@ namespace ClassicUO.Game.Scenes
         private Entity _queuedObject;
         private bool _rightMousePressed, _continueRunning;
 
+        private readonly Dictionary<SDL.SDL_Keycode, Direction> _keycodeDirection = new Dictionary<SDL.SDL_Keycode, Direction>()
+        {
+           { SDL.SDL_Keycode.SDLK_LEFT, Direction.Left },
+           { SDL.SDL_Keycode.SDLK_RIGHT, Direction.Right },
+           { SDL.SDL_Keycode.SDLK_UP, Direction.Up },
+           { SDL.SDL_Keycode.SDLK_DOWN, Direction.Down },
+        };
+
+        private readonly Dictionary<SDL.SDL_Keycode, Direction> _keycodeDirectionNum = new Dictionary<SDL.SDL_Keycode, Direction>()
+        {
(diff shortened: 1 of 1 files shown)


src/Game/Scenes/GameSceneInputHandler.cs
```diff
@@ -52,11 +54,13 @@ partial class GameScene
         private Entity _queuedObject;
         private bool _rightMousePressed, _continueRunning;
 
-
         public bool IsMouseOverUI => Engine.UI.IsMouseOverAControl && !(Engine.UI.MouseOverControl is WorldViewport);
 
 	    private bool _isShiftDown;
 
+        [DllImport("user32.dll", CharSet = CharSet.Auto, ExactSpelling = true, CallingConvention = CallingConvention.Winapi)]
```
andreakarasho: it's not crossplatform!

andreakarasho:
instead use the class GameSceneINputHandler
and inside the KeyDown event handle the arrow buttons to move char

src/Game/Scenes/GameSceneInputHandler.cs
zolter: [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
        [return: MarshalAs(UnmanagedType.Bool)]
        private static extern bool SetDllDirectory(string lpPathName);

not crossplatform too ?

zolter:
There were problems with this, now I will try again.

src/Game/Scenes/GameSceneInputHandler.cs
andreakarasho: No, but when you use mono, we have the file .config that uses dllmap

zolter:
> and inside the KeyDown event handle the arrow buttons to move char

No problem to handle the arrows. 
There is a problem with cross-platform getting numlock state. 
I'll try :)

andreakarasho:
SDLK_NUMLOCKCLEAR

zolter:
SDLK_NUMLOCKCLEAR - this is not a button state. 
Movement should occur only when the "Numlock" button is off

andreakarasho:
(keymod & KMOD_NUM) == (KMOD_NUM)

zolter:
ofc you right
strange why I forgot about it :/

andreakarasho:
Is it ready?

zolter:
Yep, it works correctly, as you wrote above. 
In the evening I will try to make the code smaller and remove 2 switch.

zolter:
Is done

PR #251, opened 2019-01-31
DebugGump() small display mode dclick & CloseWithRightClick

zolter:
- "Debug" button in top menu - restore DebugGump() 
- Left DClick changes the display mode (full/small)
- Right Click - close debug gump

3zwn7ad

I think it is very convenient.


andreakarasho (approved):

andreakarasho:
I like it

PR #254, opened 2019-02-01
debugGump.Dispose();

zolter:
I confirm the problem, it is reproduced only for this gamp. I make a temporary patch, then I will take a closer look at why this is happening.

3zyf5zg


andreakarasho (commented):
If you dispose it you cannot reopen it, but maybe you used a different approach by your one of last PR about it.
If you can confirm, everything works fine i accept your pr

zolter:
yep ofc

PR #265, opened 2019-02-05
Dispose SplitMenuGump instances when the associated ItemGump instance is disposed

mngraves:
This behavior matches client 7.0.15.1.


andreakarasho:
Please redirect your PR to the 'dev` branch

PR #271, opened 2019-02-06
Fixed vendor items order mismatch between 0x3c and 0x74 packets

mngraves:
This was causing vendor items to have the wrong names and prices.

I also tried to simplify the handling of 0x24 vendor packet so that it will just dispose other ShopGumps and display the target gump. The 0x74 packet handler creates the target ShopGump in a hidden and disabled state.

Opening a Buy or Sell window disposes other active ShopGumps.

Tested on Outlands server.

### DIFF
--- a/src/Game/Managers/UIManager.cs
+++ b/src/Game/Managers/UIManager.cs
@@ -577,6 +577,19 @@ namespace ClassicUO.Game.Managers
             return gump;
         }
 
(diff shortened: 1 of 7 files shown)


andreakarasho:
Strange, I already fixed it.
Look:
image


Where did you see the issue?

mngraves:
I noticed it at the weaponsmith on Shelter Island. How does that one look to you?

andreakarasho:
Andaria, the farmer near the city bridge

PR #281, opened 2019-02-13
(macro) BandageSelf & BandageTarget | without NewTargetSystem

zolter:


zolter:
When I get home, I'll finish BandageTarget()
No need marge now

zolter:
Now I'm done, for MacroType.BandageTarget **(>CV_5020)** - I added a stub. 
Back to this after we do NewTargetSystem. 

I also made an **Engine.Profile.Current.BandageSelfOld** to switch the heal target method, so that new customers could use the old version that expects target. This is true for many SphereSrv that have a custom heal system.

The code largely corresponds OrionUO/CrossUO
I think this is better than nothing :)

andreakarasho:
ty

PR #282, opened 2019-02-14
Recursive Item.FindItem() + modify FindBandage

zolter:


andreakarasho:
good, so now you can remove the IsBandage property. I think it is useless, basically a Graphic == [bandage_graphic] is enought

zolter:
true
+

PR #293, opened 2019-02-22
+ return last Recalculate() code

zolter:
400isia

400ivxw


SaschaKP:
Yes, I've already noted this, I've made a change that should fix it, currently the behaviour is not exactly the same as in original, original mazes at 20, while in cuo & Orion you can go over that value

SaschaKP:
I'm accepting this as a revert / quick fix. But will submit a change that behaves as original on next commit

Issue #276, opened 2019-02-12
Pets disappear offscreen


jnelson76:
**Describe the bug**
When you run to far away from pets and come back you can no longer see them.  After a bit they reappear sometimes

**To Reproduce**
Run fast away from pets until they offscreen.  Then run back to them.

**Expected behavior**
Pets to stay visible

**Additional context**
Orion had this issue also.

andreakarasho:
Hi, i need some infos:

- shard
- shard client version
- OS

jnelson76:
- Outlands
- 7.0.15.1
-W10 1803
-I was on latest dev pull

I'm Mordekai

Nirad:
any pet ? or horse or ...

jnelson76:
It was a Dragon and a Drake

andreakarasho:
try to set fixed time step = false

andreakarasho:
News?

andreakarasho:
fixed https://github.com/andreakarasho/ClassicUO/pull/298

PR #297, opened 2019-02-24
Quick fix for 'Pets disappear offscreen' issue #276

Mentos-:
Was able to reproduce the 'Pets disappear offscreen' issue #276 
Here is a video reproducing the issue: http://recordit.co/oJIyqc2pej

The client was still receiving UpdateCharacter packets for the pet but the pet's position was not properly updating and was stuck at x:65535 y:65535 z:0

Now UpdateCharacter will detect if the mobile's position is ever invalid and force a new Position update from the packet's x/y/z values


zolter:
Please use dev branch

PR #299, opened 2019-02-24
HealthBarGump name validation

zolter:


zolter:
**Client 7.0.18.1 can't use:**
1. special characters
2. numbers
3. blank name
4. russian letters and etc

PR #317, opened 2019-02-28
GetTexture ContainsKey g fix

### LINKED ISSUE #303: Elf (7.0.18)
### CUO

404kxdm

404kzkl

### 7.0.18 (all ok)

404kgs0

zolter:
#303

### DIFF
--- a/src/IO/Resources/GumpsLoader.cs
+++ b/src/IO/Resources/GumpsLoader.cs
@@ -71,13 +71,17 @@ namespace ClassicUO.IO.Resources
         {
             if (!ResourceDictionary.TryGetValue(g, out SpriteTexture texture) || texture.IsDisposed)
             {
-                ushort[] pixels = GetGumpPixels(g, out int w, out int h);
+                if (!ResourceDictionary.ContainsKey(g))
+                {
+                    ushort[] pixels = GetGumpPixels(g, out int w, out int h);
 
-                if (pixels == null || pixels.Length <= 0)
-                    return null;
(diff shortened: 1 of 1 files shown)


andreakarasho:
the `TryGetValue(g, out SpriteTexture texture)` already check if value exists

zolter:
No :)

zolter:
a2

a1

end..... HELLO 5593 :)

a3

andreakarasho:
Yes. Internally it check if key exists, if it exists return true and the SPriteTexture.
I think the problem is the second check: || texture.IsDisposed, because it means that i already exists into dictionary and when it does dictionary.Add it throws an exception.
There is a problem with that texture, it is disposed manyally somewhere around the code

zolter:
Yep, without ` || texture.IsDisposed` all work fine

andreakarasho:
Ok. Im seeing that 5593 is related to mapinfo when you chose the map.
Im investigating..

andreakarasho:
```csharp
if (mapInfo.Index == 0 || mapInfo.Index == 3)
          {
              var texture = FileManager.Gumps.GetTexture(mapInfo.Gump);
              if (texture == null)
              {
                  SkipSection = true;
                  return;
              }
              else
                  texture.Dispose(); // TOTALLY WRONG!!!!!!!
          }
```

andreakarasho:
Try it now with latest commint please

zolter:
Yep, brakepoint stop in this line

PR #319, opened 2019-03-03
Underscore as valid input

jkachhad:
The underscore character ("_") is a valid input for account names. Not sure if this was the best way to implement, or to create another Constants.RULES category.


zolter:
I fixed it a long time ago
https://github.com/andreakarasho/ClassicUO/pull/318
:)

Issue #337, opened 2019-03-07
dev build:  "no tile" icons make looting difficult


oroechimaru:
Often in inferno on uo outlands last night when looting near "big trees" by the lesser demons, the "corpse" (ctrl shift) would be blocked by "no tile" words (similar to the corpse word)

Once in a while a black square (old school 1998 bug) would also pop up saying no tile and make my mouse not work for a few seconds.

Possibly these are background tiles on the map or old corpses? One issie with classic uo is the static map pieces are all clickable even though not functional (the whole object like a tree or building or tile or art turns purple and can be clicked). This could be related.

zolter:
Try disable this flag https://c2n.me/40h5cyp.jpg

oroechimaru:
i will test tonight thx. if it works probably good to set it as default (unchecked) for users

oroechimaru:
I confirmed this resolves the issue (in dev). Please make "unchecked" the default box and/or hide it for wide use adoption. Feel free to close it after reviewing my suggestion thanks!

PR #343, opened 2019-03-08
Disable HighlightGameObjects & Reconnect by def

zolter:
#337


zolter:
Disabled Reconnect by def (in this PR)

+++

PR #360, opened 2019-03-17
Fullscreen minor update and fixes (options)

zolter:
1. https://i.imgur.com/xTgBT7W.gif
2. https://c2n.me/40s15W4.png

I think problem could be related to editing fields (window size, window padding) and setting the flag (fullscreen) at the same time. Also added support for negative values in Numeric field (+Unsigned Only Support).


zolter:
wait, have 1 problem

zolter:
+++

PR #370, opened 2019-03-26
Plugins can now be defined in settings.

DarkLotus:
Move plugin definitions to settings.json so other plugins can be loaded without source changes.


andreakarasho:
Thanks :)

PR #390, opened 2019-04-02
Load plugins via reflection

DarkLotus:
Attempts to load a plugin via the existing method and if that fails falls back and attempts to load directly via reflection, This allows Razor and other .NET plugins to be loaded on linux/macosx without a native shim.

Plugins must implement the following to be loaded as a Managed plugin.
```csharp
namespace Assistant 
  { 
    public class Engine 
    {
      public static unsafe void Install(PluginHeader *plugin){}
    }
  }
```

Note the `ref` is removed from `Install` Method signature as Mono segfaults on the Method Invoke if its there.
If anyone has an idea how to solve this please let me know.

### DIFF
--- a/src/Network/Plugin.cs
+++ b/src/Network/Plugin.cs
@@ -91,28 +91,7 @@ namespace ClassicUO.Network
             _getStaticImage = GetStaticImage;
             _getUoFilePath = GetUOFilePath;
 
-            IntPtr assptr = SDL2EX.SDL_LoadObject(_path);
-
-            Log.Message(LogTypes.Trace, $"assembly: {assptr}");
-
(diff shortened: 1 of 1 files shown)


DarkLotus:
To be clear this changes no behavior for current Windows users. CUO will still load their Razor.dll as it currently does.

This just gives Linux/OSX users a path to get plugins that doesn't involve writing a native library for their platform as a shim.

If you dont like the way I've implemented this, im happy to rewrite following any suggestions.

Linux / OSX users have missed out on Razor ( without running shit thru wine that half works ) for far too long.

zolter:
I think we can test it in dev 👍

Issue #376, opened 2019-03-29
Provocation "instant double clicks" and does not function only from skill gump


oroechimaru:
alpha 5:

a. drag the skill gump from the skills list for provocation (so you have the black button for provocation)

b. use the skill gump

c. target your first creature. it instantly will double click the same creature as target #2 which is not valid and will fail. 

d. workaround: create a hotkey for the provocation skill from razor. This works without issue .

PR #404, opened 2019-04-05
Targeting health gump for provocation fix

simarsenault:
Provocation skill didn't work when using the health bar as target (works fine when targeting directly the creature).

Video: https://youtu.be/x9K0WNmIIQs

What happens:

1. Client to server: use skill provocation
2. Server to client: please select first target
3. Client to server: send first target
4. Server to client: please select second target
5. Client to server: send first target again

It happens because both `OnMouseDown` and `OnMouseClick` handles Targeting and it happens so fast that it does this sequence:

1. Client: `OnMouseDown` (send first target to server)
2. Server: Ok, please select 2nd target
3. Client: `OnMouseClick` (send "second" target to server which is still the first target)

### DIFF
--- a/src/Game/UI/Gumps/HealthBarGump.cs
+++ b/src/Game/UI/Gumps/HealthBarGump.cs
@@ -504,15 +504,6 @@ namespace ClassicUO.Game.UI.Gumps
                 _textBox.IsEditable = false;
         }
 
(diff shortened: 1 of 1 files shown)


simarsenault:
I think it would fix #376

andreakarasho:
Thank you!

PR #408, opened 2019-04-05
they are just .so on linux

DarkLotus:


andreakarasho:
I looked at fna src code:
https://github.com/FNA-XNA/FNA/blob/master/FNA.dll.config

Do you receive errors with current settings?

DarkLotus:
https://github.com/andreakarasho/ClassicUO/tree/master/src/libs/lib If you look in our libs folder ours are named just .so not .so.0

with .so.0 Client crashes with no audio device found because the library is missing.

andreakarasho:
That's weird. Im totally agree with you, but all fna game should throw this exception.
Anyway i think you found the issue with mac also. Before merge, can you fix the whole file with the right libraries name?

DarkLotus:
The rest look okay on my end.

PR #411, opened 2019-04-06
Outlands Statusbar txt label X-Y update

Belialzao:
Updated the X and Y location of strings to be pixel perfect with the original Outlands status bar.

Changed followers counter from:
5
5 format 

to 5/5 format


zolter:
Hi,
Please use dev

PR #418, opened 2019-04-06
Outlands statusbar txt labels update

Belialzao:
Updated the X and Y location of txt labels to be more accurate with normal client. (Outlands)

Changed followers counter from:
5
5 format 

to 5/5 format.


andreakarasho:
thank you

PR #421, opened 2019-04-08
Settings system overhaul

binff04:
Right settings loading order: at first, we load settings from the main settings file (settings.json) and then rewrite some settings with those passed in command-line arguments (in shortcut).


zolter:
ty 🍺

PR #469, opened 2019-05-02
Responsive keyboard movement

roxya:
- Keyboard movement (arrow keys, numpad) runs by default, walk with shift
- Keyboard movement is much more responsive with fast direction changes, matching the abilities of mouse movement
- Arrow key movement can now move diagonally when two keys are pressed


natryl:
This is better. Still a delay when changing directions. Is it possible to have two key presses register a diagonal? like up and right = NE. And then the ability to let us hotkey those where we like to WASD

PR #470, opened 2019-05-03
Added additional options to healthbar mouseover highlight

markdwags:
On the standard client, it will always highlight on mouseover regardless of mobile state. This commit gives CUO that option along with additional options (never, warmode, targeting and always)


andreakarasho:
in review

andreakarasho:
After the recent changes this PR is not necessary anymore. 
The mob highlighting happens everytime your mouse is over the healthbargump and/or your character is in warmode and your mouse is over a mobile.
I'm not sure to add other features like: highlight in warmode only.
Atm i want to follow the standard behaviour

PR #495, opened 2019-05-16
Remove "Shift to Walk" (for both mouse and keyboard move)

roxya:
Hard coded Shift-to-Walk has caused undesirable effects, main example is slowing the player when using Ctrl+Shift while running.

The macro system has AlwaysRun as a toggle key which satisfies people who want to use Always Run but also walk when they need to.


zolter:
Agree, need add "run/walk" switching in macros next time.

PR #498, opened 2019-05-25
Fix for pathfinding diagonals (and therefore autowalk)

roxya:
At the moment, pathfinder will not move diagonally because of a typo (comparing x to wantY instead of wantX), and also because diagonals have higher 'cost' even though they shouldn't on the UO grid.

It's not perfect, some unnecessary zig-zagging towards walls at times and too many turns being made, but it's a lot better. Next step would be to potentially prefer routes with less changes of direction.


andreakarasho:
Reviewing :)

PR #500, opened 2019-05-29
Add Auto Open Doors, and Auto Open Corpses as Experimental features.

DarkLotus:
Quick does not want to add Auto Open Corpses during movement added to razor.

So add to the client instead.

Also add Auto Open Doors, for those poor souls not able to use Razor.

PR 500!


ichiokada:
Thank you!!! 😍

zolter:
ty!

PR #504, opened 2019-05-31
Clothing layer fixes

roxya:
Hair and Necklace should draw above Robe
image

Tunic "covers" Torso so torso was not rendering under any tunic (I have not found a tunic which does completely block torso rendering)
image

Arms should draw below Torso in paperdoll
image


roxya:
The paperdoll is different for different armour types... hm

PR #505, opened 2019-06-01
Clothing layerorder fix

roxya:
Hair and Necklace should draw above Robe
image

Tunic "covers" Torso and prevents rendering of Torso (I have not found a tunic which does completely block torso rendering)
image

There are also issues with paperdoll order, where arms are rendered above the torso, but in most cases this shouldn't be. Plate arms should render above torso but other arms below. Perhaps plate arms should be assigned a "fake" layer (Layers.PlateArms) but I'm not sure of the implications.

### DIFF
--- a/src/Game/Data/LayerOrder.cs
+++ b/src/Game/Data/LayerOrder.cs
@@ -28,28 +28,28 @@ namespace ClassicUO.Game.Data
         public static Layer[,] UsedLayers { get; } = new Layer[8, Constants.USED_LAYER_COUNT]
         {
             {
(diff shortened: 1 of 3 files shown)


andreakarasho:
What is the behaviour with an hooded dressed?

roxya:
Ah, Death shroud/GM robe/elven robe will "cover" hair regardless but this needs to apply to neck as well...

roxya:
In fact necklace shows through hooded as well:
image

It's "technically" wrong to see a gorget like this but people already get upset about their fashion... I think we are aiming to match the standard behaviour first?

andreakarasho:
Approved.
Thank you man!

PR #508, opened 2019-06-04
Add Further options for corpse opening, None, Hiding, Targeting, Both.

DarkLotus:
Adds a drop down to the options menu allowing you to choose between None, Not Targeting, Not Hiding, or Both for dealing with corpse opening.

Also attempts to open corpses when a new corpse appears, not just on movement.


andreakarasho:
thanks!

PR #515, opened 2019-06-10
Drag-select to get health bars

roxya:
For your consideration...

dragselect_a
dragselect_b

Pulling health bars while I'm moving, or while the opponent is moving, has always been a huge annoyance to me. I saw the "selection area" in experimental features and wondered, what the hell is that? Disappointingly, I found that it's for selecting text. But that's when my idea was formed.

This will

- Allow click and drag either just using left mouse down, or optionally requiring a modifier key (ctrl or shift)
- Allow click and drag only from whitelisted types (e.g. can drag if you start on a static, a multi, an item as long as it's locked...)

I haven't found any remaining bugs for it. Anyway whatever happens with it, I had fun making it!


Belialzao:
fucking beautiful !!!

nocturnalllama:
So does this work in the new dev? I love this!

zolter:
I think it's interesting, worth a try.
ty! ❤️

PR #520, opened 2019-06-15
Drag-select option for humanoids only

roxya:
Had a lot of requests for this


zolter:
Sorry, resolve conflict please :/ 🍻

roxya:
Sorry for that, should be OK to go now. Thanks for fixing the options before :)

PR #521, opened 2019-06-20
drag select humanoids only option

roxya:
This disappeared when I accidentally deleted the branch while "cleaning" my github! But good chance to rebase and remove any conflicts that appeared along the way.

(pvp players asked for this feature)

### DIFF
--- a/src/Configuration/Profile.cs
+++ b/src/Configuration/Profile.cs
@@ -167,6 +167,7 @@ namespace ClassicUO.Configuration
         [JsonProperty] public bool EnableDragSelect { get; set; }
         [JsonProperty] public int DragSelectModifierKey { get; set; } // 0 = none, 1 = control, 2 = shift
         [JsonProperty] public bool OpenContainersNearRealPosition { get; set; }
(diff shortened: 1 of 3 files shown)


andreakarasho:
I'm not sure about it.
Seems like a limitation in different context.
For example: if you have to drag an heal potion from the ground to survive, with this option ON you will die.

roxya:
It doesn't drag if you mouse down on an unlocked Item

https://github.com/andreakarasho/ClassicUO/blob/045b2bbceeaa7152fbc8633a0280049c822752fd/src/Game/Scenes/GameSceneInputHandler.cs#L124-L127

And this pr just adds a filter to what mobiles are included in the already existing function

andreakarasho:
Exactly if you cannot drag the potion (for example) your gameplay is really limited.
It's dangerous. Any workaround ?

andreakarasho:
Ok. It has been a misanderstand. This PR is referred to the DragSelection feature.

Can you solve conflicts?

roxya:
Done :)

PR #525, opened 2019-06-21
Prevent spell cast when dragging spell icon with "Cast spells by one click"

roxya:
Made it so there must be an offset of 1px or less to count as a click. Not sure if it's the right way - the buttons to use skills (SkillButtonGump) has a hidden feature where you must hold Alt to prevent a "click". Could copy that here instead but would anyone know to use it?

### DIFF
--- a/src/Game/UI/Gumps/UseSpellButtonGump.cs
+++ b/src/Game/UI/Gumps/UseSpellButtonGump.cs
@@ -21,6 +21,7 @@
 
 #endregion
 
+using System;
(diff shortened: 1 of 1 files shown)


andreakarasho:
I'll remove the support to MouseClick event.
Use MouseUp.
Inside MouseUp you can check the dragoffset distance (usually 5 px to figure out if the user moved the gump or clicked it).

roxya:
I'll keep it on my list and do it again soon

PR #554, opened 2019-06-30
Don't close existing bars on drag select

MayDF:
Prevents closing of health bars that are already opened when drag selecting bars.


roxya:
I didn't do this because it pushes it to be a bit too strong, but not my choice it it comes in

MayDF:
Sorry, didn't know there's some issue with this.  Thinking from pet perspective, it's qol issue.

andreakarasho:
in review, but please point to dev branch

PR #556, opened 2019-06-30
Changes for GridLoot Grab Bag and Drag Select

MayDF:
Updated to point to dev
Include changes for GridLoot grab bag and prevent closing bars if already open from drag select


SaschaKP:
merged wih latest commit https://github.com/andreakarasho/ClassicUO/commit/83b01c21c0c905e5c8ad09522dc8ddd341ea6671

PR #607, opened 2019-07-16
✨ Add UseItemInHand macro

canterberry:
This macro works just like the Razor Hot Key action of the same name.
Personally, I use it for lumberjacking.


zolter:
Please use andreakarasho:dev

PR #609, opened 2019-07-16
✨ Add -hue command for quick inspection of in-game items

canterberry:
This command is available in the UO:Renaissance client, but was not available in ClassicUO. It is a pared down version of the information available in the "-info" command gump, intended for rapid inspection of the most common "hidden" properties: item ID and hue.

Personally, I use this for organizing NPC cloth I take from brigands, escorts, evil mages, and the vendors in Buc's Den.

Screen Shot 2019-07-16 at 13 42 52

Screen Shot 2019-07-16 at 13 43 05

### DIFF
--- a/src/Game/Graphic.cs
+++ b/src/Game/Graphic.cs
@@ -87,7 +87,7 @@ namespace ClassicUO.Game
 
         public override string ToString()
         {
-            return $"0x{Value:X4}";
+            return $"{Value} (0x{Value:X4})";
         }
 
         public override int GetHashCode()
--- a/src/Game/Managers/CommandManager.cs
+++ b/src/Game/Managers/CommandManager.cs
@@ -24,6 +24,8 @@
 using System;
 using System.Collections.Generic;
 
+using ClassicUO.Game.GameObjects;
+using ClassicUO.Input;
 using ClassicUO.Utility.Logging;
 
 namespace ClassicUO.Game.Managers
--- a/src/Game/Managers/CommandManager.cs
+++ b/src/Game/Managers/CommandManager.cs
@@ -38,7 +38,7 @@ namespace ClassicUO.Game.Managers
             Register("info", s =>
             {
                 if (!TargetManager.IsTargeting)
-                    TargetManager.SetTargeting(CursorTarget.SetTargetClientSide, 6983686, 0);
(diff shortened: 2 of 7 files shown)


src/Game/Graphic.cs
```diff
@@ -87,7 +87,7 @@ public int CompareTo(ushort other)
 
         public override string ToString()
         {
-            return $"0x{Value:X4}";
+            return $"{Value} (0x{Value:X4})";
```
canterberry, line 90: :bulb: I made this adjustment to both `Graphic` and `Hue` to aid in readability of these values, since they are often referred to either by their hex or decimal representation (for example, in UO:R, hues are generally referred to by their decimal representation).

src/Game/Managers/CommandManager.cs
```diff
@@ -38,7 +38,7 @@ public static void Initialize()
             Register("info", s =>
             {
                 if (!TargetManager.IsTargeting)
-                    TargetManager.SetTargeting(CursorTarget.SetTargetClientSide, 6983686, 0);
+                    TargetManager.SetTargeting(CursorTarget.SetTargetClientSide, CursorType.Target, TargetType.Neutral);
```
canterberry, line 43: :bulb: `6983686` was a magic number, so I externalized that. Open to suggestions on where this enumeration should live -- within the `TargetManager` may not be exactly the right place for it.

src/Game/Managers/CommandManager.cs
```diff
@@ -87,5 +94,12 @@ public static void Execute(string name, params string[] args)
             else
                 Log.Message(LogTypes.Warning, $"Commad: '{name}' not exists");
         }
+
+        public static void OnHueTarget(Entity entity)
+        {
+            TargetManager.TargetGameObject(entity);
+            Mouse.LastLeftButtonClickTime = 0;
+            GameActions.Print($"Item ID: {entity.Graphic}\nHue: {entity.Hue}");
+        }
```
canterberry, line 105: :bulb: This is a pattern that didn't exist before. It didn't seem right to have the `-hue` command logic spread across 4-5 different files like `-info` is currently. In RunUO, each command is its own class altogether and is registered in a static `#Initialize()` method. That may be a pattern to follow here, too.

PR #643, opened 2019-07-28
fix moving attempt while paralyzed or frozen

Ultima-Tony:


roxya:
I confirm my 7.0.15.1 client behaves like this

PR #650, opened 2019-08-07
Minor pixel alignments

Belialzao:
Alignment corrections:
  - Virtue Button (1px to the right)
  - Item Slots (height and gap)
  - Wearables (2px up)

(wish I could decrease text line height by 1px too, to be identical to the original client)


oroechimaru:
that is some dedication!

SaschaKP:
rebased and merged, I don't see why this shouldn't be integrated, to be at least in some points equal to original client.

PR #652, opened 2019-08-13
Engine partition

Scotty1234:
Start of separating out engine parts from game to make it easier to improve rendering and underlying framework. Getting this in early as there are lots of file moves and do not want to disrupt people.


andreakarasho:
added in rendering-engine

PR #659, opened 2019-08-18
Dev - Add custom aura for party members, toggle & hue in options

Stallind:
https://gyazo.com/16705c5828f9d7d0bd081ee8d15fe82e

### DIFF
--- a/src/Game/GameObjects/Views/MobileView.cs
+++ b/src/Game/GameObjects/Views/MobileView.cs
@@ -43,6 +43,7 @@ namespace ClassicUO.Game.GameObjects
         private static int _startCharacterKneesY;
         private static int _startCharacterFeetY;
         private static int _characterFrameHeight;
+      
 
         public byte HitsPercentage { get; set; }
         public RenderedText HitsTexture { get; set; }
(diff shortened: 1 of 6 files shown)


src/Game/GameObjects/Views/MobileView.cs
```diff
@@ -96,10 +98,18 @@ private void DrawCharacter(UltimaBatcher2D batcher, int posX, int posY)
 
             bool hasShadow = !IsDead && !IsHidden && Engine.Profile.Current.ShadowsEnabled;
 
-
             if (Engine.AuraManager.IsEnabled)
-                Engine.AuraManager.Draw(batcher, drawX + 22, drawY + 22, Notoriety.GetHue(NotorietyFlag));
-
+            { 
+                if (World.Party.Contains(this) && Engine.Profile.Current.PartyAura)
```
andreakarasho: Just an optmiziation.
It's much better to evalutate `Engine.Profile.Current.PartyAura` before, cause the World.Party.Contains do a cycle that in the worst case is length 10.

Stallind:
Fixed

PR #669, opened 2019-08-24
Fix ExePath for proper work in Mono environment

binff04:
`Process.GetCurrentProcess().MainModule.FileName` in Mono environment returns a path where mono executable located and not the path where ClassicUO.exe located


binff04:
Oops, old commits joined the party. I will clean them first.

PR #672, opened 2019-08-24
Add check for non-lootable items

binff04:
When rendering standard corpse loot gump and modern GridLootGump, do not render non-lootable items (on Outlands they are identified by 0,0,0 position sent from the server).


andreakarasho:
I'd like to know if this is a common behaviour or outlands only

oroechimaru:
when i tested steam they still appeared but after the real loot instead if before (cuo)

SaschaKP:
on RunUO as well as ServUO the location inside the container is always sent without a change, so we can suppose this is an Outlands only feature, maybe they use it because steam order them internally by location, from greater to lower.

Alias-Alan:
it's an issue with custom humanoid scripts, basicslly it's disposing the outfit properly thus shows on container bounds.

binff04:
At the moment I only can confirm it on Outlands, this is why I implemented that `shard_type` check.

On an early stage Outlands sent those non-lootable items among the normal loot, but after some time (people probably complained a lot), they started to send those items at that 0,0,0 position so they were moved out of main container bounds to 0,0,0 position which is most likely a "blind zone" in a container, at least my tests showed that you can't manually place an item in that zone in any container (I used `-info` command to check items' positions).

The idea behind this change is to get rid of this non-lootable items at least temporarily before it will be resolved on the server side.

P.S. Gameplay and usage of GridLootGump is so much cleaner without these items taking place.

binff04:
Added `FIXME` comments, so we will be able to address this check when non-lootable items will be "fixed" on the server side on Outlands.

canterberry:
I'm not sure how non-lootable items work on Outlands, or which items are considered non-lootable, but before this is merged, I want to call attention to UO:R's "monster packs", which appear as non-lootable containers in the top-level of a corpse. The monster's loot is contained within their "pack". This feature on UO:R is specifically designed to combat "grid loot" and "auto loot" features.

Currently, ClassicUO's grid loot feature shows these items in the grid, but in agreement with @binff04 , non-lootable items have no reason to be in the grid loot gump.

I want to make sure, however, that these items **do** appear in the corpse gump for standard drag-looting and container interaction.

src/Game/UI/Gumps/ContainerGump.cs
```diff
@@ -63,10 +63,22 @@ public ContainerGump(Item item, Graphic gumpid) : this()
...
+            {
+                if (i == null || 
+                    i.ItemData.Layer == (int) Layer.Hair || 
+                    i.ItemData.Layer == (int) Layer.Beard || 
+                    i.ItemData.Layer == (int) Layer.Face ||
+                    // Skip non-lootable items on Outlands (server always sends their position as 0,0,0)
+                    // FIXME: Remove this check when the issue with non-lootable items will be resolved on the server side
+                    (Engine.GlobalSettings.ShardType == 2 && i.X == 0 && i.Y == 0 && i.Z == 0)
+                )
+                    continue;
```
canterberry, line 83: 📝 Follow-up to my PR comment: this doesn't seem like it should affect UO:R's monster packs, so I'm on board. However, I'd suggest leaving `ContainerGump` unmodified, start by just making the changes to `GridLootGump`.
Better yet, incorporate a configurable option to enable/disable rendering non-lootable items in the grid loot gump, in case some players would be confused to suddenly stop seeing items appear in the gump which they know should be on the corpse, even if they can't loot those items.

binff04:
@canterberry non-lootable items on Outlands, at least, have a very distinct behavior: they all appear in the upper left corner of the corpse gump and they all have X:0,Y:0,Z:0 position where no other items can be placed. When you try to pick up such items they disappear immediately.

The check that I implemented in this PR has a temporary nature: when the non-lootable items will be fixed on the server-side on Outlands shard, then the check can be removed. But right now, it greatly helps to avoid clutter in both corpse and GridLoot gumps.

src/Game/UI/Gumps/ContainerGump.cs
binff04: You are absolutely right: the `ShardType` check is there to only apply that for Outlands shard where the behavior of non-lootable items is confirmed.

If you check the code closely, then you can see that we check for the exact item position in the container. This position is a dead zone for any container and no items can be placed there manually by a player.

binff04:
Just checking in about this PR: should I just close it or should I resolve conflicts and we will be able to merge it and keep that checks for at least until it will be resolved on the Outlands' side?

I can keep an eye on the changes and as soon as the situation with the non-lootable items on Outlands will improve, I will remove that check.

Honestly speaking, those non-lootable items may be so distracting when using GridLoot, so they can occupy almost an entire page on GridLoot gump making it inconvenient to use.

oroechimaru:
I have gotten use to it. Whatever is best for devs

binff04:
Yes, of course you can get used to it, but what's the reason if it's only a matter of one small check, right?! ;)

Hezkore:
I'm at least still bugged by having floating items outside of the container.
So I'd like to see this merged.

binff04:
I'm going to close this PR because of the conflicts with the current `dev` branch. I will open a new one for review.

PR #673, opened 2019-08-24
buffs gump toggle from menubar

roxya:


andreakarasho:
Honestly i'd like to make accessible this button from status bar directly, like standard status bar does.
So my suggestion is to take the gameborder button (the resizer) and put it into the outlands status bar.

Reasons:
- most of the people disable the top menu bar 
- the topbar gump is not the right context in my opinion
- standard status bar already has this behaviour

Thanks rox for you help

roxya:
I get the point that people may disable the top menu. But that goes for the status gump as well, many people aren't using it.

I'll do it as you suggested to maintain the existing pattern but this bar is so fundamental to most people's experience... I wonder if just having it enabled in the default profile will mostly solve it. Since for most of us it's a case of enabling it and then never touching it again.

PR #675, opened 2019-08-24
Fix chat focus events and other chat improvements

binff04:
This commit addresses two issues:
1. When keyboard focus was switched from active/inactive chat input field to any other input field and then switched back to chat, the SystemChatControl didn't get proper event about it getting focus back. This invalid behavior was fixed and now when game returns keyboard focus back to system chat, its `OnFocusEnter` event will be properly fired;
2. When other than system chat input fields were focused for keyboard input, game still executed macros and passed hotkey events to plugins. This is now also fixed and if any other text input field but system chat is focused, any input from the keyboard will ignore macros' and plugins' hotkeys.

### DIFF
--- a/src/Configuration/Profile.cs
+++ b/src/Configuration/Profile.cs
@@ -151,7 +151,6 @@ namespace ClassicUO.Configuration
         [JsonProperty] public bool UseDarkNights { get; set; }
         [JsonProperty] public int CloseHealthBarType { get; set; } // 0 = none, 1 == not exists, 2 == is dead
         [JsonProperty] public bool ActivateChatAfterEnter { get; set; }
-        [JsonProperty] public bool ActivateChatStatus { get; set; } = true;
         [JsonProperty] public bool ActivateChatIgnoreHotkeys { get; set; } = true;
         [JsonProperty] public bool ActivateChatIgnoreHotkeysPlugins { get; set; } = true;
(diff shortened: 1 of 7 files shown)


binff04:
Since this PR is still open, I will ask a question that I briefly discussed in Discord: do we really need those additional checkboxes to disable macros/Razor hotkeys if we use "Activate chat by 'Enter' pressing"-thing?

Since it now perfectly tracks the focus, what if we will make it a default behavior? I mean, if the "Activate chat by 'Enter' pressing" is enabled and chat is active, then we definitely trying to enter some text and don't expect to cast spells and use other hotkeys. Also, if the keyboard focus is on any other text input field, other than SystemChat, then it's again we try to type some text somewhere else. The only case when we really expect to fire macros and execute assistant's hotkey is when SystemChat is focused and always active (open) OR when SystemChat is focused and is not active (closed) due to usage of "Activate chat by 'Enter' pressing" feature. Anyway, the default focus, if no other input field is focused, is always stays on SystemChat either active or non-active.

To check the focus you can use `-focus` command. When the chat is not active but focused, there will be a little red vertical bar `|` in the bottom left corner.

andreakarasho:
I totally agree.

binff04:
Okay, then I will get rid of those, now unnecessary, checkboxes and add a new commit here.

binff04:
Done. Ready to merge.

PR #676, opened 2019-08-24
Build on macOS

binff04:


README.md
```diff
@@ -44,6 +44,42 @@ is installed:
...
+# Building (macOS)
+All the commands should be executed in terminal. All global package installs should be done only if not yet installed.
+
+1. Install Homebrew, a package manager for macOS (if not yet installed): Follow instructions on https://brew.sh/
+
+2. Install Mono (https://www.mono-project.com/):
+`brew install mono`
+
+3. Install .NET Core SDK (https://dotnet.microsoft.com/download):
+`brew install dotnet-sdk`
```
andreakarasho: I dont see any dotnet core command.
Why do you need netcore if you use mono?

binff04: Yeah, you absolutely right. I installed it at the testing stage and it was in my notes since then.

I updated the README file and removed that unnecessary install of .NET Core SDK.

andreakarasho:
Well done

PR #680, opened 2019-08-26
Add macro CloseAllBars that are not anchored

Stallind:


src/Game/Managers/MacroManager.cs
```diff
@@ -1062,6 +1062,26 @@ private int Process(MacroObject macro)
...
+
+                    foreach (var healthbar in healthBarGumps)
+                    {
+                        if (Engine.UI.AnchorManager[healthbar] == null && (healthbar.LocalSerial != World.Player))
+                        {
+                            healthbar.Dispose();
+                        }
+                        else
+                        {
+                            continue;
```
andreakarasho: you dont need to do a continue

src/Game/Managers/MacroManager.cs
```diff
@@ -1062,6 +1062,26 @@ private int Process(MacroObject macro)
                         GameActions.DoubleClick(potion);
 
                     break;
+                    
+                 case MacroType.CloseAllHealthBars:
+
+                    var a = Engine.UI.AnchorManager;
```
andreakarasho: what is it?

andreakarasho:
good boy

PR #681, opened 2019-08-27
Remove bar from SHOW HP under self

Alias-Alan:
Requested feature.


binff04:
By looking at the changes I only see that all the contents of the entire file were removed then re-added. Is that intended? There is no way to see which lines you have changed exactly...

SaschaKP:
he changed line 167 in new, that was some line before on original file.

binff04:
Where did you see that? GitHub here shows it like I described.

SaschaKP:
just copy whole change, overwrite locally and do a diff, or just download the diff and patch locally, on changes you will see that there are a couple of removed newlines, and the only change is in 167 on the new modified file.

binff04:
Didn't expect to make so many manual steps to just find the difference in one line 🤣

Alias-Alan:
Line 165-167
///
//if (mode >= 1) <-- Default
//{	                
//To remove bar under self 
if (mode >= 1 && mobile != World.Player) <-- change to  remove under player.
/////

Sorry for confusion was only making this change for a couple who asked for it in discord/ personally I  think it's a debatable push. The opposition is that they use their healthbar to see their stats and do not need a bar under them, whereas I don't see an issue with it as it's so small so my original file has it in there  notated but not being used. I also removed the my bars pushes as I had a few errors still to fix.

binff04:
That's okay.

I also concerned about this change: first of all, in my humble opinion, there is nothing wrong with the healthline under your own character. I even use it as my main source of information about char's health. Secondly, it should probably be a configurable option then, so some players will decide to use it, and some will be fine with the line under their characters.

Alias-Alan:
As more and more people speak on it, it seems there is only 2-3 that have spoken out saying it's a wanted change, the option isn't even warranted at that point imo.

andreakarasho:
honestly i dont see why i shuld remove the bar under my character

PR #689, opened 2019-09-11
Feature/party invite gump.  Options/General/Show gump for party invite

Stallind:
https://gyazo.com/1d8c395a18a6ccbf1937eeaf1cd6c322


andreakarasho:
nvm, my bad

PR #690, opened 2019-09-11
Update HealthBarGump.cs

Alias-Alan:
Converts old health bars to new updated custom bars (thinner, sleeker, more modernized to run with updated client)


andreakarasho:
I dont want to remove the old healthbar.
Make another class and link it with an option like: "Use new healthbar style".

Alias-Alan:
If any one has suggestions on this, I've yet to toy with this realm of c# yet. Pm me.

PR #724, opened 2019-09-24
Turn delay value

roxya:
image
image

Jaedan said so


binff04:
@andreakarasho btw, what is the purpose of `TURN_DELAY_FAST = 45;`? I tried to search the entire code base and I can't find any places where it is or was used.

PR #725, opened 2019-09-24
ScrollFlag scroll speed increased from 5 to 61

Hezkore:
This makes the shop gump much easier to navigate with the mouse wheel, jumping about 1 item per step.

### DIFF
--- a/src/Game/UI/Controls/ScrollFlag.cs
+++ b/src/Game/UI/Controls/ScrollFlag.cs
@@ -120,7 +120,7 @@ namespace ClassicUO.Game.UI.Controls
             }
         }
 
-        public int ScrollStep { get; set; } = 5;
+        public int ScrollStep { get; set; } = 61;
 
         bool IScrollBar.Contains(int x, int y)
         {


andreakarasho:
61? :| 
looks really high

roxya:
I tested and 61 is very usable for me, things don't fly off too fast. But maybe depends on mouse settings? Pretty sure my wheel is all standard.

Hezkore:
The original value of 5 is painfully slow:


A value of 61 jumps about 1 item in the list:

andreakarasho:
I see a difference between the `ScrollStep` property in ScrollFlag (=5) and ScrollBar (=15).
Probably i missed the '1' char, so i think it's much better now. Sorry guys but 61 looks so high imho

Hezkore:
I'm basing this on other modern games, and Windows itself.
Any scroll area in Windows will usually scroll several items down.
So I'd argue that, if anything, a value of 61 which jumps one item down, might be a bit too slow compared the norm today.



But oh well.

andreakarasho:
Scrolls doesn't know the height of each item.
Often that value is different for each gump. 61 is a value closed to the shopgump only because there are lots of items.

So my idea is to calculate a value with this formula: 

`Scroll.Step = Scroll.MaxValue / 15`

Hezkore:
Yeah I had that idea as well, but I'm afraid that might be confusing.
Like if you're used to the shop gump behaving a certain way, but then you'd open another shop with far more items and all of the sudden you have to relearn the scroll speed.
I think the best would be to set speed individually for each gump.

andreakarasho:
No because a shop with an high Scrollstep value might have 3 items inside and another gump with a low scroll step 1000.
So the problem remains.
I can increase the step a bit and figure out if it is better a fixed value or use the formula

PR #726, opened 2019-09-24
Added option hold shift for quantity

Hezkore:
This adds the option "Hold Shift for item quantity" under the "General" tab.
It basically inverts the Shift + Drag item behaviour, making it so that you don't have to hold down shift every time you want to move an entire stack of items.
As someone who plays modern games, this is a must have option.

ClassicUO_2019-09-24_23-54-49

### DIFF
--- a/src/Game/UI/Gumps/OptionsGump.cs
+++ b/src/Game/UI/Gumps/OptionsGump.cs
@@ -116,7 +116,7 @@ namespace ClassicUO.Game.UI.Gumps
         private ColorBox _poisonColorPickerBox, _paralyzedColorPickerBox, _invulnerableColorPickerBox;
         private TextBox _spellFormatBox;
         private Checkbox _useStandardSkillsGump, _showMobileNameIncoming, _showCorpseNameIncoming;
-        private Checkbox _holdShiftForContext, _reduceFPSWhenInactive, _sallosEasyGrab, _partyInviteGump;
+        private Checkbox _holdShiftForContext, _holdShiftForQuantity, _reduceFPSWhenInactive, _sallosEasyGrab, _partyInviteGump;
 
         //VendorGump Size Option
         private ArrowNumbersTextBox _vendorGumpSize;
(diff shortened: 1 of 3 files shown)


binff04:
This is a perfect addition to already great client. Thank you, @Hezkore. I hope this will be merged soon. 👍

src/Game/UI/Gumps/OptionsGump.cs
```diff
@@ -1307,6 +1308,7 @@ private void SetDefault()
                     _holdDownKeyTab.IsChecked = true;
                     _holdDownKeyAlt.IsChecked = true;
                     _holdShiftForContext.IsChecked = false;
+                    _holdShiftForQuantity.IsChecked = false;
```
Stallind, line 1311: Should be true by default to not mess everyone up from how it used to be.

binff04: Why it should be `true` by default? It should be `false` by default to not apply the new behavior until it's checked.

Alias-Alan: _holdShiftForQuantity.IsChecked = false; will have it move quantity without shift; if it is true (checkbox is checked) it will be replicating the default behavior because it will cause the user to use shift to move quantity.

binff04: Not sure I get what you are talking about @Syrupz-UO...
Let's check what's going on in the `src/Game/Scenes/GameSceneDragAndDropHandler.cs` file on the line 68:

Assuming that `_isShiftDown` will be `true` when the actual Shift key is down, then the stack split gump will be shown when `Engine.Profile.Current.HoldShiftForQuantity == _isShiftDown`.

If `Engine.Profile.Current.HoldShiftForQuantity` is `false` which is by default, then stack split gump will be shown if `_isShiftDown` is also `false` because it is a check if two variables are the same, which is `false == false` is always `true`.

Then, let's assume that `Engine.Profile.Current.HoldShiftForQuantity` will be `true`, then if we don't hold the shift key (`_isShiftDown` is `false`) the check `Engine.Profile.Current.HoldShiftForQuantity == _isShiftDown` will be like `true == false` which is `false`, so the split stack gump will not be shown, BUT (!) if `_isShiftDown` is `true`, then `true == true` which is `true` and the split stack gump will be shown.

Everything looks correct to me.

Stallind: My bad, you're right binff04. I was thrown off by how the checkbox text was formulated. And I guess Syrupz was too :)

Stallind (approved):

src/Game/UI/Gumps/OptionsGump.cs
Alias-Alan: I think it's the confusion of it being called Hold Shift for Quantity. As you say, if we assume it is "true" then you SHOULD hold the shift key for quantity.  However, as it is currently you hold shift to move the whole quantity and do not hold shift to select quantity.  You could rename it to "UseShiftToSelectQuantity" or something of such nature.

binff04: @Syrupz-UO, you can not "move" the quantity. You can move the entire stack. "Quantity" means the number of items. You probably miss the point of the label text: "hold shift for item quantity" means that if you want to bring up that quantity split selector, then you need to hold Shift. It doesn't sound like "hold shift to move the entire stack" or "hold shift to move item quantity"...

Anyway, if it's so confusing, then we can probably ask the author to rename it into something like "Hold Shift for item quantity selector" or "Hold Shift to split item stack"

binff04 (changes requested):
Can we change the label text to something like "Hold Shift to split stack of items", so it will sound more like "You need to hold down the Shift key to split that damn stack of items".

src/Game/UI/Gumps/OptionsGump.cs
Alias-Alan: Yes, just confused with the wording is all.

andreakarasho:
yes, made me confused :D

binff04:
> yes, made me confused :D

Yeah, a little bit :) So, can we merge it and I will update the label then?

andreakarasho:
yes please

binff04:
> yes please

Done: https://github.com/andreakarasho/ClassicUO/pull/746

Issue #718, opened 2019-09-22
Anchoring a bit fiddly compared with the original client


roxya:
CUO anchoring only works when the mouse cursor is over the gump you want to anchor to. You get used to it, but it's not intuitive.
![anchor0](https://user-images.githubusercontent.com/7057924/65390147-74ea1500-dd54-11e9-8583-bd1f60cec7af.gif)

I've got two working alternatives.

**Original client anchoring style** - activate as soon as the gumps overlap
![anchor1](https://user-images.githubusercontent.com/7057924/65390185-bed2fb00-dd54-11e9-801b-26bbcad8d799.gif)

**Modern style** - show potential anchor location
![anchor2](https://user-images.githubusercontent.com/7057924/65390195-cd211700-dd54-11e9-8568-6c5d2b810b53.gif)

Just wondering if anyone has comments about this.

Hoogie1980:
I find it hard to anchor health bar gumps together in CUO for this exact reason.

andreakarasho:
the anchoring with preview is so neat

Belialzao:
Anchor location preview looks neat!

My only comment is about UNanchoring.
in the original client you can do so by holding alt + 2 click and drag the healthbar
the fact that you have to click twice looks like a bug,
but imo its better than having to click a tiny lock icon while still holding alt

My suggestion would be to unanchor if you click and drag while holding alt

the tiny locks can still show up when alt is held, but would be nice if it would disappear from the bar (or spell/skill icon) you're dragging out of the stack.

roxya:
@luuutz I am struggling to see any downside to unanchoring with alt+drag

Hezkore:
Imo. the anchoring should work like the **Original client anchoring style** by default, with an option for **Modern style** anchoring.
Then alt + drag for unanchoring.

andreakarasho:
We cannot make everything as an option because most of the users hate all of these settings.
Modernize the client is a must, so I don’t see why not implement the new one. 
It behaves like the old one, but more intuitive

binff04:
> We cannot make everything as an option because most of the users hate all of these settings.
> Modernize the client is a must, so I don’t see why not implement the new one.
> It behaves like the old one, but more intuitive

[...]

PR #727, opened 2019-09-24
Anchoring improvements

roxya:
As per https://github.com/andreakarasho/ClassicUO/issues/718

**New anchoring with preview**
Anchoring preview activates as soon as the gump overlaps another compatible gump, rather than relying on mouse position. It's easier and feels intuitive.
anchoring

**Alt+drag detaches from group** (don't need to click the lock icon)
anchor_and_detach

None of this is optional.


Alias-Alan:
could go a little lighter on the transparency... other than that I'm sure most will enjoy

PR #729, opened 2019-09-25
Fix tradegump from hidden mobiles

Stallind:


src/Network/PacketHandlers.cs
```diff
@@ -1053,13 +1053,15 @@ private static void UpdateContainedItem(Packet p)
             if (containerSerial.IsMobile)
             {
                 Mobile m = World.Mobiles.Get(containerSerial);
-                Item secureBox = m.GetSecureTradeBox();
-
-                if (secureBox != null)
+                if (m != null)
```
canterberry: 🎨 For a cleaner diff, rather than the defensive null check in a conditional here, "maybe" syntax could be a tidier change:

```cs
Item secureBox = m?.GetSecureTradeBox();
```

(similar to line 1049 just above this changeset)

Stallind: I prefer the readability from writing it all, but I will be consistent to the project and change it :)

canterberry (approved):
👍 LGTM

PR #731, opened 2019-09-25
HBG (Updated all files)

Alias-Alan:
CustomHealthBars
Still no idea on commit history


roxya:
You are missing CustomBarsToggled in the profile, so it will not build.

roxya:
If I pull healthbars from the NameOverheadGump, it is always the old style

roxya:
If I switch to the large status bar and back to a small one, my healthbar becomes the old style one

roxya:
It's supposed to use the gump graphics for bars? Looks like a strange mix of old and new
image

Alias-Alan:
That does look strange
image

Alias-Alan:
I've added checks to every file that generates a healthgumpbar -the ones I missed were 
[
packethandlers.cs
partymanager.cs
partygumpadvanced.cs
statusgump.cs
nameoverheadgump.cs
paperdollgump.cs
]
All of which have been added and throw no exceptions, fixes the paperdoll gump produced, minimizing and maximizing is fixed, name over head is fixed, as well as party handling.

Alias-Alan:
Fixing party issue

PR #746, opened 2019-10-01
Rename option

binff04:
Change the label and property field name for the new "Hold Shift for quantity" option.


Hezkore (commented):
Looks good.

PR #747, opened 2019-10-05
Custom Health Bars

Alias-Alan:
-Created (Experimental) Option for these health bars.
-This version works with statusgump, paperdoll, partybars, and nameoverhead gump.
-This version handles a default gamescene placement of healthbar upon login.
-This version handles OnDragSelect to pull custom bars.
-This version handles renaming pets, pull it's bar, click the name and enter new one.
-This version also includes modifying CloseAllHealthBars to include custom bars as well.
CustomHealthBars


Hoogie1980:
Sweet is this live now?

src/Game/Managers/MacroManager.cs
```diff
@@ -1066,6 +1066,7 @@ private int Process(MacroObject macro)
                  case MacroType.CloseAllHealthBars:
 
                     var healthBarGumps = Engine.UI.Gumps.OfType<HealthBarGump>();
```
Stallind: Could tidy this block up by just having this: 
```
var healthBarGumps = Engine.UI.Gumps.Where(g => g is HealthBarGump || g is HealthBarGumpCustom);
                    foreach (var healthbar in healthBarGumps)
                        {
                            if (Engine.UI.AnchorManager[healthbar] == null && (healthbar.LocalSerial != World.Player))
                            {
                                healthbar.Dispose();
                            }
                        }
                    break;
```

Alias-Alan: I originally did this but decided to use a different anchor name for customhealthbars because last pr someone said mentioned anchor group name thus you have 2 groups

roxya:
You're missing OnMouseExit in HealthBarGumpCustom, so when you hover a bar and the character is highlighted, it stays highlighted even when you move the cursor away.

Syrupz-UO:
Good catch, will fix shortly.

Alias-Alan:
Added OnMouseExit to solve issue stated above.

src/Game/Managers/MacroManager.cs
Stallind: Your replay makes no sense. The code above is doing exactly what the code in place is doing, just less code 🍡

src/Game/Managers/MacroManager.cs
```diff
@@ -1066,6 +1066,7 @@ private int Process(MacroObject macro)
...
+                    var customBarGumps = Engine.UI.Gumps.OfType<HealthBarGumpCustom>();
```
Stallind: > Could tidy this block up by just having this:
> 
> ```
> var healthBarGumps = Engine.UI.Gumps.Where(g => g is HealthBarGump || g is HealthBarGumpCustom);
>                     foreach (var healthbar in healthBarGumps)
>                         {
>                             if (Engine.UI.AnchorManager[healthbar] == null && (healthbar.LocalSerial != World.Player))
>                             {
>                                 healthbar.Dispose();
>                             }
>                         }
>                     break;
> ```

src/Game/Managers/MacroManager.cs
Alias-Alan: I am all for making the code cleaner, however my response above was wrong as I was on mobile and not sitting here looking at the code itself. Your way of handling the gumps via Gumps.Where throws an argument because it's trying to convert your control 'g is a or b' to an AnchorableGump. This is why I had to use Gumps.OfType for each healthbartype. If you can find another solution to this by using less code let me know, I learn as I go :)

Stallind: Oh, try 
`
var healthBarGumps = Engine.UI.Gumps.OfType<AnchorableGump>().Where(g => g is HealthBarGump || g is HealthBarGumpCustom);
`

Alias-Alan: That did the trick, revision pushed to this PR. Thanks!

Alias-Alan:
For some unknown reason this pulled the AutoOpenSecretDoors PR that was on a separate branch, that will be the next PR after this but per this PR those 5 files have been reverted back to the original and I will reload from local to that branch for that PR. I'm not sure why this issue persists with commits showing from other branches.

src/Game/Scenes/GameSceneInputHandler.cs
```diff
@@ -823,12 +957,13 @@ internal override void OnKeyUp(SDL.SDL_KeyboardEvent e)
 
             _useObjectHandles = isctrl && isshift;
 
-            if (e.keysym.sym == SDL.SDL_Keycode.SDLK_TAB && !Engine.Profile.Current.DisableTabBtn)
+            if (e.keysym.sym == SDL.SDL_Keycode.SDLK_TAB /*&& !Engine.Profile.Current.DisableTabBtn*/)
```
andreakarasho: Here reverted fix

Alias-Alan: Fixed both instances as to not overwrite your recent commits.

Alias-Alan:
Making improvements to this/ Will resubmit when finished.

PR #749, opened 2019-10-08
Settings system update (#2)

binff04:
This update brings an ability to use different `settings.json` files when starting ClassicUO.exe with the `-settings <filepath>` option.

This change is necessary in preparation for ClassicUO Launcher that will allow to create different profiles (settings files) for different shards and start ClassicUO with selected profile.

For those who want to utilize this new change without the launcher: you can just start ClassicUO with the settings option and filepath of your settings file, for example:

`ClassicUO.exe -settings "C:\Users\username\path\to\settings_1.json"`

Or if you're running it via mono on macOS or Linux:

`./ClassicUO-mono.sh -settings "/Users/username/path/to/settings_1.json"`

### DIFF
--- a/src/Configuration/Profile.cs
+++ b/src/Configuration/Profile.cs
@@ -317,14 +317,14 @@ namespace ClassicUO.Configuration
 
             Log.Message(LogTypes.Trace, $"Saving path:\t\t{path}");
 
-            // save settings.json
+            // Save profile settings
             ConfigurationResolver.Save(this, Path.Combine(path, "profile.json"), new JsonSerializerSettings
             {
                 TypeNameHandling = TypeNameHandling.All,
                 MetadataPropertyHandling = MetadataPropertyHandling.ReadAhead
             });
 
-            // save gumps.bin
+            // Save opened gumps
             SaveGumps(path, gumps);
 
             Log.Message(LogTypes.Trace, "Saving done!");
--- a/src/Configuration/ProfileManager.cs
+++ b/src/Configuration/ProfileManager.cs
@@ -41,7 +41,7 @@ namespace ClassicUO.Configuration
 
 
 
-            // this is a temporary patch to change the profile settings name safety
+            // FIXME: this is a temporary patch to change the profile settings name safety
             string fileToLoad = Path.Combine(path, "settings.json");
             string newPath = Path.Combine(path, "profile.json");
 
@@ -53,7 +53,7 @@ namespace ClassicUO.Configuration
                     {
                         File.Delete(fileToLoad);
                     }
-                    catch (Exception ex)
+                    catch (Exception)
                     {
                         Log.Message(LogTypes.Warning, $"Failed to delete file: '{fileToLoad}'");
                     }
--- a/src/Configuration/Settings.cs
+++ b/src/Configuration/Settings.cs
@@ -101,6 +101,20 @@ namespace ClassicUO.Configuration
 
 
         public const string SETTINGS_FILENAME = "settings.json";
+        public static string CustomSettingsFilepath = null;
+
+        public static string GetSettingsFilepath()
+        {
+            if (CustomSettingsFilepath != null)
+            {
+                if (Path.IsPathRooted(CustomSettingsFilepath))
+                    return CustomSettingsFilepath;
+                else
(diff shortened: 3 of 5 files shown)


andreakarasho:
I want maintain the behaviour and nature of the single settings.json file.
This in prevision of the fact that the an user can skip the launcher or runs cuo from other launcher [like Outlands does].

Launcher will store all profiles in a single file.
It will run a CUO instance with arguments then CUO will override its settings.json with the arguments values.

So probably what you thought will be an unused behaviour

binff04:
@andreakarasho you probably didn't check the code and what it does. If you start ClassicUO without `-settings` option, it does what it always did -- use `settings.json` in the same directory as `ClassicUO.exe` by default and writes the main settings there as it did. Everything stays the same as before.

But, if you want to try different set of settings, or to maintain settings for different accounts+shards in different settings files, you can utilize `-settings <filepath>` option to tell CUO which settings file it should use and where it should write its settings. The default `settings.json` will be unmodified in this case.

Honestly, I believe that this update brings a lot of additional flexibility.

Also, I would like to disagree about storing profiles in a single file: first of all, it will be easier to manage each settings file separately. This is as we code: we don't store all classes in a single file. Also, this is an additional protection against file corruption: if only one file is used, then it's better chances to lose all profiles at once.

andreakarasho:
Honestly i closed this pr accidentally so don’t be angry :D I used to wait a reply before closing or not.

So about the pr: yeah it gives more flexibility, but with a launcher you can manage all these stuff directly. Nothing stops you to make more than 1 profile per server.
So this pr might be a surplus.

I totally agree with you about the single file instead.

binff04:
I'm not angry, no worries ;) Just suing my experience in other projects and want to bring some value for the project.

So, let me describe my idea of the launcher and how I see it working: when you first open the launcher, it will only have the "Default" profile which is obviously our default `./settings.json` (personally, I don't like to call it "profile" because it messes with shard+character profiles). You can also create additional "profiles" by clicking "Add profile" button and filling out such fields like "Profile name", "UO Directory", "Client version" (something else?) and when you're done, the Launcher will save it in its own settings file. We can use a sub-directory to store custom profiles in the CUO main directory. "Settings", for example. So, let's assume the new profile is stored as `./Settings/my_uoo_acc1.json`, then when you start CUO with this profile selected, we basically run `ClassicUO.exe -settings Settings/my_uoo_acc1.json` and that's it, CUO will load main settings from that file, and, what's most importantly, will write main settings in that file as well, and will not touch default `settings.json` in the main directory.

Any thoughts?

binff04:
I'm not angry, no worries ;) Just using my experience obtained in other projects to bring some value for the ClassicUO project.

So, let me describe my idea of the launcher and how I see it working: when you first open the launcher, it will only have the "Default" profile which is obviously our default `./settings.json` (personally, I don't like to call it "profile" because it messes with shard+character profiles). You can also create additional "profiles" by clicking "Add profile" button and filling out such fields like "Profile name", "UO Directory", "Client version", "Shard Connection Address and Port", (something else?) and when you're done, the Launcher will save it in its own settings file. We can use a sub-directory to store custom profiles in the CUO main directory. "Settings", for example. So, let's assume the new profile is stored as `./Settings/my_uoo_acc1.json`, then when you start CUO with this profile selected, we basically run `ClassicUO.exe -settings Settings/my_uoo_acc1.json` and that's it, CUO will load main settings from that file, and, what's most importantly, will write main settings in that file as well, and will not touch default `settings.json` in the main directory.

So, that way, such projects like UOO will be able to launch not ClassicUO directly, but its Launcher where the end user will be able to seamlessly manage his or her profiles as well as use already used the default one.

This all brings not only a lot of flexibility for everyone using Launcher or manually managing JSON files, but also maintains backward compatibility.

Any thoughts?

andreakarasho:
Ok now makes sense.
Approved!

binff04:
Thanks!

Btw, there is a hidden feature: you can run ClassicUO with non-existent path & file given to `-settings` option and it will create missing path (directories) and the file itself, so if you do `ClassicUO.exe -settings MySettings/my_profile.json` it will automatically create both `MySettings` directory (if not yet created) and `my_profile.json` file in there, so you don't need to worry about creating everything manually -- CUO does that for you, throws an error message with the full path of the created settings file that you need to edit.

PR #752, opened 2019-10-08
Add Select{Next|Previous|Nearest} macros

binff04:
If you play without Razor on macOS or Linux, this is something that you probably miss a lot.

SelectNext: iterates through selected type of mobiles in a 12-tile radius (from top to bottom of the list);
SelectPrevious: iterates through selected type of mobiles in a 12-tile radius (from bottom to top of the list);
SelectNearest: iterates through selected type of mobiles in a 12-tile radius (from top to bottom of the list but only among the nearest mobiles);

### DIFF
--- a/src/Game/Managers/TargetManager.cs
+++ b/src/Game/Managers/TargetManager.cs
@@ -284,8 +284,13 @@ namespace ClassicUO.Game.Managers
             ClearTargetingWithoutTargetCancelPacket();
         }
 
-        public static bool IsMobileSelectableAsTarget(Mobile mobile, int type)
+        public static bool IsMobileSelectableAsTarget(Serial serial, int type)
         {
+            Mobile mobile = World.Mobiles.Get(serial);
+
+            if (mobile == null)
+                return false;
+
             if (mobile == World.Player)
                 return false;
 
@@ -327,8 +332,13 @@ namespace ClassicUO.Game.Managers
             return true;
         }
 
-        public static void SetLastTarget(Mobile target)
+        public static void SetLastTarget(Serial serial)
         {
+            Mobile target = World.Mobiles.Get(serial);
+
+            if (target == null)
+                return;
+
             GameActions.MessageOverhead($"Target: {target.Name}", Notoriety.GetHue(target.NotorietyFlag), World.Player);
             Engine.UI.RemoveTargetLineGump(TargetManager.LastTarget);
             Engine.UI.RemoveTargetLineGump(TargetManager.LastAttack);
@@ -344,7 +354,10 @@ namespace ClassicUO.Game.Managers
 
             foreach (Mobile mobile in World.Mobiles)
             {
-                if (!IsMobileSelectableAsTarget(mobile, type))
+                if (mobile == null)
+                    continue;
+
+                if (!IsMobileSelectableAsTarget(mobile.Serial, type))
                     continue;
 
                 if (TargetManager.LastTarget != null && TargetManager.LastTarget == mobile.Serial)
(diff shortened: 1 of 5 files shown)


src/Game/Managers/TargetManager.cs
```diff
@@ -283,5 +283,176 @@ public static void SendMultiTarget(ushort x, ushort y, sbyte z)
             MultiTargetInfo = null;
             ClearTargetingWithoutTargetCancelPacket();
         }
+
+        public static bool IsMobileSelectableAsTarget(Mobile mobile, int type)
+        {
```
andreakarasho, line 288: It’s a good practice adding a null check for the mobile var or pass a Serial as argument and search the mobile with World.Mobiles dictionary then add a null check.
This is to avoid Ghost reference and unexpected crahes if we want call this method out of here for any reason.
I prefer the second option honestly.

src/Game/Managers/TargetManager.cs
```diff
@@ -283,5 +283,176 @@ public static void SendMultiTarget(ushort x, ushort y, sbyte z)
...
+                // 4 - Mobile (any mobiles)
+                // No need to check anything here
+            }
+
+            return true;
+        }
+
+        public static void SetLastTarget(Mobile target)
+        {
+            GameActions.MessageOverhead($"Target: {target.Name}", Notoriety.GetHue(target.NotorietyFlag), World.Player);
```
andreakarasho, line 342: The same as above

roxya:
I think it needs to respect negotiated features and only allow Nearest if the server allows it. And personally I would not set a 12 tile limit (in some cases gives you an advantage, in others it's really difficult to preselect a target before it's in range)

But this is sorely needed for mac users! May even be nice for everyone else since we can make more reliable target indicators than Razor.

binff04:
> I think it needs to respect negotiated features and only allow Nearest if the server allows it.

These are the default macros like in original client, why they should be negotiated? Anyway, does ClassicUO support negotiating features?

> But this is sorely needed for mac users! May even be nice for everyone else since we can make more reliable target indicators than Razor.

Yeah, for sure! In this first version I implemented the base so anyone playing on Macs will be able to already use normal targeting functions available to users who play on Windows. We can definitely do a better indication for selecting new target, but right now it is the same one that Razor uses.

andreakarasho:
Honestly i dont remember if standard client has these options, i think yes and them are not affected by negotation. 
Anyone could verify?

roxya:
Nearest exists but I never tried to see how it works.

I suppose my Outlands side is talking, since it disables closest/nearest targeting options where it can (assistants which respect negotiation), and I expect this feature would become stronger when someone inevitably asks for/does a PR for humanoid/notoriety filters.

But it's a standard feature yeah, I can't argue the fact.

gaechti:
Stratics has em listed, but no explanation what they exactly do
https://uo2.stratics.com/essays-and-guides/client-user-guide-classic/the-options-menu/

I think its a bit different, select nearest from cc and target nearest from e.g. Razor

binff04:
Razor does "closest" selection differently from my implementation: it only selects one mobile, and does not iterates through all the closest.

src/Game/Managers/TargetManager.cs
binff04: I totally agree. I made the changes, so both methods now expect Serial instead of Mobile, get mobile from World.Mobiles and check for null. No place for mistake.

src/Game/Managers/TargetManager.cs
binff04: Yeah, same change.

andreakarasho:
Tested standard client.
It is a legit behaviour. (Outlands too)

Issue #748, opened 2019-10-06
Support for LBR


JakubLinhart:
When trying to run ClassicUO on UOErebor (based on LBR) some files are missing from the installation - at least Equipconv.def. It seems like there are some issues with protocol also.

I would like to take closer look to the issue if these changes are welcome in ClassicUO.

andreakarasho:
I added some file.exists check during the file load.
Can you tell me if it has been fixed ?

PR #758, opened 2019-10-08
Lbr

JakubLinhart:
#748 Thanks for the immediate fix. It was impressively fast! There are other small changes that are required in order to be able to login on a shard powered by ancient Sphere 0.99.


andreakarasho:
neat!

PR #760, opened 2019-10-11
Dragged item size uses ContainerScale

Hezkore:
Dragging items will now use `Engine.UI.ContainerScale`

Without this fix:

With this fix:


Hezkore:
Something I also noticed while digging through the code is that items that are `IsStackable` and `Amount > 1` uses a hardcoded offset value of 5, 5 when drawing the second item, which is _NOT_ scaled with the `Engine.UI.ContainerScale` value.
So that should really be addressed as well.

PR #765, opened 2019-10-13
Auto loot no longer shows empty corpses

Hezkore:
This pull request makes auto loot no longer show empty corpse containers.
It does this by checking what corpses are manually double clicked and which are opened by auto loot.
If it's a manual corpse loot; empty containers are okay.
If it's an auto corpse loot; empty containers are disposed.
That means, unlike the Grid Loot gump, that you're still able to open empty corpses and put items into them, but you won't be spammed by auto loot showing empty containers.
This will only be applied to corpse containers, which means it should be fairly safe and not close any important gumps or containers.

### DIFF
--- a/src/Game/GameCursor.cs
+++ b/src/Game/GameCursor.cs
@@ -357,19 +357,24 @@ namespace ClassicUO.Game
            
             if (_itemHold != null && _itemHold.Enabled && !_itemHold.Dropped)
             {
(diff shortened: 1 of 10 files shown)


Hezkore:
And as an added bonus, I can report that `_isCorspeContainer` is misspelled. (not my doing)

PR #767, opened 2019-10-15
Close empty auto loot corpses

Hezkore:
Here's my second attempt at "closing empty auto loot corpses".
This one doesn't mess with network packets.
It also won't close any "unknown" corpse containers that happens to be empty.

What this will do is make the "Auto open corpses" feature not open any empty corpses.
Walking over a pile of dead bodies will no longer spam you with empty containers!
But you'll still be able to open the corpse manually and place items into the empty corpse (unlike with the Grid Loot system).

The logic behind this goes:
When "Auto open corpses" tries to open a corpse it will push the serial of the corpse to `AutoOpenedCorpses` (formerly `OpenedCorpses`).
When the user clicks a corpse it pushes the serial of the corpse into `ManualOpenedCorpses`.
And when a container is built, and it's a corpse; it will check `ManualOpenedCorpses` first and see if the user opened this corpse, if so it will never close it.
However! If it was not opened by the user, **and** it exist in `AutoOpenedCorpses`; then the `_closeIfEmpty` boolean for the container will be set to `true`.
If `_closeIfEmpty` is `true` during `Update`, and if the item count reaches zero; then the corpse will be closed, unless it has previously had at least one item.


andreakarasho:
Sometime server can send the corpse container amount equals to 0. 
When you open a corpse, server sends the total amount of items within the corpse, so you receive the real items amount AFTER corpse opening.

Probably this behaviour is not predicatble.

Hezkore:
Hmm, I've been using this for about 5 days during high ping without any issues so far.

I believe the Grid Loot gump does a similar thing.
During `RedrawItems` the Grid Loot gump checks the item count, and if the item count is zero; it closes the gump.
I do the same but during `Update`.
The difference between the Grid Loot gump and mine is that mine won't close if it's opened manually by the player.
So it should really be "safer" than the Grid Loot gump.

I can continue to look for issues with this method though.

src/Game/UI/Gumps/ContainerGump.cs
```diff
@@ -193,6 +201,14 @@ public override void Update(double totalMS, double frameMS)
                 return;
             }
 
+            if (item.Items.Count() > 0)
```
andreakarasho: .Count() is very expensive in terms of performance.
Use .Count property instead.

Hezkore: Good catch!
I will fix once I get some time over.
Or feel free to edit yourself. :)

Hezkore: Fixed

PR #772, opened 2019-10-17
Esc cancels chat.

JakubLinhart:
If I start writing to chat and I need use macro suddenly, I want to close the chat immediately without sending unfinished message.


andreakarasho:
I can't accept this PR for these reasons:
- ESC can be overrided by user, so have an hardcoded hotkey is not the best option
- You don't send unfinished msg necessary if you don't press ENTER
- i'm planning to implement a function as CTRL+A to select text then press BACKSPACE as standard textbox behaviour [as regular games do]. This should be seen as an hardcorded hotkey too, but as you know CTRL + A is the common one.

PR #777, opened 2019-10-19
Client 306m ignores bodyconv.def 

JakubLinhart:
Client 306m ignores bodyconv.def without this change. 

IndexAnimation.CalculateOffset does index calculation for a new graphics based on a type of an original graphics (line 1866) which leads to an overly large offset value which is filtered out (line 506) which leads to ignoring bodyconv.def content.

Although it works perfectly after the change for 306m, I have no idea what are all consequences of this change and if it breaks anything else.

### DIFF
--- a/src/IO/Resources/AnimationsLoader.cs
+++ b/src/IO/Resources/AnimationsLoader.cs
@@ -505,6 +505,7 @@ namespace ClassicUO.IO.Resources
 
                                 if (addressOffset < currentIdxFile.Length)
                                 {
+                                    DataIndex[index].Type = CalculateTypeByGraphic(realAnimID);
                                     DataIndex[index].MountedHeightOffset = mountedHeightOffset;
                                     DataIndex[index].GraphicConversion = (ushort)(realAnimID | 0x8000);
                                     DataIndex[index].FileIndex = (byte)animFile;
@@ -800,7 +801,7 @@ namespace ClassicUO.IO.Resources
             return (uint)(((graphic - 200) * 65 + 22000) * UnsafeMemoryManager.SizeOf<AnimIdxBlock>());
         }
 
-        private ANIMATION_GROUPS_TYPE CalculateTypeByGraphic(ushort graphic)
+        internal static ANIMATION_GROUPS_TYPE CalculateTypeByGraphic(ushort graphic)
         {
             return graphic < 200 ? ANIMATION_GROUPS_TYPE.MONSTER : graphic < 400 ? ANIMATION_GROUPS_TYPE.ANIMAL : ANIMATION_GROUPS_TYPE.HUMAN;
         }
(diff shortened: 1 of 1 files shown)


src/IO/Resources/AnimationsLoader.cs
```diff
@@ -1863,7 +1864,7 @@ public long CalculateOffset(ushort graphic, out int groupCount)
 
             ANIMATION_GROUPS group = ANIMATION_GROUPS.AG_NONE;
 
-            switch (Type)
+            switch (AnimationsLoader.CalculateTypeByGraphic(graphic))
```
andreakarasho: Please just respect the style: FileManager.Animations.CalculateTypeByGraphic()

src/IO/Resources/AnimationsLoader.cs
```diff
@@ -800,7 +801,7 @@ public static uint CalculateLowGroupOffset(ushort graphic)
             return (uint)(((graphic - 200) * 65 + 22000) * UnsafeMemoryManager.SizeOf<AnimIdxBlock>());
         }
 
-        private ANIMATION_GROUPS_TYPE CalculateTypeByGraphic(ushort graphic)
+        internal static ANIMATION_GROUPS_TYPE CalculateTypeByGraphic(ushort graphic)
```
andreakarasho: static is not needed

andreakarasho (commented):
Honestly i'm not sure about this PR.
We should try with recent clients

andreakarasho:
Update: this fix breaks most of the animations on latest UO

JakubLinhart:
Yes, this is highly suspicious change. I hope for some guidance to get it right :).

I tested it on Outlands files/ServUO using these monsters: Centaur (0x0065), Rotworm (0x02DC), Hydra (0x0109), Minotaur (0x0107) and Golem (0x02F0). All creature's animations are same before and after change.

Please, what was your test setup (client version, server, etc...)?

andreakarasho:
with 0x84 it does not work

JakubLinhart:
The problem happens e.g. with Outlands files when IndexAnimation.Type of realAnimID doesn't match Type of original AnimId.

The problem never happens on older 3.x clients e.g. with UOErebor files. Straightforward fix could be to change IndexAnimation.Type only for older clients.

Type change in CrossUO's CAnimationManager is [constrained](https://github.com/crossuo/crossuo/blob/19047def9cc82dab73e3d7bd617edff410b9cee8/src/Managers/AnimationManager.cpp#L641) for older clients only also:

```
g_Config.ClientVersion < CV_500A 
```

Anyway I don't understand neither CrossUO CAnimationManager/ClassicUO AnimationManager nor correspondence between them.

I would risk breaking clients older than 5.0.0 with this change and fix potential issue when it really bites someone. Such a case could help to understand bodyconv semantics better.

andreakarasho:
This one looks good, we'll see the collatar effects :D

PR #779, opened 2019-10-19
Remove an item from a container when the item falls out of it, otherw…

JakubLinhart:
…ise client doesn't draw the item on ground and is still visible in the container.

### DIFF
--- a/src/Network/PacketHandlers.cs
+++ b/src/Network/PacketHandlers.cs
@@ -589,6 +589,12 @@ namespace ClassicUO.Network
             }
 
             item.LightID = direction;
+            if (item.Container.IsValid)
+            {
+                var cont = World.Get(item.Container);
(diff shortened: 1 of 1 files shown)


andreakarasho:
Server does not send any DeleteObject packet?

JakubLinhart:
Unfortunately old Sphere99 doesn't send any 0x1D at least in two cases: 
 - arrow hits you and falls on ground
 - you rip a human body that contains items, the body disappears and items fall on ground

This line suggests that missing 0x1D is an issue for other emulators/servers as well:

```
item.Container = Serial.INVALID;
```
https://github.com/andreakarasho/ClassicUO/pull/779/commits/417b12a17ba33c15cae2711ee5b4968a107f4a51#diff-7bf161a7d59f79b35f8ef673f3b2ca4fR598

PR #782, opened 2019-10-20
Added BandageGump

Hezkore:
This adds the option "Show gump when using bandages".
It will display a bandage timer below the player (or anywhere you want) whenever bandages are used.

For Outlands it will use the same method as the Info Bar to count down.

For other servers that do not support this method; it will instead work like Razor and count up.


andreakarasho:
Why not add it directly to infogump?

Hezkore:
Personally I feel like the Bandage Gump is, and should be, its own thing.
It works differently than the Info Bar and doesn't depend on the Info Bar in any way.
It even renders separately from the Info Bar.
Code wise I don't care where it goes though, but still find it odd placing it in the Info Bar.

Having said that; maybe the Info Bar should read data from the Bandage Gump...
Seeing as how the Info Bar's BandageTimer will be broken on any other server but Outlands.

In the end I don't mind either way though, as long as I get a Bandage Timer near my character heh.
But I won't have time for a while to do any major changes.

andreakarasho:
Honestly i prefer to see this stuff on InfoBar for 2 reasons:
- to maintain the coherence
- avoid multiple floating minigumps around the screen

binff04:
@andreakarasho personally, I would like to disagree and here is why I think that dedicated Bandage Timer (like in Razor) or Bandage Gump (like in this PR) is a necessity for a dexxer-type of character:

First of all, I will not speak about the realisation in the code itself, but as a feature, the dedicated Bandage Timer/Gump is a must have dexxer-combat-centric addition in my opinion. InfoBar is a great addition but it just basically transforms how we see everything on the Status gump and allows us to choose what we want to see. Also, InfoBar does not support bandaging "started" and "stopped" messages as in the Razor or in this PR, so it only relies on what server sends. In the case of UO:Outlands, yes, we have bandage timer instead of one of the resistance indicators, but what about other shards? Even more: InfoBar is not combat-centric gump. You suppose to have it somewhere on the top/bottom of your screen, but that milliseconds that you spend by looking for bandage timer there may and will negatively impact you in fight. I saw someone used InfoBar with only Bandage Timer on it and moving that little one under the char just for the purpose of having it in the focus. This is the problem that this PR solves and it's optional one that you may have enabled or just keep it disabled.

P.S. Mages actually have their combat-centric feature called "Range indicator", why dexxers can't have their Bandage Timer? Personally, I'm going to include it in my own build (thanks to @Hezkore) ;)

Hezkore:
You can now use `World.Player.BandageTimer.Time` to get the current bandage time.
(Something for the Info Bar to use perhaps?)

I've also done so that the max limit for bandage time (when using the Razor method) is 20 seconds.
This ensures that there's no possible way for the bandage timer to get stuck on the screen forever.
_(only ever happens on servers that don't properly use Clilocs)_

PR #787, opened 2019-10-22
Do not append characters after '\0' which terminates the string.

JakubLinhart:
Older Sphere versions don't initialize strings, so it leaks random memory content.


JakubLinhart:
Too fast and broken fix.

PR #788, opened 2019-10-22
Added option Show Buff duration

Hezkore:
This new option located under "Combat / Spells" will show a countdown timer on every buff in your buff bar.

ClassicUO_2019-10-22_23-05-44

Example:

It uses a Unicode font, which I've been told can be an issue, but I'm not sure why...


andreakarasho:
just a note:
RenderedText MUST be disposed on virtual Dispose() method or the resource will remain appended

Hezkore:
Good catch.

PR #789, opened 2019-10-22
fix missing newtonsoft json

roxya:


binff04:
This change totally breaks building process on macOS.

Hezkore:
This was always part of the project though.
It was removed in https://github.com/andreakarasho/ClassicUO/commit/c6a11dfc8ef9d05aa2322a9aa81e9fccbd9f552d and that broke some things, and now it's added back.

binff04:
The problem is that there is no such path on macOS like `..\packages\Newtonsoft.Json.12.0.2\lib\net45\Newtonsoft.Json.dll`.

paket installs it at `..\packages\Newtonsoft.Json\lib\net45\Newtonsoft.Json.dll`

PR #790, opened 2019-10-22
Fix for sound volume

Hezkore:
This sets `SOUND_DELTA` to 250 instead of 1000.
The value 250 may not be as nice and round as 1000, but it will make the volume peak at 90% instead of below 50%.
200 is rounder and nice, and seems to peak at about 98% volume, but I feel like that is cutting it close.

### DIFF
--- a/src/Game/Constants.cs
+++ b/src/Game/Constants.cs
@@ -116,7 +116,7 @@ namespace ClassicUO.Game
 
         public const int DEATH_SCREEN_TIMER = 1500;
 
-        public const float SOUND_DELTA = 1000f;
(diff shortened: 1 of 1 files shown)


binff04:
Finally, normal sound level! Thank you @Hezkore!

Hezkore:
@binff04 you can set it to 100 to get an even better volume! :)
At that point the game music _almost_ matches my Spotify music volume heh.
Problem is that if two (or more) similar sound effects play at the same time; they'll increase the volume of each other and _could_ clip/distort the audio.
The 10% head room left when using 250 ensures that this is less likely to happen.

PR #791, opened 2019-10-22
fix building for VS code ?

roxya:


Hezkore:
Since last commit it has compiled fine in VSCode on Windows at least.
If anything this should fix it for Mac users.

roxya:
I don't know if it's VS code or mac or whatever which stores the package in a folder without the version in the name. But apparently something does 😄

binff04:
No, it doesn't fix anything. The problem with the HintPath is that only last one will be used.

I returned back to this problem now and already have working solution that will fix everything without any changes (I only remove some obsolete stuff from README and will replace it with the new instructions).

So, please close this PR.

Thank you

binff04:
> Since last commit it has compiled fine in VSCode on Windows at least.

This is most likely because your compiler reuses `Newtonsoft.Json.dll` from the build directory.

Try to remove it from `bin/Debug` or `bin/Release` and the building process will most likely fail (but it depends on how it is configured again, so if you still can build it, then try to remove `src/obj/` and try to build again).

PR #792, opened 2019-10-22
Auto Open Secret Doors

Alias-Alan:
Added serials for all Secret Doors.
This creates an option to check on/off whether or not to include Secret Doors to AutoOpenDoors.

(Skipped boat door serials)

### DIFF
--- a/src/Game/GameObjects/PlayerMobile.cs
+++ b/src/Game/GameObjects/PlayerMobile.cs
@@ -1297,6 +1297,13 @@ namespace ClassicUO.Game.GameObjects
                                         IsDoor(s.Graphic) && s.X == x && s.Y == y && s.Z - 15 <= z &&
                                         s.Position.Z + 15 >= z))
                     GameActions.OpenDoor();
+                if (Engine.Profile.Current.AutoOpenSecretDoors)
+                {
+                    if (World.Items.Any(s =>
+                                        IsSecretDoor(s.Graphic) && s.X == x && s.Y == y && s.Z - 15 <= z &&
+                                        s.Position.Z + 15 >= z))
+                        GameActions.OpenDoor();
+                }
             }
         }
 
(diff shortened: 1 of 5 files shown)


Alias-Alan:
ezgif com-gif-maker

Example of use on a secret door placed on a custom house.

src/IO/Resources/TileDataLoader.cs
```diff
@@ -357,6 +357,7 @@ public StaticTiles(ulong flags, byte weight, byte layer, int count, ushort animI
             IsTransparent = (Flags & TileFlag.Transparent) != 0;
             IsContainer = (Flags & TileFlag.Container) != 0;
             IsDoor = (Flags & TileFlag.Door) != 0;
+            IsSecretDoor = (Flags & TileFlag.Door) != 0;
```
andreakarasho, line 360: Why do you replicate a flag that already exists?

src/Game/Pathfinder.cs
```diff
@@ -153,6 +153,10 @@ private static bool CreateItemList(ref List<PathObject> list, int x, int y, int
                                 {
                                     dropFlags = true;
                                 }
+                                else if (Engine.Profile.Current.AutoOpenSecretDoors && (item2.ItemData.IsDoor || item2.ItemData.IsSecretDoor))
```
andreakarasho, line 156: You can merge this ELSE if with the parent IF

src/Game/GameObjects/PlayerMobile.cs
```diff
@@ -1297,6 +1297,13 @@ private void TryOpenDoors()
                                         IsDoor(s.Graphic) && s.X == x && s.Y == y && s.Z - 15 <= z &&
                                         s.Position.Z + 15 >= z))
                     GameActions.OpenDoor();
+                if (Engine.Profile.Current.AutoOpenSecretDoors)
+                {
+                    if (World.Items.Any(s =>
```
andreakarasho, line 1302: This is really unperformant stuff.
You parse 2x the World.Items dictionary. You MUST merge the same behaves in an unique call

src/Game/GameObjects/PlayerMobile.cs
```diff
@@ -1860,6 +1867,34 @@ private bool IsDoor(ushort type)
                    || type >= 0x9B3C && type <= 0x9B4B;
         }
 
+        private bool IsSecretDoor(ushort type)
```
andreakarasho, line 1870: this code can me merge into the IsDoor function.
Basically if the Profile.SecretDoor option is false
```csharp
bool IsDoor(Graphic g)
{
    // the issecretdoor behaviour
    ...
   // end of issecretdoor behaviour

if (engine.profile.current.issecretdoor)
{
       // the standard isdoor behaviour
    ...
   // end of standard isdoor behaviour
 
}
 return false;
}
```

roxya:
Karasho, do you know why we use graphic ids to check for doors instead of the door flag in ItemData (which is used in Pathfinder for movement blocking / smoothdoors)? Are there cases where the door flag won't work well?

andreakarasho:
This is a good question. 
I think this is code taken from Razor. Maybe we can skip these graphic checks.

Good catch

Alias-Alan:
I asked 3 shard owners their views on this function being added, one supported it while two did not. When I asked in discord regarding this no one seemed to know whether or not this was to be considered cheating or gamebreaking or not. Therefore it is created in its own isSecretDoor to separate these serials for specific graphics linked to secret doors from those of regular door serials linked to normal doors. They behave in the same way as doors so therefore they are flagged to do so in this commit. I personally don't use this nor have it added in my local as UO Outlands does not have any secret doors according to Owyn. This was created by request without knowing whether or not it would be considered gamebreaking. As for your other comments, most fall within the realm of the answers aforementioned but if I missed something feel free to reiterate.

roxya:
If shard owners don't want it they can disable the auto opening on their side, since it's using the open door packet not a double click.

Alias-Alan:
Exactly, so you understand why I approached this the way I did. The ones who want the option but not have it apply to secret doors would not want it added to the list of serials in isDoor, therefore separating them instead of merging them was an obvious deterrent.

PR #795, opened 2019-10-23
Add better check for non-lootable items

binff04:
Skip non-lootable items (the check below is for UO:Outlands only).
These items appear only on humanoid NPC corpses so they don't look naked.
They are always a piece of equipment and server always forces their position as 0,0,0 which is almost the same as Position.INVALID (you can't place an item in 0,0,0).
So the check works the next way: show an item if the shard type is not Outlands, show the item if it isn't equipment, show the item if position in container is not 0,0,0.

### DIFF
--- a/src/Game/GameObjects/Item.cs
+++ b/src/Game/GameObjects/Item.cs
@@ -218,6 +218,16 @@ namespace ClassicUO.Game.GameObjects
             }
         }
 
+        public bool IsLootable =>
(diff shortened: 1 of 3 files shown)


binff04:
This is a second improved iteration. If it's not merged, maybe someone will find it useful for his/her own build.

Hezkore:
And just to make things clear...
This is how it looks without this fix:

You never actually get the items either, they just disappear.

PR #805, opened 2019-10-24
Fix for arrowkeys and macro player movement

Hezkore:
This fixes some input issues when moving with the keyboard.
Not only will it correctly reset `_isMacroMoveDown ` to `false` (stuck on `true` without this fix), it will also fix the issue where pressing any button will halt the players movement when using the keyboard to move.


Hezkore:
Here's an example of what this fixes.
In both of these examples I'm trying to constantly move right.


I have the macro "MovePlayer - Right" bound to the key 'D'.
And when I press 'TAB' the player will stop moving, even if I'm holding the 'D' key:
no_fix


But with this fix; the player won't stop moving until the 'D' key is actually released.
No matter how many other keys I press:
fix

PR #806, opened 2019-10-24
New macro OpenHelp

Hezkore:
Simple macro for opening the help menu.
Especially useful on servers like Outlands that heavily relies on the help menu.

ClassicUO_2019-10-24_19-43-04

### DIFF
--- a/src/Game/Managers/MacroManager.cs
+++ b/src/Game/Managers/MacroManager.cs
@@ -464,6 +464,11 @@ namespace ClassicUO.Game.Managers
 
                                     break;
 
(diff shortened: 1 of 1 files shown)


Alias-Alan:
Good idea!

roxya:
Correct me if I'm wrong but if you add to the middle of the enum, won't you break existing subtypes stored in profiles?

Hezkore:
That might be true, yeah.
Didn't think about existing macros.
I'll close this and do some testing later when time is on my side.

PR #809, opened 2019-10-25
Fix for item drag offset

Hezkore:
This will fix the offset when dragging items.
Items not in a container (paperdoll, world etc.) will still be dragged from the center of the texture.

Without this fix:
no_fix

With this fix:
fix


Hezkore:
This should be good now.

PR #811, opened 2019-10-26
CastSpell to get correct spell index numbers

creamcast:
Weird values for non-magery spells were being passed to GameActions.CastSpell(index) due to offsets being calculated incorrectly. I was able to test each book except spellweaving and mysticism since you need a quest to get the books.


andreakarasho:
I see now your PR, i've already fixed in dev.

PS: don't make PR to the master, but dev only thanks

PR #812, opened 2019-10-26
Adding WASD for movement option

creamcast:
I added an option for WASD movement as experimental feature... add if you like. Hoping I did this correctly and didn't ruin anything.


Hezkore:
You can already do this, in fact I've been playing for quite some time now using WASD.
Just make a macro for 'W' and set it to 'MovePlayer - Top', repeat for 'A' 'S' and 'D'.

roxya:
We've already got wasd if we bind player movement macros, it's much more flexible and already exists

creamcast:
Oh, it must of got fixed in dev then, because it wasn't working with the release.

PR #813, opened 2019-10-26
Small Fix for Enter After Option Allowing Macro Input

creamcast:
When setting Enter After option on in Speech options and assigning movement (e.g. WASD) to macros, macro was still being triggered while chatting when there is nothing in the text box.


andreakarasho:
added manually

PR #816, opened 2019-10-28
New option Skip empty corpse

Hezkore:
Added the option "Skip empty corpses" as a sub-option for "Auto Open Corpses".
This will make it so that "Auto Open Corpses" never opens any empty corpses.
Now also supports Health Bars and Overhead Names!

ClassicUO_2019-10-28_18-51-20

### DIFF
--- a/src/Configuration/Profile.cs
+++ b/src/Configuration/Profile.cs
@@ -172,6 +172,7 @@ namespace ClassicUO.Configuration
         [JsonProperty] public bool SmoothDoors { get; set; }
         [JsonProperty] public bool AutoOpenCorpses { get; set; }
         [JsonProperty] public int AutoOpenCorpseRange { get; set; } = 2;
+        [JsonProperty] public bool AutoCloseCorpses { get; set; }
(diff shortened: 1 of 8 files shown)


woolardjason:
Would there be anyway to open corpses by notoriety ? i.e. not open corpses that are blue and that would mark me criminal for looting.

Hezkore:
> 
> 
> Would there be anyway to open corpses by notoriety ? i.e. not open corpses that are blue and that would mark me criminal for looting.

I have no way of testing this, but possibly.
However; this is unrelated to the pull request.

Stallind:
It has been discussed and if I remember correctly the server doesn't send any information about the notoriety of corpses. What you could do is combine it with "all names" and take the notoriety from there - like in sallos. But meh

Alias-Alan:
Another option would be to simply close if criminal action flag is thrown (would need to make a safeaction that overrides whether that option is on or off though)

andreakarasho:
Can you solve conflicts?

PR #825, opened 2019-11-05
Fixed item drag offset

Hezkore:
This will fix the offset when dragging items.
Items not in a container (paperdoll, world etc.) will still be dragged from the center of the texture.

Without this fix:
no_fix

With this fix:
fix

### DIFF
--- a/src/Game/GameCursor.cs
+++ b/src/Game/GameCursor.cs
@@ -251,8 +251,7 @@ namespace ClassicUO.Game
         {
             _itemHold = hold;
             _draggedItemTexture = FileManager.Art.GetTexture(_itemHold.DisplayedGraphic);
-            _offset.X = _draggedItemTexture.Width >> 1;
-            _offset.Y = _draggedItemTexture.Height >> 1;
+            _offset = hold.Offset;
 
             _rect.Width = _draggedItemTexture.Width;
             _rect.Height = _draggedItemTexture.Height;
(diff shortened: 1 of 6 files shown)


BlackTeaOutlands:
FYI the default behaviour is to offset items when they are grabbed. the "fix" should be to reduce the offset a bit, because it was a bit too large. I don't think I personally like grabbing it at the exact pixel.

Hezkore:
> 
> 
> FYI the default behaviour is to offset items when they are grabbed. the "fix" should be to reduce the offset a bit, because it was a bit too large. I don't think I personally like grabbing it at the exact pixel.

I guess "fix" isn't the right word here, as the client always just acted like the original client anyways.
But what happens in the original GIF picture happens because the barrel is not centered in its image bounds.
By default every item is grabbed from the center, and there is no way of know what the center of the actual graphics is, so there's no way of getting the correct behaviour when grabbing things like the barrel and many other items with incorrect graphics in their image bounds.
So I just did what every other game, scrollbar, window, or other general interface item from the past 20 years does, and make you drag them from where you actually started dragging it.

Here's another example.
The bloody bandages aren't in the center of their image bounds either:
Photoshop_2019-11-23_17-36-32
So you'll always be dragging them slightly off center, as there's no way of knowing what the "actual" center is.

This is a moot point though, seeing as how this is an option now, rather than a standard feature.

PR #826, opened 2019-11-09
Fixed currently playing sounds volume change, disable and background

paezao:
There's something annoying about changing Sound options in the current ClassicUO release.

When you change volume, for example, it will only work for new sound effects that aren't currently playing. I fixed it so it changes the volume of the sounds that are playing currently. That's also the case for disabling sound and also background sound play.

### DIFF
--- a/src/Game/Managers/AudioManager.cs
+++ b/src/Game/Managers/AudioManager.cs
@@ -21,6 +21,7 @@
 
 #endregion
 
+using System.Collections.Generic;
 using ClassicUO.Game.Scenes;
 using ClassicUO.IO;
 using ClassicUO.IO.Audio;
@@ -32,6 +33,7 @@ namespace ClassicUO.Game.Managers
     internal class AudioManager
     {
         private UOMusic _currentMusic;
+        private List<UOSound> _currentSounds;
 
         private bool _canReproduceAudio = true;
 
@@ -39,14 +41,15 @@ namespace ClassicUO.Game.Managers
         {
             try
             {
-                new DynamicSoundEffectInstance(0, AudioChannels.Stereo)
-                    .Dispose();
+                new DynamicSoundEffectInstance(0, AudioChannels.Stereo).Dispose();
             }
             catch (NoAudioHardwareException ex)
             {
                 Log.Message(LogTypes.Warning, ex.ToString());
                 _canReproduceAudio = false;
             }
+
+            _currentSounds = new List<UOSound>();
         }
 
         public void PlaySound(int index, AudioEffects effect = AudioEffects.None, bool spamCheck = false)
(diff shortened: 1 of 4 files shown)


src/Game/Managers/AudioManager.cs
```diff
@@ -148,6 +163,25 @@ public void UpdateCurrentMusicVolume()
...
+
+            if (Engine.Profile == null || Engine.Profile.Current == null || !Engine.Profile.Current.EnableSound)
+                return;
+
+            float volume = Engine.Profile.Current.SoundVolume / Constants.SOUND_DELTA;
+
+            if (volume < -1 || volume > 1f)
+                return;
+
+            foreach (UOSound sound in _currentSounds.ToArray())
```
andreakarasho: Why .ToArray() ?

src/Game/Managers/AudioManager.cs
```diff
@@ -158,6 +192,16 @@ public void StopMusic()
             }
         }
 
+        public void StopSounds()
+        {
+            foreach (UOSound sound in _currentSounds.ToArray())
```
andreakarasho: I think you can convert this code to a simple for.

```csharp
for (int i = 0; i < _currentSounds.Count; i++)
{
   var s = _currentSounds[i];
// do stuff
_currentSounds.RemoveAt(i--);
}
```

src/Game/Managers/AudioManager.cs
```diff
@@ -173,6 +217,17 @@ public void Update()
             }
 
             _currentMusic?.Update();
+
+            foreach (UOSound sound in _currentSounds.ToArray())
```
andreakarasho: same above

andreakarasho (commented):
The most important thing while code is to avoid using LINQ when possible.
The ToArray() method basically create a new Array() filled with the list values in runtime, it's not the best thing to do :)

src/Game/Managers/AudioManager.cs
paezao: I added that so I could remove from within the loop. But you are right, I will follow your for suggestion.

src/Game/Managers/AudioManager.cs
paezao: Done

src/Game/Managers/AudioManager.cs
paezao: Done

src/Game/Managers/AudioManager.cs
```diff
@@ -158,6 +192,17 @@ public void StopMusic()
...
         }
 
+        public void StopSounds()
+        {
+            for (int i = 0; i < _currentSounds.Count; i++)
+            {
+                var sound = _currentSounds[i];
+                sound.Stop();
+                sound.Dispose();
+                _currentSounds.RemoveAt(i);
```
andreakarasho: _currentSounds.RemoveAt(i--);

src/Game/Managers/AudioManager.cs
```diff
@@ -173,6 +218,24 @@ public void Update()
...
 
             _currentMusic?.Update();
+
+            for (int i = 0; i < _currentSounds.Count; i++)
+            {
+                var sound = _currentSounds[i];
+
+                if (!sound.IsPlaying())
+                {
+                    _currentSounds.RemoveAt(i);
```
andreakarasho: _currentSounds.RemoveAt(i--);

andreakarasho (commented):
Don't forget to decrease the index if you remove an element in an array/list while parsing it

Issue #381, opened 2019-03-31
Chat gradient: ON / OFF


andreakarasho:
add an option to hide/unhide the chat blend

zolter:
What do you mean? AlphaBlendControl in systemchatcontrol?

paezao:
@andreakarasho Do you mean the semi transparent chat bar?

paezao:
@andreakarasho I guess this can be closed

PR #827, opened 2019-11-09
Added Hide Chat Gradient video option

paezao:
I added an option that hides the chat text entry gradient. Default is false so it works by default as it is currently.

#381


andreakarasho:
I tried to solve conflicts by myself but github still blocking the merge button

PR #836, opened 2019-11-13
Added borderless window mode

paezao:
Added Borderless Window functionality.


andreakarasho:
What is the behaviour if you switch scene [login -> game, game -> login]

paezao:
It stays borderless. Do you want it to be always bordered in the login scene?

andreakarasho:
the correct behaviour is: stays in bordless when you are in game only

paezao:
Okay @andreakarasho, it's done. Login scene is now always bordered.

andreakarasho:
Reviewing....

PR #841, opened 2019-11-15
New option SkipEmptyCorpse

Hezkore:
Added the option "Skip empty corpses" as a sub-option for "Auto Open Corpses".
This will make it so that "Auto Open Corpses" never opens any empty corpses.

ClassicUO_2019-10-28_18-51-20

I've updated the way this works to make it more secure.
Basically; if a corpse is opened by the auto open corpse feature, the gump will be invisible until the server sends its content.
If no content is ever sent; the gump will instead close when the player walks away, or the corpse no longer exists.
That means that there's no risk of ever having the corpse container close before the server has had time to send the containing items.


Hezkore:
So, apparently invisible gumps still grab mouse input.
I'll have to figure out how to solve that.

Hezkore:
> 
> 
> So, apparently invisible gumps still grab mouse input.
> I'll have to figure out how to solve that.

Commit 26350de519839b32e160f3cceed92ee286b68fa7 fixed this issue.

PR #844, opened 2019-11-17
New option AutoLootGold

Hezkore:
This PR might be a bit controversial, but I figured I'd share it anyway.
It adds a new experimental option called "Auto loot gold from corpses".
ClassicUO_2019-11-17_23-35-15
It will automatically loot any gold found in any corpse container you've opened and place the gold into your grab bag. It does _not_ require "Auto open corpses" to be enabled to function.

### DIFF
--- a/src/Configuration/Profile.cs
+++ b/src/Configuration/Profile.cs
@@ -174,6 +174,7 @@ namespace ClassicUO.Configuration
         [JsonProperty] public int AutoOpenCorpseRange { get; set; } = 2;
         [JsonProperty] public int CorpseOpenOptions { get; set; } = 3;
         [JsonProperty] public bool SkipEmptyCorpse { get; set; }
+        [JsonProperty] public bool AutoLootGold { get; set; }
         [JsonProperty] public bool DisableDefaultHotkeys { get; set; }
         [JsonProperty] public bool DisableArrowBtn { get; set; }
         [JsonProperty] public bool DisableTabBtn { get; set; }
--- a/src/Game/UI/Gumps/ContainerGump.cs
+++ b/src/Game/UI/Gumps/ContainerGump.cs
@@ -22,6 +22,7 @@
 #endregion
 
 using System.IO;
+using System.Collections.Generic;
 using System.Linq;
 
 using ClassicUO.Configuration;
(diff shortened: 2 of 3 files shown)


owyn-outlands:
This begins to stray too far from what we hope ClassicUO will become.   The fact you're using it on Outlands is even more troublesome.

Hezkore:
> 
> 
> This begins to stray too far from what we hope ClassicUO will become. The fact you're using it on Outlands is even more troublesome.

Yeah I figured it wasn't going to be liked.
Should I just close this down and forget about it?

expiation:
some of us play on many servers, including servers where this would be allowed.  Maybe a good compromise here could be respecting razor negotiate before allowing it to be enabled?

Hezkore:
Auto looting gold is already possible using assistants like UOSteam.
And Razor can scavenge things from the ground.
It just requires you to setup the assistant, then either find or write a script and run it.
But not everyone will have the time or knowledge on how to set that up.
Which separates the player base and gives some players an unfair advantage.
So, at least to me, using something like Razor or others assistants is straight up cheating.
But most servers still allow them, and a lot of server launchers actually comes with an assistant tool from the start.

But Auto loot gold is something everyone can enable easily.
It doesn't require any knowledge or extra work, you just enable it and everyone gets the same benefit.
You're not getting more gold from it either, it just saves you a click or two.
And it makes it fair for everyone.

> 
> 
> some of us play on many servers, including servers where this would be allowed. Maybe a good compromise here could be respecting razor negotiate before allowing it to be enabled?

I don't know if we can actually listen to Razor and what it negotiates with the server.
(we don't with targetting for example)

Having said all of this, I'd still like to say that in the end; I don't care heh.
If people hate this I'll just close the PR and save you all some time.
You can still find the code on my Github if you want to use it for your own server.

Alias-Alan:
While I understand Owyns disinterest in this specific feature especially on his server as it's breaking their rules(auto-looting), this does beg the question, Is the end goal for ClassicUO to be an All-In-One client that doesn't need to be accompanied by a third party assistant program? If that is the case than I do believe this is a good inclusion as it could essentially replace the auto-looting features that assistants offer if fully developed and designed in a manner where the end-user has the option to add/remove the things they want to loot. With that being said, because some servers allow autolooting and some do not, it is absolutely necessary for features like this to respect the shards accepted/rejected features. If the goal isn't to have an all inclusive client, I personally don't see the need for this on the client-side.

PR #850, opened 2019-11-25
Add party invite macro

Stallind:
Binding a macro to say /add doesn't work so I added a macro to invite by target

Update:
Added more commands 

https://gyazo.com/f1dc346dd2d95fba5f1212ed37805d72

### DIFF
--- a/src/Game/GameObjects/Item.cs
+++ b/src/Game/GameObjects/Item.cs
@@ -736,6 +736,35 @@ namespace ClassicUO.Game.GameObjects
                         graphic = 0x0582;
                         break;
                     }
+                        
+                    case 0x3ED1: // CoconutCrab
+                    {
+                        graphic = 0x05A0;
+                        break;
+                    }
+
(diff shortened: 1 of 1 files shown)


Stallind:
Figured why not just add them all -> Add, accept, quit =)

Alias-Alan:
This hits on the problem I had with the info command. I had to do the same thing and had to make an actual macro to execute the command because for whatever reason say isn't recognizing symbols within the string.

src/Game/Managers/MacroManager.cs
```diff
@@ -1783,6 +1831,10 @@ internal enum MacroSubType
         Right,
         Down,
         Left,
+        Add,
```
andreakarasho: These should be added at the end of the enum or you will break macroes that already exsit

andreakarasho:
I want to be sure thst i ve added all common potions in the submacro enum before merging this one

PR #856, opened 2019-12-01
Prevent sound spamming option

Ultima-Tony:


andreakarasho:
I fixed this issue like original client does.
Basically it takes the duration of the stream and do some math to avoid repetition and sound overlapping

PR #876, opened 2019-12-28
Animations: Fix frame interval calculations

jaedan:
The OSI client had a fixed frame rate of 12.5 FPS, or 80ms per frame.
Most of the math was done in units of "frames", and what "speed" means
in many of the graphic effect packets is how many of these "frame" units
to go before changing over to the next frame in the animation. These
"frameInterval" values are 0 based, so a value of 0 means every frame
(80ms), a value of 1 means every other frame (160ms), etc.

ClassicUO supports a variable framerate, so it needs to convert all of
these values to clock times. The formula for that is simple:

(frameInterval + 1) * 80


Stallind:
@jaedan please change it to merge into dev and not master =)

PR #916, opened 2020-01-23
BulletinBoard fix + swing packet

TannerBist:


src/Game/UI/Gumps/BulletinBoardGump.cs
```diff
@@ -32,9 +34,11 @@ namespace ClassicUO.Game.UI.Gumps
 {
     internal class BulletinBoardGump : Gump
     {
-        //private HtmlControl _htmlControl;
         private readonly ScrollArea _area;
         private readonly Item _item;
+        private readonly ArrayList _elements = new ArrayList();
```
andreakarasho: List<T> please :D

src/Network/PacketHandlers.cs
```diff
@@ -1318,7 +1319,57 @@ private static void EquipItem(Packet p)
...
+            Mobile def = World.Mobiles.Get(defenders);
+            Direction pdir = GetDirectionAB(World.Player.X, World.Player.Y, def.X, def.Y);
+            if (World.Player.X == _pdX && World.Player.Y == _pdY && World.Player.Direction != pdir)
+            {
+                NetClient.Socket.Send(new PWalkRequest(pdir, World.Player.Walker.WalkSequence, false, World.Player.Walker.FastWalkStack.GetValue()));
+            }
+            _pdX = World.Player.X;
+            _pdY = World.Player.Y;
+        }
+        public static Direction GetDirectionAB(int AAx, int AAy, int BBx, int BBy)
```
andreakarasho: Can you add this method to Data/Direction.cs class ?

src/Network/PacketHandlers.cs
```diff
@@ -1870,8 +1921,12 @@ private static void ClientViewRange(Packet p)
         {
             World.ClientViewRange = p.ReadByte();
         }
-
-        private static void BulletinBoardData(Packet p)
+        public static string Truncate(string value, int maxLength)
```
andreakarasho: If this method is used 1 time, take the body and push it into the method caller.
Instead add it to Utility/StringHelper.cs class

src/Network/Packets.cs
```diff
@@ -879,23 +879,25 @@ public PBulletinBoardRequestMessageSummary(uint serial, uint msgserial) : base(0
 
     internal sealed class PBulletinBoardPostMessage : PacketWriter
     {
-        public PBulletinBoardPostMessage(uint serial, uint msgserial, string subject, string message) : base(0x71)
+        public PBulletinBoardPostMessage(uint serial, uint msgserial, string subject, Game.UI.Controls.MultiLineBox _textBox) :
```
andreakarasho: Instead of use MultilineBox to take the text, just pass the Text as string.
The reason is code consistency. Packet needs a string and you use the .Text property only from MultiLineBox.

Basically rollback the sign.

PR #934, opened 2020-02-07
Updated parsing for world map gump.

Kyain:
Corrected parsing for party members in 0xF0
modified WMapEntity AddOrUpdate to ignore null or invalid data and retain previous good data
corrected removal of unupdated entities

### DIFF
--- a/src/Game/Managers/WorldMapEntityManager.cs
+++ b/src/Game/Managers/WorldMapEntityManager.cs
@@ -28,30 +28,15 @@ namespace ClassicUO.Game.Managers
         public WMapEntity(uint serial)
         {
             Serial = serial;
-
-            //var mob = World.Mobiles.Get(serial);
-
-            //if (mob != null)
-            //    GetName();
         } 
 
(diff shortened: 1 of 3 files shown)


andreakarasho:
My answer from discord: 
```
i looked at your PR. You found the inverted sign bug. GG! 
BUT...
According to the ServUO Packet 0xF0, ClassicUO parses that packet correctly and i don't see any reason to add the "IsParty" field in WmEntity class.
Infact "IsGuild" field is FALSE when WmEntity is a Party entity. So no reason to waste memory :D.
SO...
I took 2 things from your PR:
- sign fix
- update/create wmentities when updating World.Mobiles dictionary [infact servuo does not send any serial or info if the ally/party mobile is in range]
I'm going to not accept the PR due to the fact i added the above stuff to fix the worldmap system. This does not mean "your pr is wrong", but your PR is partially OK and thanks to it i found what missing :smile: 
Infact i'll add you as author of the fix to the next changelog```

PR #936, opened 2020-02-07
Network Stats is now a gump

dissident76:
Double click to change amount of info displayed

### DIFF
--- a/src/Game/UI/Gumps/NetworkStatsGump.cs
+++ b/src/Game/UI/Gumps/NetworkStatsGump.cs
@@ -1,0 +1,139 @@
+﻿#region license
+// Copyright (C) 2020 ClassicUO Development Community on Github
+// 
+// This project is an alternative client for the game Ultima Online.
+// The goal of this is to develop a lightweight client considering
+// new technologies.
+// 
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+// 
+//  This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


src/Game/UI/Gumps/NetworkStatsGump.cs
```diff
@@ -0,0 +1,129 @@
...
+                _label.Hue = NetClient.Socket.Statistics.RenderedText.Hue;
+                _label.Text = NetClient.Socket.Statistics.RenderedText.Text;
+            }
+            else
+            {
+                _label.Hue = 0x20;
+                _label.Text = "Disconnected";
+            }
+                       
+            if (!FullDisplayMode) _label.Text = _label.Text.Split('\n')[0];
```
andreakarasho: This is cause a lot of allocation in runtime. 
What is the purpose?

andreakarasho:
Now i've a clear view of the PR.
If networkstatistics class does not draw text anymore, you can remove the RenderedText field and making the string into the gump directly

src/Game/UI/Gumps/NetworkStatsGump.cs
dissident76: This is so the gump can be minimised with double click and show only ping value, without bytes in/out. So RenderedText gets trimmed right before the \n character that separates ping value and the rest. But can get rid of the minimised mode if it creates a lot of overhead

dissident76:
Thought of moving the string creation to the gump class, but I wanted to change the existing NetStatistic class the least possible. If I move the string to the gump, I will have to create an Update method to calculate bytes in/out. But yeah, NetStatistics needs an overhaul -will give it another go tomorrow

PR #940, opened 2020-02-12
container gumps now better match override option

dissident76:
container gumps locations will be loaded from gumps.xml on login, but gumps location will no longer be remembered after gumps are closed when "remember every gump" is not selected

### DIFF
--- a/src/Game/Managers/UIManager.cs
+++ b/src/Game/Managers/UIManager.cs
@@ -337,6 +337,11 @@ namespace ClassicUO.Game.Managers
             _gumpPositionCache[serverSerial] = point;
         }
 
+        public static bool RemovePosition(uint serverSerial)
+        {
+            return _gumpPositionCache.Remove(serverSerial);
+        }
+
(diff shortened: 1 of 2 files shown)


andreakarasho:
This solution is a bit... ugly :D

roxya:
What if you just RemovePosition in the Dispose() method, immediately before

roxya:
What if you remove the cache in the Dispose() method instead of every draw? (actually should be update if you were going this route).

`UIManager.RemovePosition(LocalSerial);`
`if (World.Player != null && (ProfileManager.Current?.OverrideContainerLocationSetting == 3)) UIManager.SavePosition(item, Location);`

Maybe even in BuildGump after setting location by cache, do we need it after we used it?

dissident76:
Actually, I just realised I made a mistake placing the Remove call in the wrong BuildGump place (wasn't being called), and missed the point that Draw is getting called so often. Fixed in new PR  by placing it at the end of BuildGump()

PR #941, opened 2020-02-13
container gumps better match override option expected behaviour

dissident76:
not as ugly now :D (not being called within Draw())


dissident76:
damn, wrong branch again sorry :P

PR #943, opened 2020-02-13
[MOD] Correct rendering of items with partial hue on CountersBar.

jinh86:
Items on the CountersBar were rendered without the partial hue check, this should fix the issue.


andreakarasho:
the IsPartial can be taken from TileDataLoader.Statics

jinh86:
True, didn't think about it. Gonna edit.

PR #951, opened 2020-02-23
WorldMapGump zoom level saved in settings

dissident76:


src/Game/UI/Gumps/WorldMapGump.cs
```diff
@@ -44,7 +45,7 @@ internal class WorldMapGump : ResizableGump
 
         private bool _isTopMost;
         private readonly float[] _zooms = new float[10] { 0.125f, 0.25f, 0.5f, 0.75f, 1f, 1.5f, 2f, 4f, 6f, 8f };
-        private int _zoomIndex = 4;
+        private int _zoomIndex = ProfileManager.Current.WorldMapZoom;
```
andreakarasho, line 48: add a check to validate if the WorldMapZoom value is between 0 and _zooms.Length

andreakarasho:
Store data into gumps.xml. Just follow how i save topmost, freeview and other wmap options

dissident76:
ok, I assumed nothing from WorldMap was being saved :P

PR #952, opened 2020-02-24
WorldMap zoom saved

dissident76:
this one should make more sense


src/Game/UI/Gumps/WorldMapGump.cs
```diff
@@ -83,7 +83,8 @@ public override void Restore(XmlElement xml)
             TopMost = bool.Parse(xml.GetAttribute("topmost"));
             FreeView = bool.Parse(xml.GetAttribute("freeview"));
             _showPartyMembers = bool.Parse(xml.GetAttribute("showpartymembers"));
-
+            _zoomIndex = int.Parse(xml.GetAttribute("zoomindex"));
```
andreakarasho: for older version the xml.GetAttribute("zoomindex") will return an empty string.
This will cause an exeption when passing this value to int.Parse.
Use int.TryParse

dissident76: Should be fine now... thanks for your patience, lol

PR #953, opened 2020-02-25
Protocol Extensions behind settings

DarkLotus:
Moves the use of these server specific custom packets behind a Settings.json, to allow Sphere/POL players to play without flooding their server consoles with unhanded packets.

### DIFF
--- a/src/Configuration/Settings.cs
+++ b/src/Configuration/Settings.cs
@@ -103,7 +103,8 @@ namespace ClassicUO.Configuration
         [JsonProperty(PropertyName = "plugins")]
         public string[] Plugins { get; set; } = {@"./Assistant/Razor.dll"};
 
-
+        [JsonProperty(PropertyName = "ProtocolExtensions")]
+        public bool ProtocolExtensions { get; set; } = true;
 
 
         public const string SETTINGS_FILENAME = "settings.json";
--- a/src/Configuration/Settings.cs
+++ b/src/Configuration/Settings.cs
@@ -103,7 +103,7 @@ namespace ClassicUO.Configuration
         [JsonProperty(PropertyName = "plugins")]
         public string[] Plugins { get; set; } = {@"./Assistant/Razor.dll"};
 
-        [JsonProperty(PropertyName = "ProtocolExtensions")]
+        [JsonProperty(PropertyName = "protocol_extensions")]
         public bool ProtocolExtensions { get; set; } = true;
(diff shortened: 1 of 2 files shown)


andreakarasho:
I think to send a first PQueryGuildPosition/PQueryPartyPosition as ACK.
If after a time T server does not answer to the packet, CUO will stop to send these packets.
Maybe razor acts in the same way

andreakarasho:
Pushed some days ago the fix about flood packet queries.
As i said i prefer to act into the client to automatically manage the situation.
Thank you anyway

DarkLotus:
No problems :+1:

brndd:
@andreakarasho 

821f8c1 this commit doesn't really fix the issue. The packets are still sent, but only every 2000ms instead of 250ms, so there is still some spam (although less) in the server console if the server is logging unknown packets.

I tried a fix (https://github.com/brndd/ClassicUO/commit/033e4c58a3a587955fc1f7437a439dd40fcb4599) that stops sending these packets if there is no response for 10 seconds, but I think this will break the feature on servers that *do* support the packet, because it appears there is no response to the packet if the player is not in a party/guild -- at least there wasn't on OutlandsUO, which I think supports this packet, but further testing on other servers would be good.

Maybe putting it behind a config setting like this PR did is the only feasible option to satisfy server admins that want to log unknown packets.

brndd:
Update: servers that support the packet *will* respond to it even if you are not in a party/guild, therefore automatic detection of server support is possible. The packet handler just wasn't calling any methods when the packet didn't contain info on any entities.

Here's how I would do it, if it was to be done automatically:

1. Send ONE packet to server when when world map is opened with party/guild tracking enabled.
2. If server doesn't respond to the packet after 10 seconds (with a `0xF0` packet of type `0x01` or `0x02`), disable party/guild tracking, and display a SysMessage eg. `WorldMap: This server does not appear to support party/guild tracking, and it has been automatically disabled.`. The user can re-enable it if they want to to repeat this process.
3. If server responds within 10 seconds, continue sending these packets normally every 250ms (or less frequently, 250ms delay is quite short -- maybe 500ms or 1000ms would be sufficient).

This might, however, have problems with packet loss.

I think being able to disable protocol extensions entirely would still be good though, because some servers might use them to detect if the players are using 3rd party tools, namely Razor. While CUO is a third-party tool, it is not Razor.

PR #959, opened 2020-02-28
Fix Limit3Spell MessageTypes not displaying overhead text.

### LINKED ISSUE #957: Speech text is not displayed on items in inventory if the item is "speaking"
**Describe the bug**

If an item in your inventory (or paperdoll) "speaks", by having a `0xAE` [unicode speech message packet](https://docs.polserver.com/packets/index.php?Packet=0xAE) sent for that item, the speech text is not displayed above the item. It is still displayed in your journal.

This is a bug I encountered on a small Finnish whitelist-based server with substantial amounts of custom content that I play on, so reproducing it may be difficult. I've done my best to narrow it down using packet logging, but I am unable to provide a live server environment where it can be reproduced. If a test environment is necessary, let me know and I'll get in contact with the shard staff to see if something can be worked out.

Here is an example packet from Razor's packet logger that a server might send in order to make an item in your inventory "speak":

```
05:52:32.3711: Server -> Client 0xAE (Length: 86)
        0  1  2  3  4  5  6  7   8  9  A  B  C  D  E  F
       -- -- -- -- -- -- -- --  -- -- -- -- -- -- -- --
0000   AE 00 56 71 3B 48 F0 00  00 03 03 D9 00 00 00 00   ..Vq;H..........
0010   00 00 4E E4 65 74 00 00  00 00 00 00 00 00 00 00   ..N.et..........
0020   00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00   ................
0030   00 31 00 32 00 38 00 31  00 20 00 72 00 61 00 75   .1.2.8.1. .r.a.u
0040   00 74 00 61 00 6B 00 6F  00 6C 00 69 00 6B 00 6B   .t.a.k.o.l.i.k.k
0050   00 6F 00 61 00 00                                  .o.a..
```

**To Reproduce**

1. Have a `0xAE` unicode speech message packet sent for an item in your inventory or on your paperdoll.
2. The speech text will not appear above the item, but will appear in your journal.

**Expected behavior**

The speech text should appear both above the item and in your journal.

**Screenshots**

Expected behaviour:

Actual behaviour:

**Desktop (please complete the following information):**
 - OS: Windows 10
 - ClassicUO Version: 0.1.3.200

brndd:
Fixes #957 

Some Sphere (and in this case, POL) shards use MessageType 3 for some overhead text. This has no handling in GameScene.cs, so currently this text only appears in the journal.

MessageType.cs has the following comment attached to the type:

`// Sphere style shards use this to limit to 3 of these message types showing overhead.`

This suggests that only 3 of these messages are supposed to be visible at any one time. However, in my testing with the original client, an arbitrary number of these messages could be visible so long as they came from different sources (eg. on different items in your backpack). For any single given item only 3 of these messages could ever be visible at a time, but that seems to apply to *all* overhead text in the old client.

In other words, I did not observe any special behaviour for this MessageType on the original client. For that reason I'm submitting this bandaid fix of simply treating this MessageType the same as the regular text MessageType (`0x00`).

### DIFF
--- a/src/Game/Managers/MessageManager.cs
+++ b/src/Game/Managers/MessageManager.cs
@@ -115,6 +115,7 @@ namespace ClassicUO.Game.Managers
                 case MessageType.Yell:
                 case MessageType.Regular:
                 case MessageType.Label:
+                case MessageType.Limit3Spell:
 
                     if (parent == null)
                         break;
--- a/src/Game/Scenes/GameScene.cs
+++ b/src/Game/Scenes/GameScene.cs
@@ -225,6 +225,7 @@ namespace ClassicUO.Game.Scenes
             switch (e.Type)
             {
                 case MessageType.Regular:
+                case MessageType.Limit3Spell:
 
                     if (e.Parent == null || !SerialHelper.IsValid(e.Parent.Serial))
                         name = "System";


andreakarasho:
PR must point to dev branch

brndd:
That should do it.

PR #960, opened 2020-03-01
Fixes vendor gump uint to ushort cast leading to integer overflow

brndd:
Vendor gumps would sometimes display incorrect prices because there was an unsafe series of typecasts from uint to int to ushort. Items costing over 0xFFFF would get incorrect prices due to overflow. Presumably items costing over 0x7FFFFFFF would have ended up getting negative prices too.

This converts the price to uint across the board, because the packet the price is from (0x74) is parsed as uint in the first place.

It should be noted that what type the packet actually contains depends on the emulator. RunUO and ServUO seem to use signed int, while POL seems to use unsigned int. What OSI uses I don't know. Unsigned seems like a safe bet because vendors should never sell negative-price items.


andreakarasho:
thank you!

PR #966, opened 2020-03-02
add Party members to map update

KevinEady:
When a party member comes within view range, [the client will no longer request location data](https://github.com/andreakarasho/ClassicUO/blob/4305a4f0c355dacb25f99fabe17c1c86eba0415f/src/Game/UI/Gumps/WorldMapGump.cs#L263) for them. This PR updates the World Map Manager's cache with party members, so they won't be removed in the subsequent call to `WMapManager.RemoveUnupdatedWEntity`.

### DIFF
--- a/src/Game/World.cs
+++ b/src/Game/World.cs
@@ -197,14 +197,14 @@ namespace ClassicUO.Game
 
                     if (mob.IsDestroyed)
                         _toRemove.Add(mob);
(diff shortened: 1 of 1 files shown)


Stallind:
I saw that karasho added this himself, thanks for pointing it out. 

For future reference when making PR's make sure to set the target branch to dev and not master.
Thanks again =)

PR #974, opened 2020-03-08
Additional WorldMap Options

markdwags:
Ability to hide/show your name, your healthbar, hide/show mobiles and hide/show your current coords.

### DIFF
--- a/src/Game/UI/Gumps/WorldMapGump.cs
+++ b/src/Game/UI/Gumps/WorldMapGump.cs
@@ -20,6 +20,7 @@
 #endregion
 
 using System;
+using System.Text;
 using System.Threading.Tasks;
 using System.Xml;
 using ClassicUO.Game.Data;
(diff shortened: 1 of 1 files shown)


src/Game/UI/Gumps/WorldMapGump.cs
```diff
@@ -476,29 +503,43 @@ public override bool Draw(UltimaBatcher2D batcher, int x, int y)
 
         private void DrawAll(UltimaBatcher2D batcher, int gX, int gY, int halfWidth, int halfHeight)
         {
-            foreach (Mobile mob in World.Mobiles)
+            if (_showCoordinates)
             {
-                if (mob == World.Player)
-                    continue;
+                _sb.Clear();
+                _sb.AppendLine($"{World.Player.X}, {World.Player.Y} ({World.Player.Z})");
```
andreakarasho: using SB is a bit overhead here.
Just store player.x,y,z into 3 vars and check if they change. When they change update the string

markdwags: Updated PR based on your feedback. Thanks!

PR #976, opened 2020-03-09
HealthBarGump LT-Highlight

Alias-Alan:
ClassicUO - LastTarget Highlight (Red Border)
Syrupz-UO-LastTargetHighlight

Was a small feature request by Merrit. Sounded like a proper request so I dusted off the cobwebs. This simply highlights the custom bar with a RED border(similar to tabbing into war-mode for your personal health bar). Once a mobile is set to last target, their health bar (if opened, in range, and not your own) will now have a red border around it to make identifying last target selected easier for players.

### DIFF
--- a/src/Game/UI/Gumps/HealthBarGump.cs
+++ b/src/Game/UI/Gumps/HealthBarGump.cs
@@ -530,7 +530,7 @@ namespace ClassicUO.Game.UI.Gumps
                     _bars[0].IsVisible = true;
                 }
 
-                if (TargetManager.LastTarget != World.Player && !_outOfRange && !mobile != null)
+                if (TargetManager.LastTarget != World.Player && !_outOfRange && mobile != null)
                 {
                     if (mobile == TargetManager.LastTarget)
                     {
--- a/src/Game/UI/Gumps/HealthBarGump.cs
+++ b/src/Game/UI/Gumps/HealthBarGump.cs
@@ -530,7 +530,27 @@ namespace ClassicUO.Game.UI.Gumps
                     _bars[0].IsVisible = true;
                 }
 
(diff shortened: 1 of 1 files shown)


src/Game/UI/Gumps/HealthBarGump.cs
```diff
@@ -530,7 +530,27 @@ public override void Update(double totalMS, double frameMS)
                     _bars[0].IsVisible = true;
                 }
 
+                if (TargetManager.LastTarget != World.Player && !_outOfRange)
+                {
+                    if (mobile == TargetManager.LastTarget)
```
andreakarasho, line 535: mobile could be null.
Use LocalSerial

src/Game/UI/Gumps/HealthBarGump.cs
```diff
@@ -530,7 +530,27 @@ public override void Update(double totalMS, double frameMS)
...
+                    {
+                        _border[0].LineColor = HPB_COLOR_RED;
 
+                        if (_border.Length >= 3)
+                        {
+                            _border[1].LineColor = _border[2].LineColor = _border[3].LineColor = HPB_COLOR_RED;
+                        }
+                    }
+                    else if (mobile != TargetManager.LastTarget)
```
andreakarasho, line 544: mobile could be null.
Use LocalSerial

Alias-Alan: fixed both problems by throwing null check first before checking lasttarget

Alias-Alan (commented):
I added the check though I feel it's rather redundant no? The entity null check is done before declaring mobile so it should never even get to the section I added is it returns null. Though perhaps I'm missing a way for the mobile to be destroyed after it's declaration at that point?

PR #977, opened 2020-03-11
WorldMap enhancements

markdwags:
WorldMap reads CSV, UOAM and Ultima Mapper marker files from `Data\Client` -- *.map files and *.xml files -- and displays them on the map.

Updated context menu to show/hide specific markers
Markers only render if they are in the the viewport
Labels won't render when zoomed out at a certain level

### DIFF
--- a/src/Game/UI/Gumps/WorldMapGump.cs
+++ b/src/Game/UI/Gumps/WorldMapGump.cs
@@ -1,4 +1,5 @@
 ﻿#region license
+
 // Copyright (C) 2020 ClassicUO Development Community on Github
 // 
 // This project is an alternative client for the game Ultima Online.
@@ -17,6 +18,7 @@
 // 
 //  You should have received a copy of the GNU General Public License
 //  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
 #endregion
 
 using System;
@@ -34,7 +36,6 @@ using ClassicUO.IO.Resources;
 using ClassicUO.Network;
 using ClassicUO.Renderer;
 using ClassicUO.Utility;
-
 using Microsoft.Xna.Framework;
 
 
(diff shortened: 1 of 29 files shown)


Alias-Alan:
This is an amazing add-in Quick, fantastic!! Would there be a way to import small icons(basic ones like moongate, staircase for dungeon levels, skullandcrossbones for dungeon entrances, a res shrine for shrines) that would be pretty cool. Love that you added the ability to color markers though!

src/Game/UI/Gumps/WorldMapGump.cs
```diff
@@ -86,6 +121,13 @@ public override void Restore(XmlElement xml)
             if (int.TryParse(xml.GetAttribute("zoomindex"), out int value))
                 _zoomIndex = (value >= 0 && value < _zooms.Length) ? value : 4;
 
+            _showCoordinates = bool.Parse(xml.GetAttribute("showcoordinates"));
```
andreakarasho: bool.TryParse instead of bool.Parse 
or it will throw errors

markdwags: Ok, bool.Parse is used throughout the class -- you want all of them refactored?

andreakarasho: Yes just to avoid exceptions with retro compatibility

markdwags: Made those changes to the bool parsing.

andreakarasho:
Can you solve conflicts?

markdwags:
> Can you solve conflicts?

Done though not sure it went through since I don't have write access.

markdwags:
It's messed up, and I hate git.

PR #983, opened 2020-03-14
Allows Prof.txt to support skills with spaces in it

markdwags:
Prof.txt is used to define the templates during character creation. Skills like "Animal Taming" are not parsed correctly and the character is created without skills.

This fix adds support so if the template has "Animal_Taming" it will correctly load based on the name in SkillsLoader (in this case, "Animal Taming" but will apply to any skill with a space)


markdwags:
Addressed by https://github.com/andreakarasho/ClassicUO/commit/30dfe14a19222df5c87e2377ae6ca4a146516ff3

PR #990, opened 2020-03-19
World Map now has 6 fonts of choose from

markdwags:
Menu option to select from 6 fonts that should be much easier to read on the map


markdwags:
Additional commit to support zoom levels on markers

PR #992, opened 2020-03-20
More World Map Tweaks

markdwags:
A few tweaks based on feedback

- When using ALT to freeview, it won't snap back to your position until you release ALT.
- The zoom level will show in the coordinates label (if that it showing)
- Show/hide party/group names and heathbars option


andreakarasho:
Nvm

PR #997, opened 2020-03-25
Some book fixes

brndd:
Books (regular text kind) had many bugs. They still have many bugs, but I fixed a couple.

Here's what this fixes:

- Read-only books did not work (they have their page data sent only for the visible pages, instead of all at once like for editable books).
- Editing packets were being sent for read-only books.
- Book titles were not using UTF-8 encoding, but the original client treats them as UTF-8. Left them as ASCII for version 2.x clients which had some special treatment already, but this could be incorrect.
- Book titles were multi-line fields and you could end up with newlines in them when pasting into them etc.
- Book titles were always using the new `0xD4` header packet when saving the book, but some servers (eg. POL) still use the `0x93` packet even with new client versions. This would cause book titles to not save. CUO will use the same type of packet as sent by the server now.
- Reduced amount of unnecessary book opening sounds being played when pressing the enter key. Some unnecessary sounds still play.
- Some refactoring to make the code easier to read.

Many bugs still remain with text books. I'll write them up in a bug report. Frankly I think the book code should be largely rewritten; I found it incredibly difficult to understand and there's probably a dozen different bugs lurking in there. A different design structure would make it much more robust.

Basically, reading books mostly works now. Editing books has many outstanding bugs.

### DIFF
--- a/src/Game/UI/Gumps/BookGump.cs
+++ b/src/Game/UI/Gumps/BookGump.cs
@@ -23,6 +23,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using ClassicUO.Data;
 using ClassicUO.Game.Managers;
 using ClassicUO.Game.UI.Controls;
@@ -51,7 +52,7 @@ namespace ClassicUO.Game.UI.Gumps
         private sbyte _AtEnd;
 
         private bool _scale;
-        public MultiLineBox BookTitle, BookAuthor;
+        public TextBox BookTitle, BookAuthor;
 
         private GumpPic m_Forward, m_Backward;
         public bool[] PageChanged;
@@ -167,7 +168,7 @@ namespace ClassicUO.Game.UI.Gumps
                     Width = 155,
                     IsEditable = IsEditable,
                     Text = text,
-                    MaxLines = 8
+                    MaxLines = MaxBookLines
                 };
                 Add(tbox, page);
                 m_Pages.Add(tbox);
@@ -213,19 +214,15 @@ namespace ClassicUO.Game.UI.Gumps
         private void SetActivePage(int page)
         {
             page = Math.Min(Math.Max(page, 1), MaxPage); //clamp the value between 1..MaxPage
-            if (page <= 1)
+            if (page != ActivePage)
             {
-                page = 1;
+                Client.Game.Scene.Audio.PlaySound(0x0055);
             }
-            else if (page >= MaxPage)
-            {
-                page = MaxPage;
-            }
+
             ActivePage = page;
             UpdatePageButtonVisibility();
 
-            Client.Game.Scene.Audio.PlaySound(0x0055);
-
+            //Non-editable books only have page data sent for currently viewed pages
             if (!IsEditable)
             {
                 int leftPage = (page - 1) << 1;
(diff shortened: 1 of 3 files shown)


src/Game/UI/Gumps/BookGump.cs
```diff
@@ -63,30 +64,38 @@ public BookGump(uint serial) : base(serial, 0)
...
+                return pages;
+            }
             set
             {
                 if (value != null)
                 {
                     if (_activated > 0)
                     {
-                        for (int i = 0; i < m_Pages.Count; i++)
+                        for (int i = 0; i < Math.Min(m_Pages.Count, value.Length); i++)
```
andreakarasho: Doing the math.min for each iteration it's a bit overheaded or is there a particular reason?

src/Network/Packet.cs
```diff
@@ -199,6 +199,41 @@ public string ReadUTF8StringSafe()
             return sb.ToString();
         }
 
+        public string ReadUTF8StringSafe(int length)
```
andreakarasho, line 202: I changed this code time ago, but never pushed.
I tried it a bit and looks good. I'm not sure if the length checks are really necessary here.
What do you think about? 

```csharp
        public string ReadUTF8StringSafe()
        {
            _sb.Clear();

            if (Position >= Length) 
                return Empty;

            int index = Position;

            while (index < Length)
            {
                byte b = _data[index++];

                if (b == 0)
                    break;
            }

            string s = Encoding.UTF8.GetString(_data, Position, index - Position);
          
            Seek(index);

            index = 0;

            for (int i = 0; i < s.Length && StringHelper.IsSafeChar(s[i]); i++, index++)
            {

            }

            if (index == s.Length - 1)
                return s;

            for (int i = 0; i < s.Length; i++)
            {
                if (StringHelper.IsSafeChar(s[i]))
                    _sb.Append(s[i]);
            }

            return _sb.ToString();
        }
```

src/Network/PacketHandlers.cs
```diff
@@ -2538,28 +2538,30 @@ private static void OpenBook(Packet p)
...
-                        new MultiLineBox(new MultiLineEntry(BookGump.DefaultFont, 47, 150, 150, BookGump.IsNewBookD4, FontStyle.None, 0), editable)
+                        new TextBox(new TextEntry(BookGump.DefaultFont, 47, 150, 150, BookGump.IsNewBook, FontStyle.None, 0), editable)
                         {
                             X = 40,
                             Y = 60,
                             Height = 25,
                             Width = 155,
                             IsEditable = editable,
-                            Text = oldpacket ? p.ReadASCII(60).Trim('\0') : p.ReadASCII(p.ReadUShort()).Trim('\0')
+                            Text = oldpacket ? p.ReadUTF8StringSafe(60).Trim('\0') : p.ReadUTF8StringSafe(p.ReadUShort()).Trim('\0')
```
andreakarasho: the `ReadUTF8StringSafe(X)` should never add `\0`, so probably it's unecessary the Trim('\0')

src/Network/PacketHandlers.cs
```diff
@@ -2538,28 +2538,30 @@ private static void OpenBook(Packet p)
...
-                        new MultiLineBox(new MultiLineEntry(BookGump.DefaultFont, 29, 150, 150, BookGump.IsNewBookD4, FontStyle.None, 0), editable)
+                        new TextBox(new TextEntry(BookGump.DefaultFont, 29, 150, 150, BookGump.IsNewBook, FontStyle.None, 0), editable)
                         {
                             X = 40,
                             Y = 160,
                             Height = 25,
                             Width = 155,
                             IsEditable = editable,
-                            Text = oldpacket ? p.ReadASCII(30).Trim('\0') : p.ReadASCII(p.ReadUShort()).Trim('\0')
+                            Text = oldpacket ? p.ReadUTF8StringSafe(30).Trim('\0') : p.ReadUTF8StringSafe(p.ReadUShort()).Trim('\0')
```
andreakarasho: same

andreakarasho (commented):
I totally agree with you. That class needs a refactor.
Do you want start the job?

src/Network/Packet.cs
brndd: The reason I needed a version that specifies length was because book data packets have fixed-length character arrays rather than null-terminated strings. They aren't guaranteed to end in `\0`.

brndd:
@andreakarasho 
> I totally agree with you. That class needs a refactor.
> Do you want start the job?

I don't have time to commit to rewriting it right now. If nobody else has started on it in a month or two, I will have another look at it. If you have ideas for how to better design the class but don't have time to rewrite it yourself, you can put them in an issue report and I will consider them once I have the time for this.

SaschaKP:
Tests are ok. Merged

PR #999, opened 2020-03-25
Enable/disable party/guild requests based on "Show party members" status

markdwags:
If you have the option disabled, no longer send the packet requests as they aren't needed. When enabled, start doing it again.

### DIFF
--- a/src/Game/UI/Gumps/WorldMapGump.cs
+++ b/src/Game/UI/Gumps/WorldMapGump.cs
@@ -159,6 +159,9 @@ namespace ClassicUO.Game.UI.Gumps
             TopMost = ParseBool(xml.GetAttribute("topmost"));
             FreeView = ParseBool(xml.GetAttribute("freeview"));
             _showPartyMembers = ParseBool(xml.GetAttribute("showpartymembers"));
+
+            World.WMapManager.SetEnable(_showPartyMembers);
+
             if (int.TryParse(xml.GetAttribute("zoomindex"), out int value))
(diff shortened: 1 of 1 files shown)


src/Game/UI/Gumps/WorldMapGump.cs
```diff
@@ -225,7 +228,13 @@ private void BuildOptionDictionary()
             _options["flip_map"] = new ContextMenuItemEntry("Flip map", () => { _flipMap = !_flipMap; }, true, _flipMap);
             _options["top_most"] = new ContextMenuItemEntry("TopMost", () => { TopMost = !TopMost; }, true, _isTopMost);
             _options["free_view"] = new ContextMenuItemEntry("Free view", () => { FreeView = !FreeView; }, true, FreeView);
-            _options["show_party_members"] = new ContextMenuItemEntry("Show party members", () => { _showPartyMembers = !_showPartyMembers; }, true, _showPartyMembers);
+            _options["show_party_members"] = new ContextMenuItemEntry("Show party members", () =>
+            {
+                _showPartyMembers = !_showPartyMembers;
+
+                World.WMapManager.SetEnable(_showPartyMembers);
```
andreakarasho, line 235: I think it's a good idea to add a timers reset. Just `_lastSOMETHINGTIMER = 0` or you will see a delay when you will re-enable the option

markdwags: Using `World.WMapManager.SetEnable` will trigger it immediately based on my testing.  No delay when enabling/disabling it.

PR #1001, opened 2020-03-25
Read text in some packets as UTF-8 instead of ASCII

brndd:
This could be a bad idea, but...

A lot of strings are read as ASCII (technically [CP-1252](https://en.wikipedia.org/wiki/Windows-1252) I think). For most of these, such as character names, this is alright because the official client reads them as ASCII as well. Occasionally, such as in book titles, the data is actually UTF-8 and reading it as ASCII results in garbled special characters.

Because ASCII is a subset of UTF-8, it should be safe to read ASCII text as UTF-8 instead. This changes `ReadASCII()` into `ReadUTF8StringSafe()` in a bunch of places where I thought it justified. This also includes the length-specific version of the method from #997 to replace length-specific `ReadASCII()` calls.

This doesn't change any *outgoing* packets, because sending UTF-8 data when the server doesn't expect it could be bad.

## Why?

- Less garbled text in places where server is sending UTF-8 and we're reading it as ASCII
- In the future some emulated servers (or OSI too who knows) could want to start using UTF-8 across the board, this will improve support for that.
- Why not? Since UTF-8 is backwards compatible with ASCII.

## Possible caveats

- Won't magically enable full Unicode support because some fonts, eg. the one used in paperdoll, do not support non-ASCII characters.
- SpriteFonts have caused crashes in the past when displaying characters not found in the font, and this could lead to similar things when fonts get unexpected characters -- however, those crashes can simply be fixed by making the font display eg. `?` in place of unknown characters.
- The `ReadUTF8StringSafe()` methods could be buggier or less optimized than the simple `ReadASCII()` methods.

----

Things seemed to work fine with some quick testing; if a font doesn't support some special characters, it seems to just not display them. Many fonts surprisingly do support unicode, including the login screen (so character names with unicode characters are possible there).

However, SpriteFonts have caused crashes in the past with special characters, so that's still a possibility, though those should hopefully be resolved now.

### DIFF
--- a/src/Game/Scenes/LoginScene.cs
+++ b/src/Game/Scenes/LoginScene.cs
@@ -614,7 +614,7 @@ namespace ClassicUO.Game.Scenes
 
             for (ushort i = 0; i < count; i++)
             {
-                Characters[i] = p.ReadASCII(30).TrimEnd('\0');
+                Characters[i] = p.ReadUTF8StringSafe(30);
                 p.Skip(30);
             }
         }
--- a/src/Network/Packet.cs
+++ b/src/Network/Packet.cs
@@ -136,7 +136,7 @@ namespace ClassicUO.Network
             return _sb.ToString();
         }
 
-        public string ReadASCII(int length, bool exitIfNull = false)
+        public string ReadASCII(int length)
         {
             if (EnsureSize(length))
                 return Empty;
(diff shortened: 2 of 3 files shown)


brndd:
Shortly after submitting I realized many of these fields are probably using CP-1252 in the original client, rather than actual ASCII, and that is not compatible with UTF-8. This is therefore a Bad Idea.

PR #1008, opened 2020-03-28
Zlib remove

SaschaKP:
IN TEST: removing the zlib dependency from CUO, internal tests on *nix and windows are positive, the function is PUBLIC on purpose


SaschaKP:
closed, requires more investigation on UOP files and performance-hit tests

PR #1009, opened 2020-03-29
Add a macro action to allow item use by graphic ID

binff04:
This is an essential macro action for those who play without an assistant (Razor). Since macOS users can't use Razor, this action should help significantly. To find a graphic ID you can use `-info` command. Then just use found graphic ID (usually something like 5119) in the macro.

### DIFF
--- a/src/Game/Managers/MacroManager.cs
+++ b/src/Game/Managers/MacroManager.cs
@@ -1364,6 +1364,26 @@ namespace ClassicUO.Game.Managers
                         }
                     }
                     break;
+
+                case MacroType.UseItemByGraphicId:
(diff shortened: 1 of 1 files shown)


andreakarasho:
This is a temp solution that does not reduce the gap so much between who play with and without an assist i think.
If i accept this PR i should accept whatever PR regardings the macro’s section until mac users can “play” at the same level of others.

binff04:
Yes, I agree that this is not ideal, but I consider this a minimal necessity the same way as UsePotion macro.

P.S. It reduces the gap. Did you try to play UO without the assistant when you can't even bind a key to use an item?

binff04:
It looks like this one will not be merged. I'm closing it.

PR #958, opened 2020-02-28
Fixes vendor buy/sell gumps having blank names and being reversed under certain conditions

brndd:
Fixes #956 

A server I play on sends names for items vendors sell in `0xD6` MegaCliloc packets right before the `0x3C` and `0x74` packets which open the vendor buy gump, with the item names in the `0x74` packet being blank. Under the original client, the names of items thusly sent would be displayed on the vendor gump. ClassicUO would attempt to instead display the blank names contained in the `0x74` packet.

This PR makes ClassicUO look in `World.OPL` (which is where the data from MegaCliloc packets seem to be stored) for the item's name if the name contained in the `0x74` packet is empty. If that fails, it falls back to the item's base name in `tiledata.mul` (which is also original client behaviour).

Additionally, on servers that sent zeroes for item X coordinates in the `0x74` packet, the contents of the vendor buy gump would not get properly reversed, and the first item on the list would display the price of the last item, the second item the price of the 2nd-to-last one, and so forth. This also fixes that; I am not sure if this issue was deliberate, because as things stood the list was only reversed if the X coordinate of the 1st element was non-zero. However, I tested this on the problematic server, and a freshly installed POL server, and OutlandsUO, and all three of them functioned fine with this change.

PR #1028, opened 2020-04-05
Vendor buy gump ordering... again

brndd:
The way this works has proven to be a bit strange, but I think this is getting close to the way it should be.

Through lots of testing, it seems that the order the items are supposed to be in the buy gump is:

1. Reverse the order of the item list from the `0x3C` packet.
2. **If Y coordinate of the items <= 1:** sort the items by their X coordinate in ascending order. (coordinates are also included in the `0x3C` packet)

[ServUO says that the client sorts these by their X/Y](https://github.com/ServUO/ServUO/blob/master/Server/Network/Packets.cs#L335), but that doesn't seem to always be the case. POL100 servers send the X/Y coordinates as random values > 1, and the original client doesn't use them for sorting when this is the case.

The solution here looks to work on all servers tested, including:
- ServUO 5.7
- POL100 (Modern distro)
- OutlandsUO
- UO Renaissance
- a POL98 server that sends X/Y coordinates as (0,0) for all items (item order compared to item prices used to be reversed on this server)
- an unknown emulator server that sends the `0x3C` packet in reverse order compared to all other servers and expects items to be sorted by the X coordinate (item order compared to prices was reversed on this server after #958)

### DIFF
--- a/src/Network/PacketHandlers.cs
+++ b/src/Network/PacketHandlers.cs
@@ -1034,7 +1034,11 @@ namespace ClassicUO.Network
                         continue;
                     }
 
-                    IEnumerable<Item> list = first.X > 1 ? item.Items.Reverse() : item.Items;
+                    IEnumerable<Item> list = item.Items.Reverse();
+                    if (first.Y <= 1)
+                    {
+                        list = list.OrderBy(o => o.X);
+                    }
 
                     foreach (var i in list) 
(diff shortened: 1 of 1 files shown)


brndd:
After spending the entire day debugging, I can say that

1. The issue is more complicated than I thought when writing this PR
2. This PR doesn't match the actual functionality of the original client, which seems to defy all logic
3. This PR **will still work** as a stopgap measure

brndd:
Since I've figured out the proper fix for this, I'll post that tomorrow.

PR #1062, opened 2020-05-03
Remove Zone Identifier block on plugin files.

DarkLotus:
Remove the Zone Identifier from any plugin files on windows platforms.

This will solve the "blocked" plugin issues constantly reported.


roxya:
Tested on Windows 10 1903, worked perfectly with a fresh Razor download, files were unblocked and the plugin launched without error.

PR #1063, opened 2020-05-03
add behavior to filter display of unlootable items on Outlands

IcculusC:
On UO Outlands many mobs spawn with custom hued items, rare items, or rare weapons in their inventory.  These items appear in corpses at position `(0, 0, 0)` with their `Frozen` property set to `True`.  In practice, the `Z` value of an `Item` in a container is always `0`, and it's unlikely (perhaps impossible, haven't checked) that an item will spawn at `(0, 0, 0)` naturally.  

In this PR I modified `Item.IsLootable` to include a check on the item's `X` and `Y` values both equaling `0`.  This effectively removes the items from both grid loot and the container.  I don't think this solution should have any side effects with other uses of `Item.IsLootable`, but that needs confirmation from someone with more context.

Thanks!

### DIFF
--- a/src/Game/GameObjects/Item.cs
+++ b/src/Game/GameObjects/Item.cs
@@ -229,10 +229,16 @@ namespace ClassicUO.Game.GameObjects
 
         public ref readonly StaticTiles ItemData => ref TileDataLoader.Instance.StaticData[IsMulti ? MultiGraphic : Graphic];
 
-        public bool IsLootable =>
-            ItemData.Layer != (int) Layer.Hair &&
-            ItemData.Layer != (int) Layer.Beard &&
-            ItemData.Layer != (int) Layer.Face;
+        public bool IsLootable
+        {
+            get
+            {
+                return ItemData.Layer != (int)Layer.Hair &&
+                ItemData.Layer != (int)Layer.Beard &&
+                ItemData.Layer != (int)Layer.Face &&
+                !(X == 0 && Y == 0);
+            }
+        }
 
         private static readonly DataReader _reader = new DataReader();
 


roxya:
Last time this was done, we ended up with invisible items in the player's own backpack. I think it would need to be limited to corpse containers and probably only for `Settings.GlobalSettings.ShardType == 2`,

This one was reverted: https://github.com/andreakarasho/ClassicUO/pull/795

You'll need to target the dev branch

IcculusC:
> Last time this was done, we ended up with invisible items in the player's own backpack.

I will investigate.

My goal here is more to figure this out than offer an instant solution.

roxya:
The last time we had the issue, it was possible for some items to end up at 0/0/0 in the player's backpack (100% reproducible on Outlands by buying trapped pouches from the NPC provisioner). Commit 2c461ee1bb5f1e09ecce2a1f7cddab8b05b7c6c2 is just before the revert, which shows the behaviour well.

In the current version, it doesn't seem to happen anymore. I'm guessing it's since commit https://github.com/andreakarasho/ClassicUO/commit/e42b80bdfaf06b57f21d21dd31c002e3a3de3c43 which adjusts the item X/Y based on the container bounds. I didn't get as far as working out why it leaves "unlootable" corpse items in 0/0/0 alone, but it seems to me the current container code is ready for another go at filtering them out.

IcculusC:
@roxya I did try buying pouches and they were not visible in my pack.  It seems to me that while the logic may be similar, determining whether or not to show an item in your backpack vs on a corpse are not the same condition.

I'm going to test a few more things as well.

brndd:
>it's unlikely (perhaps impossible, haven't checked) that an item will spawn at (0, 0, 0) naturally.

This is not a good assumption to make, unless you make it specific only to Outlands. There are a lot of emulators out there doing a *lot* of weird shit that you would never believe they get away with.

IcculusC:
> > it's unlikely (perhaps impossible, haven't checked) that an item will spawn at (0, 0, 0) naturally.
> 
> This is not a good assumption to make, unless you make it specific only to Outlands. There are a lot of emulators out there doing a _lot_ of weird shit that you would never believe they get away with.

totally agree... I'm of the opinion that once we find the right behavior for this we put it behind a setting (or behind the one that determines the gump style)

additionally, it does appear that an item _could_ spawn at `(0,0,0)` in RunUO and derivatives by using `AddItem` vs `DropItem`, but it only seems to happen with a very small subset of things on Outlands (trapped pouches, t8 maps, plants I think)

I'm quite confident however that it's impossible for a user to drop an item at `(0,0,0)` as long as the container bounds are defined.

andreakarasho:
Unfortunately this behavior can broke things and makes items invisible in certain situations.
Can’t accept sorry

PR #1064, opened 2020-05-03
Fix Body.def and Corpse.def file override conflict resolution

### LINKED ISSUE #1061: Body.def (and probably other .def files) override priorities
I'm just writing this bug up here so that I don't forget about it myself if I don't get to it tomorrow.

`Body.def` is used to override some animations with different animations, apparently. It's possible to define multiple overrides for the same animation in the same file. This probably shouldn't be done, but some servers with customization do this anyway, likely in error.

If the same override is defined multiple times, like here (excerpt from a real example from a real server, line numbers added by me for clarity):
```
[line 312]
669 {909} 0
[line 425]
669 {447} 0
[line 427]
669 {710} 0
```

The original client will use the **first** override and ignore all overrides after it.

ClassicUO will instead assign all the overrides as it gets to them (reading the file top-to-bottom), in effect using the **last** override. This will result in items getting different graphics than on the original client.

## suggested fixes

1. A simple, dumb fix: read `body.def` (and probably other .def files) backwards, so the first override in the file applies instead of the last one

2. A complicated, dumb fix: put some sort of a variable flag in the `IndexAnimation`, such as `AlreadyReadFromDefFileDontAssignIntoMeAnymore`, and if that flag is true then skip over the entry. https://github.com/andreakarasho/ClassicUO/blob/master/src/IO/Resources/AnimationsLoader.cs#L617

brndd:
Fixes #1061

If these .def files contain multiple conflicting overrides for the same thing, the original client seems to pick the **first** override and use that. ClassicUO would process them all, in effect picking the **last** override.

I don't know if this is the most elegant solution to the problem, but processing files backwards is non-trivial so this is probably better.

### DIFF
--- a/src/IO/Resources/AnimationsLoader.cs
+++ b/src/IO/Resources/AnimationsLoader.cs
@@ -614,9 +614,13 @@ namespace ClassicUO.IO.Resources
                                 if (checkIndex >= Constants.MAX_ANIMATIONS_DATA_INDEX_COUNT)
                                     continue;
 
+                                if (DataIndex[index].ReadFromBodyDef)
+                                    break;
+
                                 DataIndex[index].Graphic = (ushort)checkIndex;
                                 DataIndex[index].Color = (ushort)color;
                                 DataIndex[index].IsValidMUL = true;
+                                DataIndex[index].ReadFromBodyDef = true;
 
                                 break;
                             }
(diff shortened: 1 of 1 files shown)


brndd:
I'm including a bandaid fix for #1072 in this, because this PR is quite closely related. ~~This is almost definitely not the actual logic the client is using. However, here's my thinking:~~

1. ~~The results this produces are *closer* to the desired result than what's currently there.~~
2. ~~If this ends up producing incorrect results and someone reports it, we'll be one step closer to figuring out the real way this is supposed to work.~~

Apparently this is *literally* the exact way it is supposed to work; the 4 different values are possibly some leftover debugging stuff or a scrapped feature from OSI, and the client just picks the 3rd value if it's there, and the 1st value if it isn't.

Big thanks to Grimoric from the ServUO forums for his help on this.

PR #1088, opened 2020-05-22
Fix ommiting some unicode characters

janinko:
Some unicode characters like have value with lower byte 0x0D.
Converting `si` to byte left only 0x0D and then wrongly assumed
it's a carriage return.

For example letter č has value 269 or 0x010D.


janinko:
finally correct branch I hope :D

PR #1116, opened 2020-06-08
Fix crash when rendering nonASCII character as ASCII

janinko:
When character > 255 was being printed using ASCII fonts it caused
IndexOutOfRangeException to be thrown. This will prevent the crash
by casting the char to byte (like it did before) before calculating
font index.


andreakarasho:
thanks!

PR #1171, opened 2020-08-04
Fix Mac building

KevinEady:
# Mac Building
There was a failure on [the copy commands](https://github.com/andreakarasho/ClassicUO/blob/a870d1ea5217b371ec8b1ade9bfd51e4ce14fc8a/src/ClassicUO.csproj#L104-L106) on Mac OS X, you can't use quotes with `*`, so the copying fails:

```
> ls "*"
ls: *: No such file or directory
```

It works if you remove the quotes around `$(SolutionDir)....` but then it fails if there is a space in the directory. I first tried `pushd` / `popd` which succeeded on Mac but failed on Linux, so now I have a `cd` with no pushing, which I guess is fine.

# Native Libs Import

It appears using the relative paths in the app config works. Built and ran via `msbuild && mono .\bin\Debug\ClassicUO.exe`:

## Linux

dllmap-fix-linux

## Mac

dllmap-fix-mac

## Windows

dllmap-fix-windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,8 @@ Follow instructions on https://brew.sh/
   - Release version: `msbuild /t:Rebuild /p:Configuration=Release`
 
 7. Run ClassicUO via Mono:
-  - Debug version: `DYLD_LIBRARY_PATH=./bin/Debug/osx/ mono ./bin/Debug/ClassicUO.exe`
-  - Release version: `DYLD_LIBRARY_PATH=./bin/Release/osx/ mono ./bin/Release/ClassicUO.exe`
+  - Debug version: `mono ./bin/Debug/ClassicUO.exe`
+  - Release version: `mono ./bin/Release/ClassicUO.exe`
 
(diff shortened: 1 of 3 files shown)


andreakarasho:
Editing the FNA.dll.config can cause issues in different unix OS. 
I’ve directed experience about.
CUO must run with this https://github.com/flibitijibibo/MonoKickstart
Infact each release contains the portable mono runtime. 
Instead it’s interesting the post build process issue (probably it’s the only thing i can accept in this PR)

ghost:
Another solution to the "\*.\*" which doesn't require a `cd`:

      cp -Rf "$(SolutionDir)external/lib64/"*.*  "$(OutputPath)lib64/"
      cp -Rf "$(SolutionDir)external/osx/"*.*    "$(OutputPath)osx/"
      cp -Rf "$(SolutionDir)external/x64/"*.*    "$(OutputPath)x64/"

andreakarasho:
With new csproj format i fixed the build copy error so i think i can close this pr

PR #1180, opened 2020-08-23
Add custom profiles path

ghost:
I added the ability to specify `profilespath` command line parameter.  The motivation is to separate program from customization.  This allows multiple installs to share the same set of profiles and also simplifies keeping this data backed up.

The code was refactored so that the path to the profile directory is calculated once and not for every module.

### DIFF
--- a/src/Configuration/Profile.cs
+++ b/src/Configuration/Profile.cs
@@ -287,31 +287,10 @@ namespace ClassicUO.Configuration
         public string WorldMapHiddenMarkerFiles { get; set; } = string.Empty;
 
 
-        internal static string ProfilePath { get; } = Path.Combine(CUOEnviroment.ExecutablePath, "Data", "Profiles");
-        internal static string DataPath { get; } = Path.Combine(CUOEnviroment.ExecutablePath, "Data");
-
         public static uint GumpsVersion { get; private set; }
 
-        public void Save(List<Gump> gumps = null)
+        public void Save(string path, List<Gump> gumps = null)
         {
-            if (string.IsNullOrEmpty(ServerName))
-            {
-                throw new InvalidDataException();
-            }
-
-            if (string.IsNullOrEmpty(Username))
-            {
-                throw new InvalidDataException();
-            }
-
-            if (string.IsNullOrEmpty(CharacterName))
-            {
-                throw new InvalidDataException();
-            }
-
-            string path = FileSystemHelper.CreateFolderIfNotExists
-                (ProfilePath, Username.Trim(), ServerName.Trim(), CharacterName.Trim());
-
             Log.Trace($"Saving path:\t\t{path}");
 
             // Save profile settings
@@ -368,11 +347,8 @@ namespace ClassicUO.Configuration
             SkillsGroupManager.Save();
         }
 
-        public List<Gump> ReadGumps()
+        public List<Gump> ReadGumps(string path)
         {
-            string path = FileSystemHelper.CreateFolderIfNotExists
-                (ProfilePath, Username.Trim(), ServerName.Trim(), CharacterName.Trim());
-
             List<Gump> gumps = new List<Gump>();
 
 
--- a/src/Configuration/ProfileManager.cs
+++ b/src/Configuration/ProfileManager.cs
@@ -30,14 +30,24 @@ namespace ClassicUO.Configuration
     internal static class ProfileManager
     {
         public static Profile CurrentProfile { get; private set; }
+        public static string ProfilePath { get; private set; }
 
         public static void Load(string servername, string username, string charactername)
         {
-            string path = FileSystemHelper.CreateFolderIfNotExists
-                (CUOEnviroment.ExecutablePath, "Data", "Profiles", username, servername, charactername);
-
+            string rootpath;
+            if (string.IsNullOrWhiteSpace(Settings.GlobalSettings.ProfilesPath))
+            {
+                rootpath = Path.Combine(CUOEnviroment.ExecutablePath, "Data", "Profiles");
+            }
+            else
+            {
+                rootpath = Settings.GlobalSettings.ProfilesPath;
+            }
+            
+            string path = FileSystemHelper.CreateFolderIfNotExists(rootpath, username, servername, charactername);
             string fileToLoad = Path.Combine(path, "profile.json");
 
+            ProfilePath = path;
             CurrentProfile = ConfigurationResolver.Load<Profile>(fileToLoad) ?? new Profile();
 
             CurrentProfile.Username = username;
@@ -53,6 +63,21 @@ namespace ClassicUO.Configuration
             if (profile == null)
             {
                 return;
+            }
+
+            if (string.IsNullOrEmpty(profile.ServerName))
+            {
+                throw new InvalidDataException();
+            }
+
+            if (string.IsNullOrEmpty(profile.Username))
+            {
+                throw new InvalidDataException();
+            }
+
+            if (string.IsNullOrEmpty(profile.CharacterName))
+            {
+                throw new InvalidDataException();
             }
 
             if (profile.WindowClientBounds.X < 600)
--- a/src/Configuration/Settings.cs
+++ b/src/Configuration/Settings.cs
@@ -45,6 +45,8 @@ namespace ClassicUO.Configuration
         [JsonProperty("ultimaonlinedirectory")]
         public string UltimaOnlineDirectory { get; set; } = "";
 
+        [JsonProperty("profilespath")] public string ProfilesPath { get; set; } = string.Empty;
+
         [JsonProperty("clientversion")] public string ClientVersion { get; set; } = string.Empty;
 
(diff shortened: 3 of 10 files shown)


ctrl-alt-jorge:
Can someone check the fix to OSX build?  Currently, OSX build is failing because FNA is not in the expected directory.  The DLL is supposed to be copied to the correct directory as part of the FNA post build, but the build was never occurring.

andreakarasho:
The settings.json file is considered deprecated.
I prefer avoid to implement features linked to it.
Prefer to use launch arguments instead  If you have noticed all entries in the settings.json are accessible with args too.

Time ago i proposed to move all profile folder to local/roaming folder like cuo launcher does. In this way you can keep things without worring about location or be scared to lose macros or other settings.

ghost:
> The settings.json file is considered deprecated.
> I prefer avoid to implement features linked to it.
> Prefer to use launch arguments instead If you have noticed all entries in the settings.json are accessible with args too.
> 
> Time ago i proposed to move all profile folder to local/roaming folder like cuo launcher does. In this way you can keep things without worring about location or be scared to lose macros or other settings.

Thanks.  I changed this from an entry in `settings.json` to a command line parameter `profilesdirectory`.

src/Configuration/Profile.cs
```diff
@@ -287,31 +287,10 @@ public int
...
-            }
-
-            if (string.IsNullOrEmpty(CharacterName))
-            {
-                throw new InvalidDataException();
-            }
-
-            string path = FileSystemHelper.CreateFolderIfNotExists
-                (ProfilePath, Username.Trim(), ServerName.Trim(), CharacterName.Trim());
-
```
ghost, line 314: The path is now provided to this function.  All of the checks done here have been move to where the path string is created.

src/Configuration/Settings.cs
```diff
@@ -123,6 +125,8 @@ public void Save()
                 settingsToSave.Password = string.Empty;
             }
 
+            settingsToSave.ProfilesDirectory = string.Empty;
+
```
ghost, line 129: I don't think we want to save this.  Consider running the game in the following sequence

`classicuo.exe -profilesdirectory ~` // saves to home directory
followed by
`classicuo.exe` // should save to `pwd`, but will save to the recently specified `~` if this settings is saved

ghost:
@andreakarasho Please review again, you're not in the reviewer list.

src/Configuration/ProfileManager.cs
```diff
@@ -55,6 +65,15 @@ private static void ValidateFields(Profile profile)
                 return;
             }
 
+            if (string.IsNullOrEmpty(profile.ServerName))
+                throw new InvalidDataException();
```
deccer, line 70: Please wrap one liners with `{ }`

src/Configuration/ProfileManager.cs
```diff
@@ -55,6 +65,15 @@ private static void ValidateFields(Profile profile)
...
+            
+            if (string.IsNullOrEmpty(profile.Username))
+                throw new InvalidDataException();
```
deccer, line 75: Please wrap one liners with `{ }`

src/Configuration/ProfileManager.cs
```diff
@@ -55,6 +65,15 @@ private static void ValidateFields(Profile profile)
...
+            
+            if (string.IsNullOrEmpty(profile.CharacterName))
+                throw new InvalidDataException();
```
deccer, line 80: Please wrap one liners with `{ }`

src/Game/Managers/InfoBarManager.cs
```diff
@@ -81,11 +81,7 @@ public void Clear()
 
         public void Save()
         {
-            string path = Path.Combine
-            (
-                CUOEnviroment.ExecutablePath, "Data", "Profiles", ProfileManager.CurrentProfile.Username,
-                ProfileManager.CurrentProfile.ServerName, ProfileManager.CurrentProfile.CharacterName, "infobar.xml"
-            );
+            string path = Path.Combine(ProfileManager.ProfilePath, "infobar.xml");
```
deccer, line 84: is `.ProfilePath` now the `ProfilesDirectory`? if possible can we keep the naming consistent?

src/Game/UI/Gumps/OptionsGump.cs
```diff
@@ -2848,7 +2848,8 @@ private void Apply()
             ProfileManager.CurrentProfile.TooltipDisplayZoom = _tooltip_zoom.Value;
             ProfileManager.CurrentProfile.TooltipFont = _tooltip_font_selector.GetSelectedFont();
 
-            ProfileManager.CurrentProfile?.Save(UIManager.Gumps.OfType<Gump>().Where(s => s.CanBeSaved).Reverse().ToList());
+            ProfileManager.CurrentProfile?.Save
+                (ProfileManager.ProfilePath, UIManager.Gumps.OfType<Gump>().Where(s => s.CanBeSaved).Reverse().ToList());
```
deccer, line 2854: Maybe you could store the `ProfileManager.CurrentProfile` calls into a `var currentProfile` variable.

deccer (changes requested):

src/Game/Managers/InfoBarManager.cs
ghost: `ProfilesDirectory` is the directory containing all profiles (e.g.) `/path/to/classicUO/Data/Profiles/`.  `ProfilePath` is the path of the currently loaded profile (e.g.) `<ProfilesDirectory>/account/server/charactername/`.

deccer: Its just the naming of the two things.
So there is 
`ProfilesPath` for `./Data/Profiles` and 
`ProfilePath` for `$(ProfilesPath)/account/server/character`

Then `ProfilesDirectory` should really be `ProfilesPath`.

Sorry for being picky here.

src/Configuration/ProfileManager.cs
andreakarasho: jorge, can you apply the ``` { } ``` pattern as described by @deccer ?

j-rge: I did, haven't pushed yet.  One more item to fix.

src/Configuration/ProfileManager.cs
ghost: fixed.

src/Configuration/ProfileManager.cs
ghost: fixed

src/Configuration/ProfileManager.cs
ghost: fixed

src/Game/UI/Gumps/OptionsGump.cs
ghost: Sorry, I don't have any experience in C#.  Can you point me to documentation on this feature?  (I'm an embedded linux driver developer in C).

ghost:
Fixed most issues, commented on one.

src/Game/Managers/InfoBarManager.cs
j-rge: renamed `directory` to `path`

src/Game/UI/Gumps/OptionsGump.cs
deccer: In this file you have quite a lot of `ProfileManager.CurrentProfile.` occurences which make certain lines rather long.
Instead of calling these all the time, create a local variable `Profile currentProfile = ProfileManager.CurrentProfile;` and use that one instead.

deccer (approved):
It looks good to me, obviously would be great if we had supporting Unit-Tests ;)

And that one minor last comment from me can possibly be fixed too.

src/Game/UI/Gumps/OptionsGump.cs
ghost: Ahh, ok.  I completely misunderstood that.

ghost: Fixed.

ghost:
Fixed the last comment regarding the `currentProfile` variable.  Squashed and rebased.

src/Game/UI/Gumps/OptionsGump.cs
```diff
@@ -211,6 +211,7 @@ internal class OptionsGump : Gump
         private Checkbox _use_tooltip;
         private Checkbox _useStandardSkillsGump, _showMobileNameIncoming, _showCorpseNameIncoming;
 
+        Profile currentProfile = ProfileManager.CurrentProfile;
```
andreakarasho: missing the `private` access modifier and the name convention is wrong: missing the “_” before the name

andreakarasho (commented):
I found a little thing in optionsgump.cs.
For the rest seems fine, good job

src/Game/UI/Gumps/OptionsGump.cs
ghost: Fixed.

ghost:
Addressed the final comment

PR #1194, opened 2020-08-27
This PR adds non functional changes, code formatting/style only

deccer:
Fix unused usings
Fix some typos
Fix Spacing (to be continued)
Add Braces for one liners (to be continued)
Better variable naming
Simplification of method bodies


andreakarasho:
Everything seems good, but i prefer to see the TYPE instead of `var`

deccer:
Even if its obvious what the type is already due to proper naming of things? 

For example:
`var map = _maps[...];`
vs
`Map map = _maps[...];`

andreakarasho:
Yes, because from my pov it's easier to read. In particular in an big project like this.
I'm old school :D

deccer:
Alright, I will find the rule which wants to turn types into `var`s :)
Since some pieces of code use `var` already. Do you mind if we go with those for now?

As I get to know the code base/refactor/change/fix things and encounter `var` I will switch to the actual type then.
Deal? :)

andreakarasho:
Yes sometime me or other guys use var, but there are not so much around the code.
If we want (and we want!) apply a schema i prefer to use explicit types

deccer:
So you agree on leaving it as is for now, and we change as we see it flying around.
I modified the .DotSettings for ReSharper to not suggest to use var.

andreakarasho:
Resharper should have a option to convert all `var` to `explicit type` if i recall

deccer:
Yes it has, running it takes only a few seconds.

deccer:
Converted all vars in CUO project. Then we have one thing clarified for code styling already.

PR #1197, opened 2020-08-28
Store strings in resource files to enable localisation

janinko:
This PR should enable better localization of the client itself. It replaces the proof of concept in #1112. 

In the third commit there is czech localization. If you would prefer to not have it in your upstream, we can drop the commit and keep it in our fork only.


andreakarasho:
Can you update your branch with dev? There are some conflicts due to the recents updates in dev

janinko:
@andreakarasho branch updated

PR #1198, opened 2020-08-28
HighlightGameObjects Fix on Mobile

Alias-Alan:
Fixed it so when Current.HighlightGameObjects is checked; it will show notoriety when attacked/hover over as well as show flag changes (ispoisoned, isparalyzed, isinvul).

cleared history on gitdesk - not sure why it' kept certain ones...


andreakarasho:
This pr is a mess. Can you just push your mod?

Syrupz-UO:
Will resubmit per Quicks method to see if it will stop pushing the dev update and merge hìstory adding all this extra useless information

Alias-Alan:
https://github.com/andreakarasho/ClassicUO/pull/1198/commits/374459d5636a6ffca80431015e1f8de5266ce3f2

I'm too incompetent to realize my own incompetence, here's the fix I'll close the commit with.

PR #1200, opened 2020-08-29
Add support for multi account login.

wiscow:


src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -0,0 +1,142 @@
...
+            _maxWidth = maxWidth;
+            _maxComboHeight = maxComboHeight;
+            _font = font;
+            SetItems(items);
+        }
+
+        internal void SetItems(string[] items)
+        {
+            _items = items;
+            if (_items.Length > 0 && _arrow == null)
```
wiscow: Check for null reference here.

src/Configuration/AccountManager.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.IO;
+using System.Linq;
+
+namespace ClassicUO.Configuration
+{
+    internal static class AccountManager
+    {
+        public static List<Account> Accounts;
+
+        public static string[] GetAccountNames(string serverName)
```
wiscow: should this include port number?  I'm thinking this is probably overkill but i suppose it is possible that a given server may have different servers running on the same IP and point to different ports.

src/Configuration/AccountManager.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+        private static List<Account> LoadAccountsFromFile()
+        {
+            var accounts = ConfigurationResolver.Load<List<Account>>(PathToAccountFile()) ?? new List<Account>();
+            return accounts;
+        }
+
+        private static string PathToAccountFile()
+        {
+            string path = FileSystemHelper.CreateFolderIfNotExists(CUOEnviroment.ExecutablePath, "Data", "Profiles");
```
wiscow, line 95: Will need to adjust if the profile directory becomes custom according to: https://github.com/andreakarasho/ClassicUO/pull/1180

src/Game/Scenes/LoginScene.cs
```diff
@@ -510,8 +510,18 @@ private void NetClient_PacketReceived(object sender, Packet e)
             switch (e.ID)
             {
                 case 0xA8: // ServerListReceived
+                    //Save account information if current login step is Verifying account.
+                    //Otherwise, we are backing from character selection and we dont want to resave.
+                    if(CurrentLoginStep == LoginSteps.VerifyingAccount)
+                    {
+                        AccountManager.SaveAccount(Settings.GlobalSettings.IP, Settings.GlobalSettings.Username, Settings.GlobalSettings.Password, Settings.GlobalSettings.SaveAccount);
+                        if (!Settings.GlobalSettings.SaveAccount)
```
wiscow, line 522: This is questionable as it appears to move away from the current behavior.  I am just not sure if the current functionality to keep these fields populated if the user does not have the "Save Account" checkbox enabled and then go back to the login screen is desired.  It seems like a potential current bug, but this may be original functionality.  Could be easily removed.

src/Game/UI/Gumps/Login/LoginGump.cs
```diff
@@ -226,14 +226,21 @@ public LoginGump(LoginScene scene) : base(0, 0)
...
+            var savedAccounts = AccountManager.GetAccountNames(Settings.GlobalSettings.IP);
+            Add(_textboxAccount = new StbTextBoxCombo(savedAccounts, 5, 16, 190, false, hue: 0x034F, maxComboHeight : 100)
             {
                 X = offsetX,
                 Y = offsetY,
                 Width = 190,
                 Height = 25,
             });
 
+            _textboxAccount.OnOptionSelected += (sender, e) =>
```
wiscow: I was looking for a place to unhook this but I couldn't find a good spot.  It would probably be a good idea for memory management.

src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -30,178 +30,178 @@
...
+            Add(background = new ResizePic(0x0BB8));
+            HoveredLabel[] labels = new HoveredLabel[items.Length];
+            int index = 0;
+
+            for (int i = 0; i < items.Length; i++)
+            {
+                string item = items[i];
+
+                if (item == null)
+                    item = string.Empty;
```
deccer: Please wrap one-liners with curly braces, whenever you stumble upon them.

src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -30,178 +30,178 @@
...
+            ControlInfo.IsModal = true;
+            ControlInfo.Layer = UILayer.Over;
+            ControlInfo.ModalClickOutsideAreaClosesThisControl = true;
+            _setIndex = setIndex;
+        }
+
+        private void Label_MouseUp(object sender, MouseEventArgs e)
+        {
+            if (e.Button == MouseButtonType.Left)
+                _setIndex((int)((Label)sender).Tag);
```
deccer: Here as well... see one-liners comment above

src/Configuration/AccountManager.cs
```diff
@@ -9,42 +10,64 @@ internal static class AccountManager
     {
         public static List<Account> Accounts;
 
-        private static void Load(string serverName)
+        public static string[] GetAccountNames(string serverName)
```
deccer: I would prefer an `IReadOnlyCollection<Account>` as return value

src/Configuration/AccountManager.cs
```diff
@@ -9,42 +10,64 @@ internal static class AccountManager
...
         {
-            if (Accounts == null)
+            try
             {
-                var accounts = LoadAccountsFromFile();
-                Accounts = accounts.Where(x => x.Server == serverName).ToList();
+                Load(serverName);
+                return Accounts?.Where(x => x.Server == serverName).Select(x => x.UserName).ToArray() ?? new string[] { };
```
deccer: I see this a lot as well, `x` instead of a proper name for what x might be. Why not `account`? And in case the line gets "too long", you can break the line before `.Select` ;)

src/Configuration/AccountManager.cs
```diff
@@ -9,42 +10,64 @@ internal static class AccountManager
...
+            }
+            catch(Exception ex)
+            {
+                //Log error?
```
deccer: Yes, please.

src/Configuration/AccountManager.cs
```diff
@@ -9,42 +10,64 @@ internal static class AccountManager
...
         public static string GetAccountPassword(string serverName, string userName)
         {
-            Load(serverName);
-            return Accounts?.FirstOrDefault(x => x.UserName == userName)?.Password;
+            try
+            {
+                Load(serverName);
+                return Accounts?.FirstOrDefault(x => x.UserName == userName)?.Password;
+            }
+            catch(Exception ex)
```
deccer: Hmm why catching Exception here, same above in `GetAccountNames`, I don't think that is necessary, `Load` could gracefully return `bool` or a `Result` object, which then can be handled in these methods. What do you think?

src/Configuration/AccountManager.cs
```diff
@@ -9,42 +10,64 @@ internal static class AccountManager
...
+            {
+                //Log Error?
```
deccer: Accounts dont have to have a password, if i remember correctly, so no error, technically, right?

src/Configuration/AccountManager.cs
```diff
@@ -9,42 +10,64 @@ internal static class AccountManager
...
-            var existingRecord = Accounts.FirstOrDefault(x => x.Server == serverName && x.UserName == userName);
-            if(existingRecord == null)
+            try
             {
-                Accounts.Add(new Account() { UserName = userName, Server = serverName, Password = password }); //TODO: Figure if the pw needs to be encrypted here or before.
+                Load(serverName);
+                var existingRecord = Accounts.FirstOrDefault(x => x.Server == serverName && x.UserName == userName);
+                if (existingRecord == null)
+                {
+                    Accounts.Add(new Account() { UserName = userName, Server = serverName, Password = password });
```
deccer: Can you make `Account` immutable? ie `new Account(serverName, userName, password)`

src/Configuration/AccountManager.cs
```diff
@@ -38,21 +38,28 @@ public static string GetAccountPassword(string serverName, string userName)
...
                 }
-                ConfigurationResolver.Save<List<Account>>(Accounts, PathToAccountFile());
+                //Otherwise if there is no existing record, the user has opted to save the account, and that account has a username save the record.
+                else if (existingRecord == null && saveAccount && !string.IsNullOrWhiteSpace(userName))
+                {
+                    Accounts.Add(new Account() { UserName = userName, Server = serverName, Password = password });
+                }
+                ConfigurationResolver.Save(Accounts, PathToAccountFile());
             }
             catch(Exception ex)
```
deccer: What exceptions do you expect here? besides the usual OOM/SecurityException?

src/Configuration/AccountManager.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        private static void Load(string serverName)
+        {
+            if (Accounts == null)
+            {
+                var accounts = LoadAccountsFromFile();
+                Accounts = accounts.Where(x => x.Server == serverName).ToList();
+            }
+        }
+        private static List<Account> LoadAccountsFromFile()
```
deccer: I would return `IReadOnlyCollection<Account>` here

src/Configuration/Accounts.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace ClassicUO.Configuration
+{
+    internal class Account
```
deccer, line 3: Please make this class immutable, or at least provide a `ctor` which initializes the properties already.

src/Game/UI/Controls/Combobox.cs
```diff
@@ -75,8 +76,12 @@ public int SelectedIndex
 
                 if (_items != null)
                 {
-                    _label.Text = _items[value];
+                    //Cleanup Context Menu
+                    _contextMenu.OnItemSelected -= ItemSelectedHandler;
                     UIManager.GetGump<ComboboxContextMenu>()?.Dispose();
+                    _contextMenu = null;
+
+                    _label.Text = _items[value];
```
deccer: Those lines could be pulled out into some `CleanupContextMenu()`

src/Game/UI/Controls/Combobox.cs
```diff
@@ -117,86 +126,16 @@ protected override void OnMouseUp(int x, int y, MouseButtonType button)
...
-                ControlInfo.Layer = UILayer.Over;
-                ControlInfo.ModalClickOutsideAreaClosesThisControl = true;
-            }
-
-            private void Label_MouseUp(object sender, MouseEventArgs e)
-            {
-                if (e.Button == MouseButtonType.Left)
-                    _box.SelectedIndex = (int) ((Label) sender).Tag;
-            }
+            if (_contextMenu.Height + ScreenCoordinateY > Client.Game.Window.ClientBounds.Height) _contextMenu.Y -= _contextMenu.Height + ScreenCoordinateY - Client.Game.Window.ClientBounds.Height;
```
deccer: Please wrap one-liners with curly braces.

src/Game/UI/Controls/ComboboxContextMenu.cs
```diff
@@ -0,0 +1,80 @@
...
+            Add(background = new ResizePic(0x0BB8));
+            HoveredLabel[] labels = new HoveredLabel[items.Length];
+            int index = 0;
+
+            for (int i = 0; i < items.Length; i++)
+            {
+                string item = items[i];
+
+                if (item == null)
+                    item = string.Empty;
```
deccer, line 27: Please wrap one-liners with curly braces.

src/Game/UI/Controls/ComboboxContextMenu.cs
```diff
@@ -0,0 +1,80 @@
...
+            Height = background.Height;
+            ControlInfo.IsModal = true;
+            ControlInfo.Layer = UILayer.Over;
+            ControlInfo.ModalClickOutsideAreaClosesThisControl = true;
+        }
+
+        private void Label_MouseUp(object sender, MouseEventArgs e)
+        {
+            if (e.Button == MouseButtonType.Left)
+                OnItemSelected?.Invoke(this, (int)((Label)sender).Tag);
```
deccer, line 80: Here as well

src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -0,0 +1,142 @@
...
+            set
+            {
+                _selectedIndex = value;
+
+                if (_items != null)
+                {
+                    //Cleanup Context Menu
+                    _contextMenu.OnItemSelected -= ItemSelectedHandler;
+                    UIManager.GetGump<ComboboxContextMenu>()?.Dispose();
+                    _contextMenu = null;
```
deccer: Can this be pulled out into a `CleanupContextMenu()`?

src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -0,0 +1,142 @@
...
+                Add(_arrow = new GumpPic(_maxWidth - 6, 4, ARROW_UP, 0));
+            }
+        }
+
+        public int SelectedIndex
+        {
+            get => _selectedIndex;
+            set
+            {
+                _selectedIndex = value;
```
deccer: Maybe
```cs
if (_selectedIndex == value || value < 0)
{
    return;
}
...
```

src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -80,18 +87,27 @@ protected override void OnMouseUp(int x, int y, MouseButtonType button)
...
 			}
 
-			
-			if (x > _arrow.X && x < _arrow.X + _arrow.Width && y > _arrow.Y && y < _arrow.Y + _arrow.Height)
+			if (_arrow != null)
 			{
-				_arrow.Hue = 0;
-				TextboxComboContextMenu contextMenu = new TextboxComboContextMenu(_items, Width, 100, _font, selectedIndex => SelectedIndex = selectedIndex)
+				_arrow.Graphic = ARROW_UP;
+				if (x > _arrow.X && x < _arrow.X + _arrow.Width && y > _arrow.Y && y < _arrow.Y + _arrow.Height)
```
deccer: Hmm hmm hmm, I've seen this quite a bit, Maybe `Gump` should have a public method `bool Contains(Point point)`.

src/Game/UI/Gumps/Login/LoginGump.cs
```diff
@@ -226,14 +226,23 @@ public LoginGump(LoginScene scene) : base(0, 0)
             offsetX += 7;
 
             // Text Inputs
-            Add(_textboxAccount = new StbTextBox(5, 16, 190, false, hue: 0x034F)
+            // var savedAccounts = ProfileManager.GetAccounts(World.ServerName);
+            var savedAccounts = new string[] { "Hello World", "hello", "hello", "hello", "hello", "hello", "hello", "hello" };
```
deccer: Those two lines look bogus ;) Probably remains from testing.

I really should start merging my unit-test-branch. Then this can be covered by unit tests as well.

src/Game/UI/Gumps/Login/LoginGump.cs
```diff
@@ -226,14 +226,23 @@ public LoginGump(LoginScene scene) : base(0, 0)
...
                 X = offsetX,
                 Y = offsetY,
                 Width = 190,
                 Height = 25,
             });
 
+            _textboxAccount.OnOptionSelected += (sender, e) =>
+            {
+                // var savedPassword = ProfileManager.GetSavedPassword(World.ServerName, _textboxAccount.Text);
+                var savedPassword = "HelloWorld";
```
deccer: Also probably remains from testing around

deccer (changes requested):

src/Configuration/AccountManager.cs
wiscow: bad habits die hard :)

src/Configuration/AccountManager.cs
wiscow: that would work.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: heh, this was a copy pasta from ComboBox, normally VS would pick it up for me but i turned it off because i normally use tabs and i kept ending up with tabs everywhere. will fix.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: will fix.

src/Configuration/AccountManager.cs
wiscow: yeah i think that would work.  I normally go try/catch heavy where i don't want an unhandled exception to crash the app, but can still produce a default result.  However, you are correct, in this case the try catch can go closer to the file operations.

src/Configuration/AccountManager.cs
wiscow: correct, this was again in regards to the file operation. If the try/catch is moved closer this would make sense.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: I actually considered this.  This is a pretty standard function for game development.  I peeked in the `Utilities` folder to see if i could find anything, but no luck.  I definitely agree on this.

andreakarasho:
don't forget to pull updates from latest dev

src/Configuration/AccountManager.cs
wiscow: This was mostly in regards to file system operations.  I have now moved the try/catch closer to the source of file access.

src/Configuration/AccountManager.cs
wiscow: This one is tricky.  I have since made the list private and the get will return the `ReadOnlyCollection`.  If it is read-only we cannot remove accounts internally if one needs to be deleted. I suppose this could be `ReadOnlyCollection`.   However, it would get turned right back into a list in the `Load` function.

src/Configuration/Accounts.cs
wiscow: yeah I'm not sure it can be 100% immutable.  Depends on the JSON Serializer.  Ill try it out and see, but i think the setters need to remain public. I'll add a constructor though for sure.

Yeah, as i suspected that is a no go. I will add an additional contstructor though.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: indeed.

wiscow: Actually. This was also some copy pasta from the `ComboBox` and now that i look at it...  None of this logic should really reside in this setter.  This is probably better handled in the event handler for the `OnItemSelected` event that i added to the `ComboboxContextMenu`.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: Upon further investigation the GumpPic has its own event for mouse down and mouse up.  Instead of doing the hitbox check i just subbed to the events on the GumpPic control itself and then the hitbox test is no longer needed.

src/Game/UI/Gumps/Login/LoginGump.cs
wiscow: I removed the anonymous function and make it named function.  Overrode disposed, and unhooked it there.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: fixed.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: fixed

src/Configuration/AccountManager.cs
wiscow: changed

src/Configuration/AccountManager.cs
wiscow: changed.

src/Configuration/AccountManager.cs
wiscow: moved the exception handling and added logging.

src/Configuration/AccountManager.cs
wiscow: moved the exception handling.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: I have removed the auto property and moved all the relevant functionality to the event handler.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: subbed to the mouse events and moved the logic there. no more need for hitbox detection.

src/Game/UI/Gumps/Login/LoginGump.cs
wiscow: removed.

src/Game/UI/Gumps/Login/LoginGump.cs
wiscow: removed.

src/Game/UI/Controls/StbTextBoxCombo.cs
wiscow: done.

src/Configuration/AccountManager.cs
deccer: If you want to remove stuff, you can always create an actual list out of the readonly one, remove the desired item, and return a readonly list again.

src/Configuration/Accounts.cs
deccer: Oki, I wonder why we have to use such a limited serializer tho ;)

src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            Text = _items == null ? string.Empty : _items[value];
+            CaretIndex = Text.Length;
+            SetKeyboardFocus();
+            OnOptionSelected?.Invoke(this, value);
+            CleanupContextMenu();
+        }
+
+        public event EventHandler<int> OnOptionSelected;
+        public event EventHandler OnBeforeContextMenu;
```
deccer: Properties should go below the constructor.

src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -0,0 +1,127 @@
...
+            _arrow.MouseDown += ArrowMouseDownHandler;
+            _arrow.MouseUp += ArrowMouseUpHandler;
+            Add(_arrow);
+        }
+
+        private void DestoryArrowButton()
+        {
+            _arrow.MouseDown -= ArrowMouseDownHandler;
+            _arrow.MouseUp -= ArrowMouseUpHandler;
+            _arrow = null;
```
deccer: `_arrow.Dispose();` should be enough.

src/Game/UI/Controls/StbTextBoxCombo.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            UIManager.Add(_contextMenu);
+        }
+
+        private void CleanupContextMenu()
+        {
+            _contextMenu.OnItemSelected -= ItemSelectedHandler;
+            UIManager.GetGump<ComboboxContextMenu>()?.Dispose();
+            _contextMenu = null;
```
deccer: `_contextMenu.Dispose();` should be enough here too

src/Game/UI/Controls/Combobox.cs
```diff
@@ -72,16 +75,23 @@ public int SelectedIndex
...
                 }
             }
         }
 
+        private void CleanupContextMenu()
+        {
+            //Cleanup Context Menu
+            _contextMenu.OnItemSelected -= ItemSelectedHandler;
+            UIManager.GetGump<ComboboxContextMenu>()?.Dispose();
+            _contextMenu = null;
```
deccer: `_contextMenu.Dispose();` should be enough

src/Configuration/Accounts.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace ClassicUO.Configuration
+{
+    internal class Account 
+    {
+        public Account()
+        {
+
```
deccer: Sorry for being picky regarding formatting. Can you please remove the empty line here?
(SonarLint/Cube/StyleCop will cry about it, if we decide to add it :D)

src/Configuration/AccountManager.cs
```diff
@@ -0,0 +1,100 @@
...
+            //There is an existing record and they have updated their password.  save the udpated password.
+            else if (existingRecord != null && existingRecord.Password != password)
+            {
+                existingRecord.UpdatePassword(password);
+            }
+            //Otherwise if there is no existing record, the user has opted to save the account, and that account has a username save the record.
+            else if (existingRecord == null && saveAccount && !string.IsNullOrWhiteSpace(userName))
+            {
+                _accounts.Add(new Account(serverName, userName, password));
+            }
```
deccer, line 50: hmm hmm what about a switch block with pattern matching?

deccer (changes requested):
Just some minor changes I believe.

src/Configuration/Accounts.cs
wiscow: yeah I'm pretty sure newtonsoft supports private setters.  I just checked the TinyJson repo and it doesn't appear to be possible.

src/Configuration/Accounts.cs
wiscow: done

PR #1213, opened 2020-09-11
Highlight Gameobject fix

Alias-Alan:
Fixes the bug where HighLightGameObjects(on) would override notoriety colors & by state (IsPoisoned, IsParalyzed, Invul)

### DIFF
--- a/src/Game/GameObjects/Views/MobileView.cs
+++ b/src/Game/GameObjects/Views/MobileView.cs
@@ -82,8 +82,15 @@ namespace ClassicUO.Game.GameObjects
 
             if (ProfileManager.Current.HighlightGameObjects && SelectedObject.LastObject == this)
             {
-                _viewHue = Constants.HIGHLIGHT_CURRENT_OBJECT_HUE;
-                HueVector.Y = 1;
+                Item item = World.Items.Get(Serial);
+
+                if (this == item)
+                {
+                	_viewHue = Constants.HIGHLIGHT_CURRENT_OBJECT_HUE;
+                	HueVector.Y = 1;
+                }
+                else
+                	_viewHue = Notoriety.GetHue(NotorietyFlag);
             }
             else if (SelectedObject.HealthbarObject == this)
             {


src/Game/GameObjects/Views/MobileView.cs
```diff
@@ -82,8 +82,15 @@ public override bool Draw(UltimaBatcher2D batcher, int posX, int posY)
 
             if (ProfileManager.Current.HighlightGameObjects && SelectedObject.LastObject == this)
             {
-                _viewHue = Constants.HIGHLIGHT_CURRENT_OBJECT_HUE;
-                HueVector.Y = 1;
+                Item item = World.Items.Get(Serial);
+
+                if (this == item)
```
roxya, line 87: I don't see that this will ever be true as this draw is only called for mobiles.

Alias-Alan: Aye, at present it was an oversight brought up in bugs channel. If you paralyze, poison, or stun a mobile, with highlightgamebjects checked this will override highlight by state instead of returning the mobiles highlight color to their notoriety forcing you to tab out/in to reset.

roxya: I'm suggesting that

`if (this == item)`

is always false, and therefore useless.

roxya:
I think that Highlight Game Objects is doing what it's expected to do. All this PR does is show notoriety instead of the highlight hue (I'm not sure that this is what we want?), I don't see how Highlight By State is affected.

Alias-Alan:
> I think that Highlight Game Objects is doing what it's expected to do. All this PR does is show notoriety instead of the highlight hue (I'm not sure that this is what we want?), I don't see how Highlight By State is affected.

"If you are actively targeting someone and stun them, the stun highlight(paralyzed color) doesn't override the target color. On other clients when you stun someone it will overide whatever target highlight and show the stun highlight." -@thunk --- The initial inquiry for the fix, not sure why highlight by state is overridden by this perhaps one or the other, if both can not be supported?

src/Game/GameObjects/Views/MobileView.cs
Alias-Alan: It seemed useful for healthbars, corpses, nametags, but I may have missed others in testing

Alias-Alan: remembering why I added this---(was made on the fly last week)

if mobile is selectedtarget and i hover over healthbar/mobile it will only show the overridden color and not the notoriety of the mobile. If you have a better method to achieve this result instead, please I am all ears @roxya

oroechimaru:
I like that target noto could be the top/visible shading , I usually take my mouse off and hadn’t noticed

andreakarasho:
It is not a bug, this is the right feature behaviour.
What you describe here: 

> If you are actively targeting someone and stun them, the stun highlight(paralyzed color) doesn't override the target color. On > other clients when you stun someone it will overide whatever target highlight and show the stun highlight.

it's purely an assistant feature that sends a custom packet to the client with the custom hue

andreakarasho:
closed: reason above

Issue #1212, opened 2020-09-04
[Discussion] Localisation continuation


PoodyCZ:
With https://github.com/andreakarasho/ClassicUO/pull/1197 merged, I tried to ask you on discord what the next step is, but it got spammed away, so here's an issue thread.
FYI, I will be speaking from my POV, translating client to Czech, but it can be any other shard, any other language there is.

We have a few options going forward.

1.  **Consider localisation "Solved" on the side of the client.** That would mean that I myself add Resource.cs.resx files to my fork of CUO, and I can simply change the default CultureInfo to CS culture. I would not commit the .cs.resx files, and only one line in Program.cs should be in conflict. Any other shard wishing to use their own translation would have to do the same.

2.  **Add translation files into CUO and allow the player to pick one** A dropdown box on login page with selection of "Supported languages". This would store one in the player's settings and on client launch, change the UI culture to the one from settings. This would be similar to what was originaly in the pull request 1197. Regarding any issues changing the culture, default would remain invariant, changing to a language would mean changing the culture. But it is by far the easiest and safest way. I have not had any problems using CS culture while preparing the PR. For any of the turkish players that had issues, for them to even have the option would mean to create a turkish translation, even then they would have the option to keep the client in english and invariant culture. But I think if any problems come up, there should be a simpler solution than "changing the culture to invariant". Many C# functions, e.g. parsing, comparisons etc, can take a culture as an argument, if you want to specify a culture-independent usage.

3. **Create a singleton resource manager** Similar to what C# generates from the .resx files into the .designer.resx, The manager would keep an information on what the selected language is, but would not change the culture. Instead, each time you wanted a localised string, you would call the manager. It would require passing the name of the key as string, so nameof(ResGumps.FPS) to keep the link to the original file

Solution #1 and #2 would add a label like this
`Label text = new Label(ResGumps.FPS, true, HUE_FONT);`
Solution #3 would probably look more like this
`Label text = new Label(ResManager.Instance().GetLocalisedString(nameof(ResGumps.FPS)), true, HUE_FONT);`

My personal choice:
Go with #2, if any problems occure, there should be a solution to that specific problem, if not, I'll eat my hat and do #3 :)

You can run this by your team of devs, or if you have any other questions or ideas, post them here.

deccer:
Its actually much easier. (hoping i didnt misunderstand anything)

Resources in c# are handled in a way that you create resx files per language.

Resources.resx (contains invariant language)
[...]

PR #1220, opened 2020-09-18
Added Localisation

PoodyCZ:
PR for issue #1212

### DIFF
--- a/src/Resources/ResGumps.cs.resx
+++ b/src/Resources/ResGumps.cs.resx
@@ -1,0 +1,141 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+    
+    Example:
+    
+    ... ado.net/XML headers & schema ...
+    <resheader name="resmimetype">text/microsoft-resx</resheader>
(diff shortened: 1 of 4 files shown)


src/Resources/ResGumps.cs.resx
```diff
@@ -0,0 +1,141 @@
...
+    </xsd:element>
+  </xsd:schema>
+  <resheader name="resmimetype">
+    <value>text/microsoft-resx</value>
+  </resheader>
+  <resheader name="version">
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
andreakarasho, line 115: Is there any reason why "System.Windows.Forms" ?

src/Resources/ResGumps.cs.resx
```diff
@@ -0,0 +1,141 @@
...
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
andreakarasho, line 118: same as above

src/Resources/ResGumps.cs.resx
PoodyCZ: That's just autogenerated header for the resx file, setting the Reader and Writer (below) to be the default ResXResourceReader.
It's the same as https://github.com/andreakarasho/ClassicUO/blob/dev/src/Resources/ResGumps.resx#L114

PR #1221, opened 2020-09-22
Mouse Input Improvements (Round 2)

jaedan:
This is best reviewed commit by commit, not as an entire unit. Each commit does exactly one thing.

This continues to clean up mouse input handling and simplifying the code. There will be additional mouse input handling clean up in the future, plus keyboard input clean up too. Ultimately this builds to a flexible keybind system for ClassicUO.

### DIFF
--- a/src/Game/GameActions.cs
+++ b/src/Game/GameActions.cs
@@ -324,12 +324,12 @@ namespace ClassicUO.Game
 
                     gump = new SplitMenuGump(item, new Point(x, y))
                     {
-                        X = Mouse.LClickPosition.X - 80,
-                        Y = Mouse.LClickPosition.Y - 40
+                        X = Mouse.Position.X - 80,
+                        Y = Mouse.Position.Y - 40
                     };
 
                     UIManager.Add(gump);
-                    UIManager.AttemptDragControl(gump, Mouse.Position, true);
+                    UIManager.AttemptDragControl(gump, true);
 
                     return true;
                 }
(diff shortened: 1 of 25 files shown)


src/Game/GameActions.cs
```diff
@@ -324,8 +324,8 @@ public static bool PickUp
 
                     gump = new SplitMenuGump(item, new Point(x, y))
                     {
-                        X = Mouse.LClickPosition.X - 80,
-                        Y = Mouse.LClickPosition.Y - 40
+                        X = Mouse.Position.X - 80,
```
andreakarasho, line 327: This has been fixed time ago because it caused lots of issues when fast dragging

jaedan: I just tried to pick up a stacked item while my mouse was moving as fast as possible and I can't get any problem to happen no matter how fast I go. This change should make the gump always appear under where your mouse currently is, as opposed to where you clicked initially. Most of the time that's the same place, but if your mouse is moving very quickly it can shift a few pixels off, and I think this behavior is better.

andreakarasho: Vbl78RdYDw

just noticed this with sallos easy grab enabled.

The fast-drag issue seems not happening.

jaedan: ok I can reproduce this. I'll get to the bottom of it.

jaedan: ok, updated the PR with this fixed. I actually reworked a bit how control dragging works and I think it's much simpler than before. I don't think previously the _dragOffset stuff really worked with easy-grab (it would set it to wherever you started dragging, which may be very far away), and there was some weird offset math with a few of the gumps.

andreakarasho:
Seems good for  me, merged

PR #1223, opened 2020-09-26
Naming Cleanup

deccer:
Match naming class/file
Fix typos
Pull types out into folders f.i. Renderer/Batching contains all commands


src/Network/Plugin.cs
```diff
@@ -681,8 +682,11 @@ private static bool UnblockFile(string fileName)
             return DeleteFile(fileName + ":Zone.Identifier");
         }
 
-        private void HandleCmdList
-            (GraphicsDevice device, IntPtr ptr, int length, Dictionary<IntPtr, GraphicsResource> resources)
+        private void HandleCmdList(
+            GraphicsDevice device,
+            IntPtr ptr,
+            int length,
+            IDictionary<IntPtr, GraphicsResource> resources)
```
andreakarasho, line 689: Why IDictionary and not a Dictionary class directly?

src/Configuration/Profile.cs
```diff
@@ -527,31 +527,31 @@ public List<Gump> ReadGumps()
 
                         try
                         {
-                            GUMP_TYPE type = (GUMP_TYPE) int.Parse(xml.GetAttribute("type"));
-                            int x = int.Parse(xml.GetAttribute("x"));
-                            int y = int.Parse(xml.GetAttribute("y"));
-                            uint serial = uint.Parse(xml.GetAttribute("serial"));
+                            GumpType type = (GumpType) int.Parse(xml.GetAttribute(nameof(type)));
```
andreakarasho, line 530: Why did you prefer the nameof approach? Not because it’s wrong, i’m not a fan of floating strings. Im just curious

src/Network/Plugin.cs
deccer: that way other `IDictionary<TKey, TValue>` impls can be passed in. SortedDict/ConcurrentDict/MyFancyDict....

src/Configuration/Profile.cs
deccer: Hmm looks like i forgot the other 3 hehe.

nameof() is just compiletime and the name is really the same, renaming stuff is easier for most IDEs - especially a VS without Resharper

deccer:
@andreakarasho What do we do with this one?

andreakarasho:
There are some conflicts. GitHub blocked me out the merge command

PR #1226, opened 2020-09-28
Cliloc Lookup Error

mschwarzgorn:
In the OPL state package for example, the value 430BA0A2 was returned for some dagger, while the GetMegaCliloc package returned 30BA0A2 as the hash code for just the same item
The client than tried to reread the megacliloc again and again for every item in reach, leading to massive traffic and later client freeze.


andreakarasho:
thanks for the PR/issue report.
PR must be addresses to 'dev' branch.
Switch to it and then we'll continue the review

andreakarasho:
can i ask you what server is it?

mschwarzgorn:
The Issue happens at Lethentir with a range of underlying clients (7.0.50.0, 7.0.74.75). ServUO codebase.

andreakarasho:
Yes i see, it happens on all servuo servers. Nice catch!
I pushed the fix directly because your PR is pretty dirty with old commits.
This is my mod:

```csharp
 public bool IsRevisionEqual(uint serial, uint revision)
        {
            if (_itemsProperties.TryGetValue(serial, out ItemProperty prop))
            {
                return (revision & ~0x40000000) == prop.Revision || // remove the mask
                       revision == prop.Revision;                   // if mask removing didn't work, try a simple compare.
            }

            return false;
        }
```

Issue #1225, opened 2020-09-28
Bounty Board Text


markdwags:
**Describe the bug**
Viewing items on the bounty board are cut off and not wrapping right.

**To Reproduce**
Go to any bank, open bounty board, open bounty

**Screenshots**

![image](https://user-images.githubusercontent.com/577652/94381003-58b64c00-00fd-11eb-8d12-c82692bb2727.png)

**Shard**
UO:Renaissance 5.0.8.3

Running latest dev build, 0.1.5.35

markdwags:
I did a little research, looks like this commit to PacketHandlers.cs is the culprit.

https://github.com/andreakarasho/ClassicUO/commit/4a7d866bdae840c7de9d586b4c9d094679a22e9e

Not proposing this as the fix but removing `len` from those 2 edits on `BulletinBoardData` corrects the data, at least for 5.0.8.3.

![image](https://user-images.githubusercontent.com/577652/94449980-517d5580-0172-11eb-9d9d-fd091f8c75ea.png)

into..

![image](https://user-images.githubusercontent.com/577652/94449992-5510dc80-0172-11eb-9b0d-754ee3327a14.png)

markdwags:
I did a little research, looks like this commit to PacketHandlers.cs is the culprit.

https://github.com/andreakarasho/ClassicUO/commit/4a7d866bdae840c7de9d586b4c9d094679a22e9e

When I remove `len` from `string putta = p.ReadUTF8StringSafe(len);` in `BulletinBoardData` corrects the data, at least for 5.0.8.3. 

Example (`string putta = p.ReadUTF8StringSafe(len);`)

![image](https://user-images.githubusercontent.com/577652/94449980-517d5580-0172-11eb-9d9d-fd091f8c75ea.png)

vs without it..

![image](https://user-images.githubusercontent.com/577652/94449992-5510dc80-0172-11eb-9b0d-754ee3327a14.png)

PR #1227, opened 2020-09-28
Fixes issue with bulletin items being cut off

markdwags:
It's using the length of the date to read each line of text resulting in cut off text, I don't believe `len` is supposed to be there per the packet spec.  Please close out if this fix isn't needed.

See https://github.com/andreakarasho/ClassicUO/issues/1225

### DIFF
--- a/src/Network/PacketHandlers.cs
+++ b/src/Network/PacketHandlers.cs
@@ -2376,7 +2376,7 @@ namespace ClassicUO.Network
 
                             if (lineLen > 0)
                             {
-                                string putta = p.ReadUTF8StringSafe();
+                                string putta = p.ReadUTF8StringSafe(lineLen);
                                 sb.Append(putta);
                                 sb.Append('\n');
                             }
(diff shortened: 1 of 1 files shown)


andreakarasho:
from POL:
```
Subcommand 2 - Message (Server)
BYTE[4] BoardID
BYTE[4] MessageID
BYTE[1] posterLen
BYTE[posterLen] poster (null terminated string)
BYTE[1] subjectLen
BYTE[subjectLen] subject (null terminated string)
BYTE[1] timeLen
BYTE[timeLen] time (null terminated string with time of posting) ("Day 1 @ 11:28")
BYTE[29] constant: (01 91 84 0A 06 1E FD 01 0B 15 2E 01 0B 17 0B 01 BB 20 46 04 66 13 F8 00 00 0E 75 00 00)
BYTE[1] numlines
For each line:
BYTE[1] linelen  <--- THIS ONE
BYTE[linelen] body (null terminated)
```
I think there is a little error in the code. This should be the right code. Basically the code read wrong variable. Now should work. Can you test it?

```csharp
for (int i = 0; i < lines; i++)
{
	len = p.ReadByte();

	if (len != 0)
	{
		string putta = p.ReadUTF8StringSafe(len);
		sb.Append(putta);
		sb.Append('\n');
	}
}
```

PR #1228, opened 2020-09-29
[NO MERGE] Example of .net core integration

jaedan:
I don't know that you'd necessarily want to merge this because it will break several plugins (especially Razor on Linux). But I thought I'd provide an example of how to do native DLL loading with .net core. I've chosen to reimplement mono's dllmap behavior since I find that convenient.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "external/FNA"]
-	path = external/FNA
-	url = https://github.com/FNA-XNA/FNA.git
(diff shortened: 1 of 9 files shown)


andreakarasho:
Unfortunately this does not work on unix systems due to the library path env variable, unless you do not specify BEFORE the env path with a .sh script

jaedan:
This works fine on Unix systems. I developed it on Linux. The path searching is all handled by the unmanaged dll loader function.

PR #1237, opened 2020-10-05
Fixes blank message on older client if three skills aren't selected

markdwags:
CliLoc 1080032 "You must have three unique skills chosen!" only exists on newer clients. At least on 5.0.8.3 the max is 1078060 until it jumps to 3000000 so unless the person selects all 3, they get a blank error message.

To note, on the the 5.0.8.3 OSI client, if you only selected one skill, it will display "You must have three unique skills chosen!", but if you choose 2, it lets you go through so perhaps the validation should only check for 2 out of the 3 instead?

I can't find "You must have three unique skills chosen!" anywhere in the older cliloc file so it must have been hard coded before they moved it to 1080032.

Close out the PR if there is a better way to handle this. 👍

### DIFF
--- a/src/Game/UI/Gumps/CharCreation/CreateCharTradeGump.cs
+++ b/src/Game/UI/Gumps/CharCreation/CreateCharTradeGump.cs
@@ -28,6 +28,7 @@ using ClassicUO.Game.GameObjects;
 using ClassicUO.Game.Managers;
 using ClassicUO.Game.UI.Controls;
 using ClassicUO.IO.Resources;
+using ClassicUO.Resources;
 
 namespace ClassicUO.Game.UI.Gumps.CharCreation
 {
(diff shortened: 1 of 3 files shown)


src/Game/UI/Gumps/CharCreation/CreateCharTradeGump.cs
```diff
@@ -250,7 +251,13 @@ private bool ValidateValues()
             }
             else
             {
-                UIManager.GetGump<CharCreationGump>()?.ShowMessage(ClilocLoader.Instance.GetString(1080032));
+                UIManager.GetGump<CharCreationGump>()
+                         ?.ShowMessage
+                         (
+                             Client.Version <= ClientVersion.CV_5090 ?
+                                 "You must have three unique skills chosen!" :
```
andreakarasho: ClassicUO now supports Resources files to manage hardcoded strings like above.
Can you add it to ResGump res file?

markdwags: Done!

PR #1243, opened 2020-10-16
Consolidate project targets to net472 and let it compile on Linux without error MSB3644

deccer:
make sure ReferenceAssemblies are pulled in for non windows platforms to eradicate

```
...Microsoft.Common.CurrentVersion.targets(1175,5): 
error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.7" 
were not found. To resolve this, install the SDK or Targeting Pack for this framework 
version or retarget your application to a version of the framework for which you have 
the SDK or Targeting Pack installed. Note that assemblies will be resolved from the 
Global Assembly Cache (GAC) and will be used in place of reference assemblies. 
Therefore your assembly may not be correctly targeted for the framework you intend.
```

and I also retargeted all projects to `net472`, not sure if `net48` will run/exist on win7

### DIFF
--- a/PackageReferences.targets
+++ b/PackageReferences.targets
@@ -1,8 +1,0 @@
-<Project Sdk="Microsoft.NET.Sdk">
-  <ItemGroup>
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies.net472" Version="1.0.0">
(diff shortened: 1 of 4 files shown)


andreakarasho:
This is huge!

Issue #1216, opened 2020-09-14
Counter Bar Enhancements


aclark13:
Suggestions for counter bar enhancements

1. Highlight on use.  I used this a lot on steam/razor counters to see which regs I needed to give someone for certain spells, and just to give myself a visual cue to what regs i'm using in a dungeon run.
2. Drag and drop items to rearrange.  If I have an item in cell 1, it'd be nice if i could drag it to cell 3 to move it.  If cell 3 is occupied, then the items would switch cells.  This could be expanded to clear items out of cells (replace or addition to alt click).  Drag items out onto the screen to clear the cell.

Issue #575, opened 2019-07-10
Counter Bar : More Flexibility & Accurate Counters


BlackTeaOutlands:
I'd like the counter bar to have some more flexibility, A) specifically able to remove stuff from it (i'm not sure how apart from copying something else over it). B) Being able to split the squares up (perhaps with Alt-Click) and move them around like spell icons. C) The number needs to be accurate, i.e. 1k+ is not good enough, suggest text size is scaled so that it will always fit.

And some "bugs", D)  "Highlight on Use" Doesn't seem to do anything. It is useful if it would highlight (perhaps by changing the background) the reagent cost of spells being cast.

aclark13:
A) Alt right click to remove
B) An option for multiple bars could be nice.  Like an "Add Bar +" in options and then you set the size of the next one.  It could technically be infinite.
C) N/A
D) Known bug, I just don't think Karasho has gotten to it yet.

PR #1245, opened 2020-10-17
Reimplementation of the counterbar

andreakarasho:
Got some suggestion from: 
#575
#1216

qKcC2d96Uw

- doubleclick on empty space will enable/disable resizing
- drag & drop items over empty space to add items in sequence [replace of the awful & useless grid system]
- auto item arrange


SaschaKP:
really really nice, I like it

K4T:
In my opinion, the last element should be moved to the new line earlier, when there is not enough space to show it as a whole (+5px margin right). Not like it is now - when last element is almost completely invisible.

Great feature anyway!

andreakarasho:
> In my opinion, the last element should be moved to the new line earlier, when there is not enough space to show it as a whole (+5px margin right). Not like it is now - when last element is almost completely invisible.
> 
> Great feature anyway!

qKcC2d96Uw

Pushed the fix :)

BlackTeaOutlands:
* Would be nice to be able to clone the window to pull different windows out (i.e. place some counters on the left of the screen, some ontop).

* Also would be nice to be able to pull one slot directly and swap places for another slot in the counter bar.

* Also having empty tiles was pretty nice I think to make it easier to get an overview (it looks like in the preview that empty slots will be gone). So maybe an option (right click) to insert an empty slot (spacer) would be nice.

binff04:
Awesome job @andreakarasho! I also have one suggestion in mind of how to improve this UI even further: add an ability to bind keys to the slots on the counter (action) bar.

One way may be to add a special macro action what will use (double click on) an item under the slot X, where X is the number of the slot in the counter bar. This way you can just bind keys, let's say 1, 2, 3 to slots 1, 2, 3, for example, and by pressing that or another key the macro action will check if there is a slot on the counter bar under the given number and it has an item and then it will just execute double click action for that item.

This will make the counter bar more like an action bar in World of Warcraft where you mostly don't bind keys to certain skills but rather to the slots on the action bar. IMHO this will simplify UO's keybinding/macro system a lot. Instead to binding keys to different items you will just bind keys to the slots and then you can just drag required items to desired slots and use them immediately (perfect for potions, equipping items, etc).

broxen:
Love this improvement! Any testing I can do to help this merge into dev?

Lichtblitz:
Is there any interest in getting this feature landed? If so, I'd be happy to look into rebasing it and maybe adding another feature or two. For example having difference counters where it shows how much more or less there is compared to a number that's been set.

Lichtblitz:
@andreakarasho Would you be comfortable with me picking up this PR and continue working on it?

See also my posting here: https://discord.com/channels/458277173208547350/674900239421014016/1395519763655954553

Lichtblitz:
@andreakarasho 
This PR was integrated and became obsolete with PR #1791

PR #1249, opened 2020-10-20
Fix: Char Creation: Color selection for skirts on females

hwirth:


hwirth:
Sorry for the messed up PR, I am totally new to git and it kept doing weird things. I didn't mean to send the PR actually, I just wanted to edit the commit message...
Well. The diffs for the two .cs files seem correct, at least.

PR #1260, opened 2020-11-25
Fix reopening container gumps in correct order

PoodyCZ:
Fixed issue with reopening gumps after login where containers would open after their child container causing child gump to close.

Added functionality to sort ContainerGumps by their new ParentSerial property to reopen in correct order.

### DIFF
--- a/src/Game/UI/Gumps/ContainerGump.cs
+++ b/src/Game/UI/Gumps/ContainerGump.cs
@@ -63,6 +63,8 @@ namespace ClassicUO.Game.UI.Gumps
                 return;
             }
 
+            ParentSerial = item.Container;
+
             Graphic = gumpid;
 
             BuildGump();
@@ -88,6 +90,8 @@ namespace ClassicUO.Game.UI.Gumps
         }
 
         public ushort Graphic { get; }
+
+        public uint ParentSerial { get; set; }
 
         public override GumpType GumpType => GumpType.Container;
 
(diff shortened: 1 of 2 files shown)


src/Game/UI/Gumps/ContainerGump.cs
```diff
@@ -89,6 +91,8 @@ public ContainerGump(uint serial, ushort gumpid, bool playsound) : base(serial,
 
         public ushort Graphic { get; }
 
+        public uint ParentSerial { get; set; }
```
andreakarasho: probably it's better to set the property as readonly. Removing the `set` is enough

PoodyCZ: Removed.

andreakarasho:
I spent a little bit of time to make this PR clean by Linq and from the "ParentSerial" object state.

This is the whole commit: https://github.com/andreakarasho/ClassicUO/commit/8fa19ec9411ef655ddb43171c7f935b230fc4f69
This is the most important part: https://github.com/andreakarasho/ClassicUO/blob/8fa19ec9411ef655ddb43171c7f935b230fc4f69/src/Configuration/Profile.cs#L337

### TODO:
- [ ] Need to verify the "You are dead." behaviour because of conversion from `LinkedList<Control>` to `LinkedList<Gump>` 
https://github.com/andreakarasho/ClassicUO/blob/8fa19ec9411ef655ddb43171c7f935b230fc4f69/src/Game/Scenes/GameScene.cs#L1150

andreakarasho:
merged: https://github.com/andreakarasho/ClassicUO/commit/90babf290d6907fbcd11174d3c352609dbc8561c

PR #1263, opened 2020-12-02
Attempt at fixing bogus speech input from hotkeys on Linux

### LINKED ISSUE #1233: ALT+character creates text in bottom input line
**Describe the bug**
Macros that use the ALT key tread the key like normal text input without modifier being pressed.

**To Reproduce**
1. Create a macro and assign a key like ALT+a
2. See that "a" appear in the bottom text input line

**Expected behavior**
Characters pressed with a modifyer key should not appear in the text line

**Information (please complete the following information):**
 - OS: Arch Linux
 - ClassicUO Version: 0.1.5.47

sorsarre:
Trying to fix #1233 here -- this thing bugs me like hell on Linux since all my HK are on Alt+something.
Somewhat naive I guess, but during some poking around it doesn't seem to mess with the editing/history hotkeys.

### DIFF
--- a/src/Game/UI/Controls/StbTextBox.cs
+++ b/src/Game/UI/Controls/StbTextBox.cs
@@ -735,7 +735,7 @@ namespace ClassicUO.Game.UI.Controls
 
         protected override void OnTextInput(string c)
         {
-            if (c == null || !IsEditable)
+            if (c == null || !IsEditable || Keyboard.Alt || Keyboard.Ctrl)
             {
                 return;
             }


andreakarasho:
Well the issue seems to be into SDL or FNA.
We can act directly here i guess and remove your mod from stdtextbox:
https://github.com/andreakarasho/ClassicUO/blob/0fe05d065c5cb2186875eed026be88bb3ed87b14/src/GameController.cs#L601

sorsarre:
So basically move that alt-or-ctrl check up to the GC's text input handling piece?

andreakarasho:
added the patch
https://github.com/andreakarasho/ClassicUO/commit/2a3dd42338532efade7c5fb5ea6a44802739da4d

PR #1274, opened 2020-12-15
Fix dropping item.Z while dropping item on desk/etc.

Attanon:
I found a little bug with dropping items on desk. If you put item on desk, it will fall out of desk and appear on the ground.  

Probably it is problem only on old sphere, which isn't fixing Z on item. I Found that if I send item z as ground, sphere doesn't fix it by items on that same spot.

### DIFF
--- a/src/Game/Scenes/GameSceneInputHandler.cs
+++ b/src/Game/Scenes/GameSceneInputHandler.cs
@@ -441,7 +441,7 @@ namespace ClassicUO.Game.Scenes
 
                             dropX = obj.X;
                             dropY = obj.Y;
-                            dropZ = obj.Z;
+                            dropZ = (sbyte)(obj.Z + it2.ItemData.Height);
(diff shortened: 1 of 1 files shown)


andreakarasho:
thanks!
i  tried with an very old sphere version and i confirm the issue.

Altought there are few considerations to do, that i already applied to my last commit that fix the issue:
https://github.com/andreakarasho/ClassicUO/commit/c131694a3413d0756a3641e11f6e98fcf2d15178

We have to consider:
- surfaces only are affected by height
- there are 2 places to fix [yeah it's horrible i know]
- a check for invalid height [0xFF]

If for you is enough i'll close the PR!

Attanon:
Ouch. It didn't occur to me. I am still new to ClassicUO codebase, but I am trying :-) 

Thank you very much for help and clarify. I close the PR.

PR #1280, opened 2020-12-30
Implement character creation feature flag restrictions

brndd:
Disables the "Next" button if you're trying to create an elf or a gargoyle when the server has disabled those in feature flags.

Right now the effect is a bit ugly, the button should grey out but it seems there's no easy way to change the hue after the button is created.

### DIFF
--- a/src/Game/UI/Controls/Button.cs
+++ b/src/Game/UI/Controls/Button.cs
@@ -223,7 +223,9 @@ namespace ClassicUO.Game.UI.Controls
         {
             UOTexture texture = GetTextureByState();
 
-            ResetHueVector();
+            int hue = IsEnabled ? 0 : 944;
+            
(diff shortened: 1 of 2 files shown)


andreakarasho:
Button.cs misses to handle the hue property here:
https://github.com/andreakarasho/ClassicUO/blob/6398eed80043821eac559debc9de91bcb18faed2/src/Game/UI/Controls/Button.cs#L228

the way to go is replacing the line above with

```ShaderHueTranslator.GetHueVector(ref HueVector, Hue, false, Alpha, true);```

then you can call: `button.Hue = somehue` to trigger the shader

brndd:
~~Well, I did it like this, automatically greying them out when the button is disabled. It's not really the focus of this PR so I didn't want to worry about making it too complicated.~~

Ah what the hell, maybe it's simpler to just add a Hue property to buttons and change that where desired. This way I'm not changing the look of all disabled buttons everywhere.

PR #1287, opened 2021-01-01
Refactor SeasonManager to be configurable and support winter

markdwags:
The SeasonManager is now configurable via a file in `Data\Client` (created by default if it doesn't exist) and allows you to define which statics and landtiles change based on the season.  The file supports both decimal and hex values.

The format is `season,type,original,replacement` so for example, to change all the green walnut leaves to red for fall, you'd define this `fall,static,0x0CE1,0x0CE2` in the file.

This PR also includes all the correct winter statics/landtiles mapped out to replace the correct grass and forest tiles so it looks like actual snow cover. It doesn't replace jungle or sand so some transitions may look weird from snow to those.


markdwags:
Closing PR to create a new one

markdwags:
Instead of dealing with merge issues, just created a new PR at https://github.com/andreakarasho/ClassicUO/pull/1293

PR #1289, opened 2021-01-06
Mac users can use 'cmd' key as macro modifier 

birdinforest:
# Feature

On mac system it is common to use 'command' key to be a modifier of hotkeys. This PR does OS checking and allows Mac users to create macro with 'command' key.

# Changes of file `macros.xml`

## Mac

A new property `cmd` will be added into `Macro` tag when users run ClassicUO at MacOS.

## Windows

There is no change on Windows.
**A corner case:** If users copy&replace a mac copy. `cmd` properties are remained, however macros using 'cmd' key won't work. **Users have to re-assigned them to make them work again.**

This is a arguable implementation. **The reason I don't remove `cmd` property:** I don't want to break user's data without any notification. I think a better solution is that to pop out a notification, suggesting users to reassign hotkeys for those effectted macros. Another solution could be popping out the macro editing GUI to force users to re-assign hotkeys. Please advice me.

# Test

I have tested on Mac and Windows 10. 

**Please backup the `macros.xml` file before test.** Although I don't think there is any risk, just in case any bug to destory your macros.

**Off-topic:** I find that the parse error will cause recreation of `macros.xml`. It is an issue in my optionion, as users will lost all of defined macros without any notification. According to my experience, users will copy macros time to time. There is chance of tag missing during copy&parse. Maybe we could create a backup before recreation? I will submit a issue later for discussion.

# OS checking variables
Three OS checking variables under namespace `Utility.Platforms.PlatformHelper` for MacOS detection.
I find that there are two usage of `SDL.SDL_GetPlatform()` and one usage of `RuntimeInformation.IsOSPlatform()` in this project. I think it is better to use uniform OS checking. So that I will commit a PR later to apply those OS checking variables globally. Please advice me if it is a good practice.

### DIFF
--- a/src/Game/Managers/MacroManager.cs
+++ b/src/Game/Managers/MacroManager.cs
@@ -253,13 +253,13 @@ namespace ClassicUO.Game.Managers
         }
 
 
-        public Macro FindMacro(SDL.SDL_Keycode key, bool alt, bool ctrl, bool shift)
+        public Macro FindMacro(SDL.SDL_Keycode key, bool alt, bool ctrl, bool shift, bool cmd = false)
         {
             Macro obj = (Macro) Items;
 
             while (obj != null)
             {
-                if (obj.Key == key && obj.Alt == alt && obj.Ctrl == ctrl && obj.Shift == shift)
+                if (obj.Key == key && obj.Alt == alt && obj.Ctrl == ctrl && obj.Shift == shift && obj.Cmd == cmd)
                 {
                     break;
                 }
@@ -1548,12 +1548,13 @@ namespace ClassicUO.Game.Managers
 
     internal class Macro : LinkedObject, IEquatable<Macro>
     {
-        public Macro(string name, SDL.SDL_Keycode key, bool alt, bool ctrl, bool shift) : this(name)
+        public Macro(string name, SDL.SDL_Keycode key, bool alt, bool ctrl, bool shift, bool cmd = false) : this(name)
         {
             Key = key;
             Alt = alt;
             Ctrl = ctrl;
             Shift = shift;
+            Cmd = cmd;
         }
 
         public Macro(string name)
@@ -1567,6 +1568,7 @@ namespace ClassicUO.Game.Managers
         public bool Alt { get; set; }
         public bool Ctrl { get; set; }
         public bool Shift { get; set; }
+        public bool Cmd { get; set; }
 
         public bool Equals(Macro other)
         {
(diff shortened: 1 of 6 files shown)


andreakarasho:
Correct me if im wrong:

macOS mods translation is the following:

| Any other OS | macOS |
| --- | --- |
| CTRL | CMD |
| ALT | CTRL |
| SHIFT | SHIFT |

It's necessary to write a conversion like ImGui does for osx probably:

```cpp
        ImGuiIO& io = ImGui::GetIO();
	auto shift = io.KeyShift;
	auto ctrl = io.ConfigMacOSXBehaviors ? io.KeySuper : io.KeyCtrl;
	auto alt = io.ConfigMacOSXBehaviors ? io.KeyCtrl : io.KeyAlt;
```

binff04:
@andreakarasho unfortunately you are wrong. macOS and Windows both support Ctrl, Alt, Shift keys as is, but they treat Meta key differently: on Windows Meta key is Windows (Win) key and on macOS Meta key is Command (Cmd) key.

There are more than 3 modifier keys in total: Ctrl, Alt, Shift, Meta (Win, Cmd), Fn (less used)...

"The (Sun) Meta key, Windows key, (Apple) Cmd key, and the analogous "Amiga key"(A) on Amiga computers, are usually handled equivalently. Under the GNU/Linux operating system, the desktop environment KDE calls this key Meta, while GNOME calls this key, neutrally, Super. (This is a bit confusing, since the original space-cadet keyboard and the X Window System recognize a "Meta" modifier distinct from "Super".)" https://en.wikipedia.org/wiki/Modifier_key

P.S. For SDL2 there seems to be a constant for left/right Meta (Win/Cmd) keys: KMOD_GUI https://wiki.libsdl.org/SDL_Keymod

birdinforest:
> @andreakarasho unfortunately you are wrong. macOS and Windows both support Ctrl, Alt, Shift keys as is, but they treat Meta key differently: on Windows Meta key is Windows (Win) key and on macOS Meta key is Command (Cmd) key.
> 
> There are more than 3 modifier keys in total: Ctrl, Alt, Shift, Meta (Win, Cmd), Fn (less used)...
> 
> "The (Sun) Meta key, Windows key, (Apple) Cmd key, and the analogous "Amiga key"(A) on Amiga computers, are usually handled equivalently. Under the GNU/Linux operating system, the desktop environment KDE calls this key Meta, while GNOME calls this key, neutrally, Super. (This is a bit confusing, since the original space-cadet keyboard and the X Window System recognize a "Meta" modifier distinct from "Super".)" https://en.wikipedia.org/wiki/Modifier_key
> 
> P.S. For SDL2 there seems to be a constant for left/right Meta (Win/Cmd) keys: KMOD_GUI https://wiki.libsdl.org/SDL_Keymod

I didn't know those, that is intersting.
In this RP it applys `KMOD_GUI` to identify both left&right CMD and take them to be the same.

PR #1303, opened 2021-02-12
Backpack Skin Options

### LINKED ISSUE #1269: Backpack skin options
image

broxen:
Implements Backpack Skin Options added in Classic Client Version 7.0.53.1

- Closes #1269 
- Adds Combo Box to Options Container menu
- Adds profile setting for BackpackStyle
- Adds ContainerGump definitions for Suede, Polar Bear, GhoulSkin backpacks
- Updates Backpack Gump and Backpack on Paperdoll when applying new backpack style from settings

https://user-images.githubusercontent.com/10159505/107808606-bb139000-6d1e-11eb-8ef1-354169a8b914.mp4


src/Game/UI/Gumps/OptionsGump.cs
```diff
@@ -4017,6 +4040,15 @@ private void Apply()
             _currentProfile.RelativeDragAndDropItems = _relativeDragAnDropItems.IsChecked;
             _currentProfile.HighlightContainerWhenSelected = _highlightContainersWhenMouseIsOver.IsChecked;
 
+            // TODO: Fix minimize button
```
broxen: I need to remove this comment. Minimize is now functioning.

SaschaKP: > I need to remove this comment. Minimize is now functioning.

The minimize for certain gumps was done by me back then, only thing it lacks is a different placing for iconized vs non iconized GUMPS, but this is something that I cannot do without some internal mods (they are identified as same gump as original); at this time I don't see a real need for them, but this is a different story.

Anyway, why you want different styles of backpack universally? Isn't it more easyer to maintain your own mods without such a mainstream change? is this something present in original client?

broxen: Hi SaschaKP, thank you for your review.

Yes, these additional backpack designs are featured as an option in the original client as of Publish 95 from Nov 17, 2016 (scroll to very bottom)
https://uo.com/wiki/ultima-online-wiki/publish-notes/publish-95/

andreakarasho:
Thanks for your PR!
Everything is ok, but i think it’s necessary a client version check for seeing the option in optionsgump

broxen:
> Thanks for your PR!
> Everything is ok, but i think it’s necessary a client version check for seeing the option in optionsgump

Thank you Andrea! This should be good to go then. Let me know if there is anything else I can do.

andreakarasho:
The way you put the client version check is wrong because it will make the options gump crashes when apply [null pointer exception].
Just set:
``` control.IsVisible = Client.Version >= XXX```

broxen:
> The way you put the client version check is wrong..

Ah, my mistake. I revised per your comment, but I also added some logic to hide the label if Client.Version is too low.

PR #1309, opened 2021-02-18
small / medium / large custom bars

gaechti:
this PR adds an option to have different sizes of the custom bars
small (~80%) / normal (100%) / large (~120%)

some healers that have 20+ bars open will enjoy this
4K users will enjoy the larger bars

grafik


src/Game/UI/Gumps/OptionsGump.cs
```diff
@@ -3986,6 +4007,21 @@ private void Apply()
...
             _currentProfile.CBBlackBGToggled = _customBarsBBG.IsChecked;
+
+            if (_currentProfile.CustomBarType != _customBarType.SelectedIndex)
+            {
+                _currentProfile.CustomBarType = _customBarType.SelectedIndex;
+
+                List<HealthBarGumpCustom> hbgcustom2 = UIManager.Gumps.OfType<HealthBarGumpCustom>().ToList();
+                foreach (HealthBarGumpCustom customhealthbar in hbgcustom2)
+                {
+                    UIManager.Add
```
andreakarasho, line 4018: what's happen here if these hp bars are anchored?

gaechti: they get unanchored, same with when you toggle normal <-> custom bar

I dont think theres a way to prevent this when the size changes

but it only happens when you change the type, ie option not changed nothing happens

gaechti:
closing this, seems there's no use...

finger scrossed we get UI scale in a whole at some point

PR #1310, opened 2021-02-19
Implementation of FilesOverride

Uniphix:
This feature is pretty simple.  You simply provide a file path via doing "-filesoverride or -uofilesoverride" to the file that contains a list of entries where each line is the following concept:

filesoverride.txt (or whatever it is but needs to be a UTF-8 StreamWriter/StreamReader based file
```
map0.mul={absolute path}\map0.mul
statics0.mul={absolute path}\statics0.mul
statidx0.mul={absolute path}\statidx0.mul
```

The logic behind this is simply keep the "UltimaOnlineDirectory" at the directory that contains all the default files.  So instead of having to have multiple installs of UO servers can provide an override file of files that they have customized for their server.  The player then can not have to worry about having multiple installs, they just configure ClassicUO to use a specific UO Directory as their primary install and then they can just start playing the server.

The purpose of this is for the new (yes I will be announcing this publicly soon) ConnectUO rebirth which was formerly like it's previous ConnectUO Desktop and ConnectUO 2 along with UO Gateway. 

The first release is coming soon, and I have yet to figure out how I want the public server listing to work.  (I think for now will be sorted randomly) but eventually will have some sort of logic that will seem fair.

If you have any questions just ask.  

Thanks!


andreakarasho:
link the PR to the dev branch

Uniphix:
Is that what you meant?   just point master to dev?

andreakarasho:
im not really convinced about pointing files to different ones.
What’s happen if i have a 7.0.80.0 client and i  get files from a 4.0.0 version?

And again: once the overridedfile txt has been created, that CUO installation will always read the patch so the user has to manually delete that txt to make cuo working again.

I wonder if it’s better to avoid extra files.
ClassicUO has a settings.json, we can just add a variable over there.

Uniphix:
> im not really convinced about pointing files to different ones.
> What’s happen if i have a 7.0.80.0 client and i get files from a 4.0.0 version?

First of all, I plan to specify server options to assign ClassicUO to a specific version, the server owner will then be responsible to provide those files that the version 4.0.0 is needed.

> 
> And again: once the overridedfile txt has been created, that CUO installation will always read the patch so the user has to manually delete that txt to make cuo working again.

The ONLY way that this can actually happen is if the user creates a MANUAL shortcut to the ClassicUO.exe with the -filesoverride argument passed.  

> 
> I wonder if it’s better to avoid extra files.
> ClassicUO has a settings.json, we can just add a variable over there.

What extra files?  The only extra files is managed by ConnectUO.  ConnectUO will use the -filesoverride to map server specific patches (Which are maintained by the server owner who created the server on ConnectUO), without having to have server owners uploading "A full installation" of Ultima Online (which is the ONE thing I am trying to avoid here) and having to set the UltimaOnlineDirectory, which again I don't want to do.

PR #1312, opened 2021-02-22
Improvements to NameOverheadGump

gaechti:
Improvements to NameOverheadGump:

+add red background if lasttarget
(if there is a better color than red, let me know, may conflict with red noto hue)

+keep name and backround updated on noto change


src/Game/UI/Gumps/NameOverheadGump.cs
```diff
@@ -488,6 +488,18 @@ public override void Update(double totalTime, double frameTime)
             {
                 Dispose();
             }
+            else
+            {
+                if (entity == World.Get(TargetManager.LastTargetInfo.Serial))
```
andreakarasho: You can just compare serials without calling extra functions

gaechti:
if you want a can put it all into a single commit

PR #1324, opened 2021-04-02
Fix for *nix file system mp3 names

### LINKED ISSUE #1323: Music does not work on Case-Sensitive FileSystems
Thanks to @roxya for helping me troubleshoot this on Discord.

## Issue Summary
Since *nix file systems have case sensitive file names, music will not work in the client on *nix based systems. 

This can be tested by renaming `Trinsicpos.mp3` to `trinsicpos.mp3` and running into the guarded area of Trinsic.

CUO is pulling the names of those music files from Config.txt, which lists some of those filenames in a mixed case, but not all of them.
https://github.com/andreakarasho/ClassicUO/blob/5889a2d6fb35ba5e7f06b7b1d6d70ad1ded30b01/src/IO/Resources/SoundsLoader.cs#L144

The actual files are in saved in an inconsistent Mixed Case (sample of files below)
```
GreatEarthSerpentsTheme.mp3
grizzle_dungeon.mp3
Holycity.mp3
Trinsicpos.mp3
Turfin.mp3
Yew1.mp3
```

We're neglecting to account for case here:
https://github.com/andreakarasho/ClassicUO/blob/5889a2d6fb35ba5e7f06b7b1d6d70ad1ded30b01/src/IO/Audio/UOMusic.cs#L57

## Proposed solution
Since different/future versions of the client may have varied naming schema for those files, the ideal solution may be to search the contents of that music folder for the first instance of a file with a similar name regardless of case, and store that to the list of music files with the proper file name during client load. This may be overkill, and it doesn't seem like any other parts of CUO handle file names this way.

A simpler solution is to correct the filename case for these music files manually on per file basis, but this may break compatibility with past or future clients if the naming format doesn't remain consistent.

broxen:
Closes #1323 

This PR cross references the names in the music `Config.txt` against files in the file system and pulls the proper case-sensitive name to load into the musicData array.

## Changes:
- Creates `GetTrueFileName(string)` method to iterate through `.\Music\*` folders looking for mp3s and cross-check them against the file listed in `Config.txt`, returning the "true" filename.

## Notes
- I don't have a windows machine to test this on. Presumably it works, but should probably still be verified before merge.
- `Config.txt` has an inconsistent schema, sometimes filenames have an extension, sometimes they do not. I'm removing the extension from all names before processing.
- I tested locally on Fedora 32 and it reads the files properly, even if I rename them such as `TrInsICPoS.mp3`

### DIFF
--- a/src/IO/Resources/SoundsLoader.cs
+++ b/src/IO/Resources/SoundsLoader.cs
@@ -33,7 +33,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
@@ -318,19 +317,21 @@ namespace ClassicUO.IO.Resources
             // This will list all case variants of the filename even on file systems that
             // are case sensitive.
             Regex  pattern = new Regex($"^{name}.mp3", RegexOptions.IgnoreCase);
-            string[] fileList = Directory.GetFiles(dir, "*.mp3", SearchOption.AllDirectories).Where(path => pattern.IsMatch(Path.GetFileName(path))).ToArray();
+            //string[] fileList = Directory.GetFiles(dir, "*.mp3", SearchOption.AllDirectories).Where(path => pattern.IsMatch(Path.GetFileName(path))).ToArray();
+            string[] fileList = Directory.GetFiles(dir, "*.mp3", SearchOption.AllDirectories);
+            fileList = Array.FindAll(fileList, path => pattern.IsMatch(Path.GetFileName(path)));
 
-            if (fileList.Any())
+            if (fileList != null && fileList.Length != 0)
             {
-                if (fileList.Count() > 1)
+                if (fileList.Length > 1)
                 {
                     // More than one file with the same name but different case spelling found
                     Log.Warn($"Ambiguous File reference for {name}. More than one file found with different spellings.");
                 }
 
-                Log.Debug($"Loading music:\t\t{fileList.First()}");
+                Log.Debug($"Loading music:\t\t{fileList[0]}");
 
-                return Path.GetFileName(fileList.First());
+                return Path.GetFileName(fileList[0]);
             }
             
             // If we've made it this far, there is no file with that name, regardless of case spelling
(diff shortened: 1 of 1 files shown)


src/IO/Resources/SoundsLoader.cs
```diff
@@ -298,6 +304,41 @@ private bool TryParseConfigLine(string line, out Tuple<int, string, bool> songDa
...
+
+            // Enumerate all files in the directory, using the file name as a pattern
+            // This will list all case variants of the filename even on file systems that
+            // are case sensitive.
+            Regex  pattern = new Regex($"^{name}.mp3", RegexOptions.IgnoreCase);
+            string[] fileList = Directory.GetFiles(dir, "*.mp3", SearchOption.AllDirectories).Where(path => pattern.IsMatch(Path.GetFileName(path))).ToArray();
+
+            if (fileList.Any())
+            {
+                if (fileList.Count() > 1)
```
andreakarasho: use `.Length`. The linq `Count()` function can be very expensive.

src/IO/Resources/SoundsLoader.cs
```diff
@@ -298,6 +304,41 @@ private bool TryParseConfigLine(string line, out Tuple<int, string, bool> songDa
...
+            // don't worry about subdirectories, we'll recursively search them all
+            string dir = Settings.GlobalSettings.UltimaOnlineDirectory + $"/Music";
+
+            // Enumerate all files in the directory, using the file name as a pattern
+            // This will list all case variants of the filename even on file systems that
+            // are case sensitive.
+            Regex  pattern = new Regex($"^{name}.mp3", RegexOptions.IgnoreCase);
+            string[] fileList = Directory.GetFiles(dir, "*.mp3", SearchOption.AllDirectories).Where(path => pattern.IsMatch(Path.GetFileName(path))).ToArray();
+
+            if (fileList.Any())
```
andreakarasho: just check for `fileList != null && fileList.Length != 0`.
Avoiding linq is a good practice in a game env.

src/IO/Resources/SoundsLoader.cs
broxen: Sounds good :)

src/IO/Resources/SoundsLoader.cs
broxen: Makes sense.

Given this, I'll also try to use `.Find()` instead of `.Where()` and will revise. Then I can eliminate the Linq dependency

broxen:
This should be ready to go.

PR #1325, opened 2021-04-02
Collection of minor Outlands changes

jaedan:
I think these should all be suitable for inclusion in ClassicUO. If you want me to drop any of them, just let me know.


src/Game/World.cs
```diff
@@ -576,304 +577,158 @@ GraphicEffectBlendMode blendmode
             );
         }
 
-        public static uint SearchObject(uint serial, ScanTypeObject scanType, ScanModeObject scanMode)
+        public static uint FindNearest(ScanTypeObject scanType)
```
jaedan, line 580: All of the changes here just avoid doing multiple iterations over the set of items/mobiles as it is searching.

src/Game/Constants.cs
```diff
@@ -92,7 +92,7 @@ internal static class Constants
         public const int MAX_STATIC_DATA_INDEX_COUNT = 0x10000;
         public const int MAX_LAND_TEXTURES_DATA_INDEX_COUNT = 0x4000;
         public const int MAX_GUMP_DATA_INDEX_COUNT = 0x10000;
```
jaedan, line 94: Technically, ClassicUO is right on this value. But if a server provides a file with more sounds, the old client does not crash - it just can't play them. Increasing this in ClassicUO makes those new sounds work. Outlands currently does this - we have extra sounds that you can only hear when playing with ClassicUO.

src/Game/UI/Gumps/PaperdollGump.cs
```diff
@@ -433,17 +433,24 @@ private void PartyManifest_MouseDoubleClickEvent(object sender, MouseDoubleClick
         {
             if (args.Button == MouseButtonType.Left)
             {
```
jaedan, line 435: This one is questionable for upstream, but ClassicUO already does have CUOEnvironment.IsOutlands and displays the custom status bar, so I figured why not.

PR #1330, opened 2021-04-15
Update language packet code

Merix:
Set the Language packet with the three-letter code for the language as defined in Windows


andreakarasho:
i have a branch active about https://github.com/andreakarasho/ClassicUO/tree/lang

andreakarasho:
FYI: https://github.com/andreakarasho/ClassicUO/pull/1351

PR #1336, opened 2021-04-26
Extensible Tables

jaedan:
Both seat definitions and the buff table can now optionally be loaded from config files.


andreakarasho:
Thinking if it’s worth to fill the txt with the default table if you don’t have the file yet

jaedan:
Just let me know if you want to do that - it's easy enough.

Issue #1262, opened 2020-11-26
Spell hues don't look right in CUO


BlackTeaOutlands:
Already discussed with Karasho

Spell hues (in this case flame strike), don't take on the full range of hue in CUO.

Picture is CUO vs. Steam

![image](https://user-images.githubusercontent.com/39037141/100395844-d4bc0a00-3042-11eb-8505-f5297d602204.png)

PR #1337, opened 2021-04-26
Fix hues on spell effects

jaedan:


andreakarasho:
Going to merge this as a temporary patch.
https://github.com/andreakarasho/ClassicUO/issues/1262 will back.
I prefer to show solid hues than nothing.

note: shader works, but for semi-white hues it does not. I think it’s needed a precise blendstate when is not zero hue.

PR #1339, opened 2021-04-26
Properly restore UseAbilityButtonGump on reconnect (from profiles/gumps.xml)

lvcmenezes:
Problem:
- Ability Button Gump does not restore at all after entering world. Gump data is stored properly upon logout but is never restored afterwards. This example shows 2 gumps with type 14 - one for the primary and another for the secondary ability.

```
       <anchored_group_gump matrix_w="2" matrix_h="1">
		<gump type="14" x="1351" y="695" serial="0" isprimary="True" matrix_x="0" matrix_y="0" />
		<gump type="14" x="1395" y="695" serial="0" isprimary="False" matrix_x="1" matrix_y="0" />
	</anchored_group_gump>
```

Cause(s):
(1) At the moment when user enters world and client tries to restore the gump from gumps.xml: World.Player.Abilities returns Ability.Invalid and, thus, fails to identify the ability for the previously stored gump. The first attempt to fix this involved a call to World.Player.UpdateAbilities() before this point.

(2) The first fix caused a second problem. Moving a new ability gump (taken directly from the combat book) would not remove the old one because the index would not match. I fixed this problem by setting the gump Index while it is being built, instead of setting passing it directly to the constructor.


andreakarasho:
Conflicts between your branch and the dev one prevent me to merge this PR

andreakarasho:
Still having conflicts.
Try to pull from origin dev then redo the PR with your changes. It’s necessary to have a clean PR

PR #1355, opened 2021-05-21
Make checkerTrans less horrible

PoodyCZ:
Fixed some weird behavior regarding the checkertrans gump element. It would cause elements outside of its defined space to also be transparent.

### DIFF
--- a/src/Network/PacketHandlers.cs
+++ b/src/Network/PacketHandlers.cs
@@ -6335,7 +6335,6 @@ namespace ClassicUO.Network
             int page = 0;
 
 
-            bool applyCheckerTrans = false;
             bool textBoxFocused = false;
 
             for (int cnt = 0; cnt < cmdlen; cnt++)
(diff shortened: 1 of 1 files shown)


src/Network/PacketHandlers.cs
```diff
@@ -6786,6 +6735,24 @@ bool applyTrans(int ii, int current_page)
             return gump;
         }
 
+        private static void ApplyTrans(Gump gump, int current_page, int x, int y, int width, int height)
+        {
+            int x2 = x + width;
+            int y2 = y + width;
```
andreakarasho: There is an error here i guess
```csharp
int y2 = y + height;
```

src/Network/PacketHandlers.cs
```diff
@@ -6359,8 +6358,9 @@ string[] lines
                 }
                 else if (string.Equals(entry, "checkertrans", StringComparison.InvariantCultureIgnoreCase))
                 {
-                    applyCheckerTrans = true;
                     gump.Add(new CheckerTrans(gparams), page);
+                    var checkerTrans = gump.Children.Last();
```
andreakarasho: Declaring a `new Checkertrans()` out of the `gump.Add` method  and  use it instead of call  `.Last()` will look a way better

PoodyCZ: good shout, fixed

src/Network/PacketHandlers.cs
PoodyCZ: good shout, fixed

PR #1359, opened 2021-05-23
Fixed output of empty cliloc messages.

andyten:
If the Localization message is empty, the cliloc.enu message of the corresponding number should be displayed.

### DIFF
--- a/src/IO/Resources/ClilocLoader.cs
+++ b/src/IO/Resources/ClilocLoader.cs
@@ -30,6 +30,7 @@
 
 #endregion
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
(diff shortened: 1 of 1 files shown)


src/IO/Resources/ClilocLoader.cs
```diff
@@ -90,6 +90,35 @@ public override Task Load()
                         return;
                     }
 
+                    if (_cliloc != "Cliloc.enu")
```
andreakarasho: Because CUO runs in all OS we have to ignore capitalization, so i suggest to replace this IF statement with:

```csharp

if (string.Compare(_cliloc, "cliloc.enu", StringComparison.InvariantCultureIgnoreCase) == 0)
```

andreakarasho: Sorry i was wrong: the comparisoon has to be != 0

andyten: sorry I am new to git, you can change it?

andreakarasho: just apply the modify and push it

PR #1358, opened 2021-05-23
Fix login music playing in gamescene

ghost:
Issue:

- If the server does not send PlayMusic on login (Outlands), the login music stays in the "normal" music index.
- Season change starts playing the season music, but this is loaded into the "war" music index.
- Switching between war mode and peace mode will result in login music being played in the game scene.

This would be fixed by loading the season music into the "normal" music index and playing it over login music.

PR #1360, opened 2021-05-23
Fix loginmusic playing in game scene

ghost:
Issue:

- If the server does not send PlayMusic on login (Outlands), the login music stays in the "normal" music index.
- Season change starts playing the season music, but this is loaded into the "war" music index.
- Switching between war mode and peace mode will result in login music being played in the game scene.

This would be fixed by loading the season music into the "normal" music index and playing it over login music.


ghost:
Hmm. I see no conflict error anymore (like in https://github.com/andreakarasho/ClassicUO/pull/1358) but the build check is an issue now.

ghost:
It seems that the build is now fixed.

PR #1367, opened 2021-06-09
Lighting similar to standard client

roxya:
- Implement custom shader per light type, with a texture lookup table for remapping light values
- Correct some "hue" (shader) selections
- Stopped using hueshelper to convert light.mul to 32bit - I think the remapping was making the lights brighter than in the standard client

I don't usually touch this stuff so I figured I'd do a draft PR for you to look at.


roxya:
I'm thinking instead of if/else in the shader, if I should compute all the final values and put them in a texture, so it's just the same as a hue lookup. There are only 14 types of light. The texture can be created with code to make it clear how the values are derived.

roxya:
Yeah I think it's neater now, and clear to see what each light does:
image

andreakarasho:
seems fine! good job
Yes keeping the shader cleaner as possible is the better choice because it's an hot path in terms of peformance

PR #1370, opened 2021-06-12
Memory usage optimizations

andreakarasho:
- Implemented `System.Buffers` & `System.Memory` [and relative subdeps] for net472.
- High use of `ArrayPool<T>` to rent temp buffers avoiding GC time wast and allocation time
- Added new struct `ValueStringBuffer` to create and manage strings with 0~ allocation [i got the idea from the good @kamronbatman]
- `PixelPicker` class to store compressed non zero rgb pixels instead of allocate `UOTexture::Data` field [from UltimaXNA]
- Added a `UnmanagedMemoryPool` struct that stores a linkedlist of unmanaged data. It works like `ArrayPool<T>` not sure about the usage but it's cool.

NOTE: `System.Buffers.ArrayPool<T>.Shared.Return` accept as 2nd param `clearArray`. I set it to `true` when creating `uint[]` otherwhise textures will contains garbage.


src/IO/DataReader.cs
```diff
@@ -226,19 +227,24 @@ internal string ReadASCII(int size)
         {
             EnsureSize(size);
 
-            StringBuilder sb = new StringBuilder(size);
-
-            for (int i = 0; i < size; i++)
+            ValueStringBuilder sb = new ValueStringBuilder(size);
             {
```
kamronbatman, line 233: Is this a naked context? I think this is generally not recommended unless you have a specific need for it since this can create another stack. In this particular case it doesn't look needed.

src/IO/Resources/ClilocLoader.cs
```diff
@@ -197,157 +217,165 @@ public string Translate(int clilocNum, string arg = "", bool capitalize = false)
...
 
-            bool has_arguments = arguments.Count != 0;
+            if (trueStart == -1)
+            {
+                trueStart = 0;
+            }
 
-            arguments.Add(arg);
+            // store index locations
+            Point* locations = stackalloc Point[++totalArgs];
```
kamronbatman, line 254: You brought in `Span/Memory`, so I think this can be changed to `Span<Point>` and the unsafe dropped?

src/IO/Resources/FontsLoader.cs
```diff
@@ -543,66 +546,71 @@ ushort flags
 
             ref FontCharacterData[] fd = ref _fontData[font];
 
-            StringBuilder sb = new StringBuilder();
+            int strLen = str.Length;
 
-            if (IsUsingHTML)
+            ValueStringBuilder sb = new ValueStringBuilder(strLen);
```
kamronbatman: If we can predetermine the end size of the string, we can avoid resizes, which means we generally don't need rented buffers unless the Span<char> will be large.
There is a break-even point where stackallocating a buffer is slower than calling out to the array pool. This is because for right now, stack allocating always clears the the buffer. IIRC, it's like 100KB or so which is usually plenty.
So if needed, you can do an explicit check for the size of the final string (if it is known), and either stackalloc, or rent accordingly.

src/IO/StackDataReader.cs
```diff
@@ -34,77 +47,247 @@ public StackDataReader(byte[] data, long len)
...
+
+        [MethodImpl(IMPL_OPTION)]
+        public T* CastTo<T>() where T : unmanaged
+        {
+            return (T*)PositionAddress;
+        }
+
+        public string ReadASCII(int size)
+        {
+            ValueStringBuilder sb = new ValueStringBuilder(size);
```
kamronbatman: I think it is faster to use `Encoding.ASCII.GetString()` because the conversion is vectorized. We should benchmark to see.
But that might be a .NET Core/5 optimization, not .NET Framework. 🤔

src/IO/StackDataWriter.cs
```diff
@@ -0,0 +1,451 @@
...
+        [MethodImpl(IMPL_OPTION)]
+        public void WriteUnicodeBE(string str, int length)
+        {
+            EnsureSize(length);
+
+            for (int i = 0; i < length && i < str.Length; ++i)
+            {
+                char c = str[i];
+
+                if (c != '\0')
```
kamronbatman: When I did testing on this, I found that doing an `indexOf` from a `ReadOnlySpan<char>`, slicing from that point, counting the chars and multiplying by 2, then getting an equal sized buffer and just doing `Encoding.Unicode.GetBytes()` to write directly to it was actually faster.

I highly recommend benchmarks for this kind of thing since regular for loops are slower than whatever magic they are doing under the hood (from my experience). This is also specific to .NET Core/5, I am not sure how well .NET Framework performs for encoding.

src/Network/Packets.cs
```diff
@@ -41,1708 +42,4335 @@
...
-    internal sealed class PFirstLogin : PacketWriter
-    {
-        public PFirstLogin(string account, string password) : base(0x80)
-        {
-            WriteASCII(account, 30);
-            WriteASCII(password, 30);
-            WriteByte(0xFF);
-        }
-    }
+                writer.WriteUInt8(0x20);
```
kamronbatman: Fun fact, if you benchmark this against doing the same thing but with 8byte values, you will find the latter is much faster.

src/IO/DataReader.cs
andreakarasho: done! ty

src/IO/StackDataReader.cs
andreakarasho: From my experience i met some server [POL/Sphere] that sends garbage inside the string like:
` 'C', 'L', 'A', 'S', 'S', 'I', 'C', '\0', 'U', 'O' ` but the original client understand it as `CLASSICUO`.
Using `Encoding.ASCII.GetString()` will cut the string at '\0'

src/IO/StackDataWriter.cs
andreakarasho: https://github.com/andreakarasho/ClassicUO/pull/1370/commits/9ea4049a7c4d4b4c079b305267694c55a001bb88
Done some mod, what do you think about?

src/IO/StackDataReader.cs
kamronbatman: This is true, and that is due to fixed size vs non-fixed size IIRC. If that is the case, you would need to remove all invalid chars, otherwise you can use ASCII get string.

Personally I would not support old/bugged client behavior. It is probably an edge case anyway. ModernUO doesn't support this either because bad behavior is bad...

src/IO/StackDataWriter.cs
kamronbatman: Looks great!

src/IO/StackDataWriter.cs
```diff
@@ -0,0 +1,344 @@
...
+
+        [MethodImpl(IMPL_OPTION)]
+        public void WriteASCII(string str, int length)
+        {
+            WriteString(Encoding.ASCII, str, length);
+        }
+
+
+        [MethodImpl(IMPL_OPTION)]
+        public void WriteZero(int count)
```
kamronbatman, line 295: It is more efficient to write a byte if it is 1, a ushort if 2, a uint if 4, a ulong if 8, or use `Span<T>.Fill()` if bigger.
I believe Fill uses vectors of all 0's with low level mem copies to keep it efficient.

andreakarasho: net472 does not have tthe `Fill` method :(

kamronbatman: Even with `Span<>` pulled in from nuget? 🤔 
If not then you should do it manually.. and write in chunks of 8bytes and step down accordingly to decrease the amount of looping.
Also I recommend using a directive here:

```cs
#if NET_CORE
// use fill
#else
// Use ghetto replacement
#endif
```

andreakarasho: According to the doc, the `Fill` method is part of the `System.Runtime`    ¯\_(ツ)_/¯

kamronbatman: So you might want to make a `Clear` method and do something like:
if 1 byte, write 0 byte
if 2 bytes, write 0 short
if 4 bytes, write 0 int
if 8+ bytes, write 0 long until less than 8 bytes, then write the remaining amounts.

Something like that I guess.

andreakarasho: Do not ask me what happened but today i can see the `Fill` method
this should be ok now:

```csharp
 [MethodImpl(IMPL_OPTION)]
        public void WriteZero(int count)
        {
            if (count > 0)
            {
                EnsureSize(count);

                _buffer.Slice(Position, count).Fill(0);

                Position += count;
            }
        }
```

src/IO/Resources/FontsLoader.cs
andreakarasho: makes sense. I'll check it out

PR #1385, opened 2021-07-15
Fast assign hotkeys

jaedan:
Holding ctrlr+alt shows a + sign on spell icons which can be clicked to create a hotkey.

### DIFF
--- a/src/Game/UI/Gumps/SpellbookGump.cs
+++ b/src/Game/UI/Gumps/SpellbookGump.cs
@@ -209,7 +209,6 @@ namespace ClassicUO.Game.UI.Gumps
             base.Dispose();
         }
 
-
         private void CreateBook()
         {
             _dataBox.Clear();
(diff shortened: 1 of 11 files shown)


src/Game/UI/Gumps/SpellbookGump.cs
```diff
@@ -1437,6 +1448,68 @@ public override void Update(double totalTime, double frameTime)
...
+            /// <returns></returns>
+            public override bool Draw(UltimaBatcher2D batcher, int x, int y)
+            {
+                base.Draw(batcher, x, y);
+
+                if (ShowEdit)
+                {
+                    ResetHueVector();
+
+                    UOTexture lockTexture = GumpsLoader.Instance.GetTexture(0x1086);
```
roxya, line 1467: I don't think this art exists in the standard files (not UOR or latest from EA), may need to search for something that works for most client versions

andreakarasho: https://github.com/andreakarasho/ClassicUO/blob/d5c3ded99bc3480b30d118ce2b1b49e4dd4b0f2c/src/Game/UI/Gumps/AnchorableGump.cs#L145

this is the lock texture from standard UO

jaedan: We almost need a similar system to the localization system but for art/anim/gump graphic IDs to avoid hard coding. But this graphic should be there on a normal install.

PR #1387, opened 2021-07-15
Add UI Window Manager

jaedan:
This allows for resetting UI window positions in-game

### DIFF
--- a/src/Game/UI/Gumps/OptionsGump.cs
+++ b/src/Game/UI/Gumps/OptionsGump.cs
@@ -346,6 +346,13 @@ namespace ClassicUO.Game.UI.Gumps
                 ) { ButtonParameter = 12 }
             );
 
+            Add(
+                new NiceButton(
+                    10, 10 + 30 *i++, 140, 25, ButtonAction.SwitchPage, ResGumps.WindowManager)
+                {
+                    ButtonParameter = 13
+                }
+            );
 
             Add
             (
@@ -421,6 +428,7 @@ namespace ClassicUO.Game.UI.Gumps
             BuildCounters();
             BuildInfoBar();
             BuildContainers();
+            BuildWindowManager();
             BuildExperimental();
 
             ChangePage(1);
@@ -3010,7 +3018,6 @@ namespace ClassicUO.Game.UI.Gumps
             Add(rightArea, PAGE);
         }
 
-
         private void BuildInfoBar()
         {
             const int PAGE = 10;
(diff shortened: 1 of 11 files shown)


src/Game/UI/Gumps/UiManagerGump.cs
```diff
@@ -0,0 +1,154 @@
...
+                Add(new Label(_gump.Y.ToString(), true, HUE_FONT, 250) { X = 330 });
+                //Gump Reset button
+                Add(new Button(1, 0xFAB, 0xFAC) { X = 380, ButtonAction = ButtonAction.Activate });
+            }
+
+            public override void OnButtonClick(int buttonId)
+            {
+                //Center of Game Window
+                var x = ProfileManager.CurrentProfile.GameWindowSize.X >> 1;
+                var y = ProfileManager.CurrentProfile.GameWindowSize.X >> 1;
```
roxya: Should be ProfileManager.CurrentProfile.GameWindowSize.Y

src/Game/UI/Gumps/OptionsGump.cs
```diff
@@ -3311,6 +3311,23 @@ private void BuildContainers()
...
+            )
+            {
+                ButtonParameter = -1,
+                IsSelectable = true,
+                IsSelected = true
+            };
+
+            openUiManagerGump.MouseUp += (sender, e) =>
+            {
+                UIManager.Add(new UiManagerGump());
```
roxya: I think this could have its own options page/category instead of spawning a new gump? It doesn't make much sense to be in the containers group anyway.

jaedan: Yeah - on Outlands I renamed "containers" to "interface" because we have more options in there. It doesn't make sense for vanilla ClassicUO currently.

jaedan:
Ok this is now updated and ready for review again

andreakarasho:
ClassicUO has this function to "reset" the gump to { 0, 0 }, but is not visible to the enduser [usually reopening the same gump will reset the position if that gump is out of the screen range].
https://github.com/andreakarasho/ClassicUO/blob/master/src/Game/UI/Gumps/Gump.cs#L111

Generally i prefer to avoid to add logic available to the end user and i like to automatize behaviour like this.
What do you think about?

anethus:
This feature was requested by some people in Outlands, so I introduce such possibility - to reset gums without relog

PR #1396, opened 2021-08-04
Last characters are saved per account and server

markdwags:
Instead of the last character being a global setting either via `settings.json` or a `-param`, this change makes it per server, per account and saves the list in `Data/Profiles/lastcharacter.json`

```json
[
	{
		"AccountName" : "quick",
		"ServerName" : "UOServerA",
		"LastCharacterName" : "Quick"
	},
	{
		"AccountName" : "quick1",
		"ServerName" : "UOServerB",
		"LastCharacterName" : "TestUser"
	},
	{
		"AccountName" : "quick2",
		"ServerName" : "UOServerB",
		"LastCharacterName" : "TestTwo"
	}
]
```

### DIFF
--- a/src/Main.cs
+++ b/src/Main.cs
@@ -39,6 +39,7 @@ using System.Threading;
 using ClassicUO.Configuration;
 using ClassicUO.Data;
 using ClassicUO.Game;
+using ClassicUO.Game.Managers;
 using ClassicUO.IO;
 using ClassicUO.Resources;
 using ClassicUO.Utility;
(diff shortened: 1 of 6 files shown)


src/Main.cs
```diff
@@ -336,12 +336,6 @@ private static void ReadSettingsFromArgs(string[] args)
 
                         break;
 
-                    case "lastcharactername":
-                    case "lastcharname":
-                        Settings.GlobalSettings.LastCharacterName = value;
```
andreakarasho, line 341: this logic must exist.
ClassicUOLauncher uses it per profile, so it must be handled in some way

markdwags: But why? This isn't a profile setting -- it's an account/server setting. It has nothing to do with specific profile.

It should be checked when ClassicUO knows what the account name, and server is selected.

PR #1425, opened 2021-09-20
Fix ReadASCII/WriteASCII using ASCII instead of cp-1252/iso-8859-1

brndd:
ReadASCII was reading (and WriteASCII was writing) text as actual ASCII, but Ultima Online actually uses [CP-1252](https://en.wikipedia.org/wiki/Windows-1252) (at least on English locale -- I do not know about other locales, but presumably those would not use ASCII either). This would cause non-ASCII characters such as Ä or é to appear as question marks in some contexts (eg. in the journal when included in the `name` field of the 0x1C packet).

This PR makes ReadASCII and WriteASCII use the [ISO 8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1) / Latin-1 encoding, which is *not* the same thing as CP-1252, but is the closest thing supported by the `Encoding` class on .NET Core and .NET 5. A better solution would require using a custom Encoding type, but that seems annoying and I don't feel like doing it right now. Take a look at the [ISO 8859-1 encoding class source](https://referencesource.microsoft.com/#mscorlib/system/text/latin1encoding.cs) to get an idea of what this would entail.

Note the difference below -- Latin-1 is missing some of the special characters included in CP-1252. These characters will appear blank, as far as I know. This is still better than current behaviour, where any character past 0x7F will appear as a question mark.

## CP-1252

Windows-1252-infobox

## ISO 8859-1 / Latin-1

Latin-1-infobox

### DIFF
--- a/src/ClassicUO.csproj
+++ b/src/ClassicUO.csproj
@@ -125,6 +125,10 @@
     <DataFiles_vulkan Include="$(ProjectDir)..\external\vulkan\icd.d\*.*" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
+  </ItemGroup>
+
   <Target Name="CopyExternalDeps_build" AfterTargets="Build">
     <Copy SourceFiles="@(DataFiles_x64)" DestinationFolder="$(TargetDir)\x64\" SkipUnchangedFiles="true" />
     <Copy SourceFiles="@(DataFiles_lib64)" DestinationFolder="$(TargetDir)\lib64\" SkipUnchangedFiles="true" />
--- a/src/IO/StackDataReader.cs
+++ b/src/IO/StackDataReader.cs
@@ -276,12 +276,12 @@ namespace ClassicUO.IO
 
         public string ReadASCII(bool safe = false)
         {
-            return ReadString(Encoding.ASCII, -1, 1, safe);
+            return ReadString(Encoding.GetEncoding("iso-8859-1"), -1, 1, safe);
         }
 
         public string ReadASCII(int length, bool safe = false)
         {
-            return ReadString(Encoding.ASCII, length, 1, safe);
+            return ReadString(Encoding.GetEncoding("iso-8859-1"), length, 1, safe);
         }
 
         public string ReadUnicodeBE(bool safe = false)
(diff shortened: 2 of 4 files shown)


SaschaKP:
could we use UTF8?

the codepages 1252 or 1 of iso-8859 isn't available on .net core, not without adding extra things, and usually isn't available outside certain geographic areas and in certain OS.

this is a problem we already faced and that I discussed with @andreakarasho but at the end we didn't came to a conclusion (other things at that time were more urgent), but switching to a fixed codepage is a "no no"

brndd:
> could we use UTF8?

No. UTF-8 is compatible with ASCII's 127 code points, but it's not compatible with CP-1252 (or any of the other 8-bit Windows codepages) that UO is using, because those use 255 code points.

> the codepages 1252 or 1 of iso-8859 isn't available on .net core, not without adding extra things, and usually isn't available outside certain geographic areas and in certain OS.

ISO 8859-1 [is available on .NET Core and .NET 5](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=net-5.0#list-of-encodings) but other encodings such as ISO 8859-15 unfortunately aren't -- though you wouldn't want to use 8859-15 (or any of the other 8859 encodings) anyway, because it's not compatible with CP1252 (it has the € symbol where CP1252 has the ¤ symbol, for example).

The "proper" solution would probably be to figure out what code pages UO is using for different locales (the only one I know for a fact it uses is CP-1252 with the English locale -- presumably it will use it on most European locales too though) and then manually implement Encoding classes for those locales. Since .NET is now open source you can look at the reference source for the ISO 8859-1 that I linked above for an example of what this would entail. CP-1252 is very similar to ISO 8859-1, it just replaces some rarely used control characters in the 0xC1 row with human-readable characters.

brndd:
https://www.nuget.org/packages/System.Text.Encoding.CodePages/

I figured someone would have already made a .NET 5 version of CP-1252 and sure enough there's a Microsoft package that does it. I've added that and changed ASCII to use this, and it seems to work.

SaschaKP:
without this you would crash, you have to test your code, since the codepages aren't initialized:
`Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
            AsciiEncoding = Encoding.GetEncoding(1252);`
            
            (this is a sample, providing the encoding, but codepagesEncoding needs initialization)

SaschaKP:
nevermind, I see you corrected it, ok

PR #1432, opened 2021-10-22
Fix ReadASCII/WriteASCII using ASCII instead of cp-1252

brndd:
Reopening #1425 because this is still an issue. There was apparently something wrong with this PR, but I do not know what because I have been playing with this fix for the past month and have had no issues. Double-check that the new NuGet dependency is added and loaded correctly before merging.

### DIFF
--- a/src/ClassicUO.csproj
+++ b/src/ClassicUO.csproj
@@ -125,6 +125,10 @@
     <DataFiles_vulkan Include="$(ProjectDir)..\external\vulkan\icd.d\*.*" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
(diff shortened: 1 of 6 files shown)


andreakarasho:
Does the `System.Text.Encoding.CodePages` add a new dll?
If it does, we need to add the module name to `/tools/ManifestCreator.csproj`

brndd:
Yeah it does, `System.Text.Encoding.CodePages.dll`. Perhaps that's the issue; I've added it there now.

andreakarasho:
In the past I read somewhere that calling `Encoing.GetEncoding` may be very unperformant.
It's better to use a `static` var i guess.

brndd:
I moved this to StringHelper, which handles the RegisterProvider call and caches the encoding so we don't do `GetEncoding()` calls every time.

PR #1434, opened 2021-10-25
Fix skills.idx case sensitivity on Unix

brndd:
UOFiddler saves this file as all lowercase and some servers probably therefore use the all lowercase format. We have similar workarounds for eg. Gumpart.mul. In the long term it would probably be better to see if we can support case-insensitive filenames somehow.

### DIFF
--- a/src/IO/Resources/SkillsLoader.cs
+++ b/src/IO/Resources/SkillsLoader.cs
@@ -32,6 +32,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Text;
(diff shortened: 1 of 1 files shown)


SaschaKP:
this should fix the issue: https://github.com/ClassicUO/ClassicUO/commit/cee38f33886e8f3f80661bf9336b5ad1d7242f63

brndd:
An even better solution would probably be to list all the files in the directory with `Directory.GetFiles()`, then iterate over the files converting them to lowercase and checking the desired filename against that.

PR #1436, opened 2021-10-31
Fix buy gump item names

brndd:
Some servers send the names of items in the buy gump in 0xD6 MegaCliloc packets before opening the buy gump. ClassicUO was deleting and then recreating any existing items in the packet handler for the 0x3C UpdateContainedItems packet, which was causing the names sent in the MegaCliloc packets to be lost. This would result in a bug where opening a buy gump, closing it, and then immediately opening it again would display incorrect item names.

This PR fixes that, but it may have some side effects because I am honestly not sure why we were recreating these items in the first place. It seemed pointless but perhaps it serves some unknown purpose.

### DIFF
--- a/src/Network/PacketHandlers.cs
+++ b/src/Network/PacketHandlers.cs
@@ -1955,7 +1955,7 @@ namespace ClassicUO.Network
 
                     if (container != null)
                     {
-                        ClearContainerAndRemoveItems(container, container.Graphic == 0x2006);
+                        ClearContainer(container);
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


andreakarasho:
It's necessary to delete items to avoid data duplications or fake items.
If i remember well i think 0xD6 [cliloc packet] is sent suddenly an item/container update.

brndd:
The problem is that the 0xD6 packets are sent and processed *before* we delete the items, so the data contained in them gets deleted, resulting in incorrect item names. I think simply clearing the container (as done in this PR) should be enough to avoid fake or duplicate items, and all the items seem to get deleted permanently after the shopkeeper walks out of range or something anyway.

andreakarasho:
i wonder if forcing a new Cliloc request will fix the issue

brndd:
Then we'll end up sending a bunch of extra packets for IMO no good reason when the original client doesn't.

Is it not enough to just clear the container? The buy list is populated by items in the container, so we don't need to *delete* them from existence as long as we remove them from the container.

brndd:
This is still an issue.

PR #1439, opened 2021-11-04
Fix CP-1252 in Read/WriteASCII without relying on external dependency

brndd:
This is awful. This is just the worst. But this is what we're setting ourselves up for if we don't want to (or can't, as it happens) use [System.Text.Encoding.CodePages](https://www.nuget.org/packages/System.Text.Encoding.CodePages/). The only other alternative I can think of would be to write our own CP-1252 Encoding class for System.Text.Encoding to use, but that would be *a lot* of work because those classes are pretty complicated. I don't want to do that.

The alternative is this. The code here is sinful, because this is **not** how C# wants you to use strings, but it works. I even wrote tests! As awful as this is, it should work as a stop-gap solution until the dependency issue can be sorted (possibly by upgrading the project to .NET 5/6, or by fixing whatever issue prevents us from using app.config).


src/IO/StackDataReader.cs
```diff
@@ -394,6 +399,61 @@ private string ReadString(Encoding encoding, int length, int sizeT, bool safe)
...
+                    if (StringHelper.IsSafeChar(b))
+                    {
+                        bytes.Add(b);
+                    }
+                }
+            }
+            else
+            {
+                bytes = new List<byte>(slice.ToArray());
+            }
```
andreakarasho, line 444: we could just use the `Cp1252ToUnicode` function in combo with `ValueStringBuilder` to build an unicode compliant string

src/IO/StackDataWriter.cs
```diff
@@ -282,17 +282,38 @@ public void WriteUTF8(string str, int len)
...
+        /// padding with null characters if <paramref name="length"/> is greater than the length of the string.
+        /// </summary>
+        /// <param name="str"></param>
+        /// <param name="length"></param>
         [MethodImpl(IMPL_OPTION)]
         public void WriteASCII(string str, int length)
         {
-            WriteString<byte>(StringHelper.Cp1252Encoding, str, length);
+            //Even more horrible than above.
+            var cp1252bytes = StringHelper.StringToCp1252Bytes(str);
```
andreakarasho, line 308: `UnicodeToCp1252` + `ValueStringBuilder`

brndd:
Thanks for finishing this, @andreakarasho :+1:

PR #1442, opened 2021-11-07
Font renderer

### LINKED ISSUE #1371: Allow way to scale font size
Using a 1440p monitor allows for a nice large game window. But for those of us with bad eyesight already it would be nice to have a way to make all text like 150% larger or something.

andreakarasho:
This PR aims to implement the new dynamic font renderer thanks to the new spritesheet concept partially implemented in CUO.
The main goal of this PR is to avoid constant texture creation.

TODO list:
- [x] replace overhead text [speech, mobile names, HP %]
- [x] whitespaces break the text in chunks in different multilines if the text's width is greater of `maxwidth` param
- [x] text highlight support 
- [ ] replace gumps text
- [ ] RGBA [aka Color] support into shader. This will be useful to render hued html text  
- [ ] html parser


AkaraEssex:
Hey andreakarasho I posted a ticket I believe once you implement the  RGBA [aka Color] support into shader. This will be useful to render hued html text
 html parser
steps that this may fix my issue of artifacts and my custom rss runics having  their names show instead of just a number?  Would you mind checking my ticket and delete it if this will indeed fix it anyways.
Thanks for your time
-AkaraEssex

PR #1446, opened 2021-11-14
Make speech maximum width a configurable value

ghost:
Currently the speech width is hardcoded.

With larger screens, different fonts etc. it might be desirable to be able to configure how much text is on one line.

### DIFF
--- a/src/Configuration/Profile.cs
+++ b/src/Configuration/Profile.cs
@@ -70,6 +70,7 @@ namespace ClassicUO.Configuration
         public bool ForceUnicodeJournal { get; set; }
         public bool IgnoreAllianceMessages { get; set; }
         public bool IgnoreGuildMessages { get; set; }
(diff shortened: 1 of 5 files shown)


andreakarasho:
Speech width is hardcoded because it denies to the user to flood the server with huge messages

ghost:
> Speech width is hardcoded because it denies to the user to flood the server with huge messages

Thanks for the reply!

My understanding is that my changes should not change the amount of characters a player can send to the server. I will attach some images to show what I mean.

This is with speech width `200`:

image

This is with speech width `1000`:

image

The number of characters I am allowed to send stays the same, only the "speech bubble" width is configurable, with the speech width setting under speech tab in options:

image

From my experience, the small speech bubbles can appear more spammy on large resolutions, since the text takes a lot of space on a small area of the screen, while setting a higher width makes the speech scale horizontally instead of vertically.

PR #1447, opened 2021-11-17
Unsets female flag for character creation when male is selected

kamronbatman:
### Summary
At character creation if you select female, then the flag is set to female, but never unset when male is selected after that.

### Test
At the character creation screen, select female, then select male. Create the character. it will actually be female.


andreakarasho:
point to dev please :)

PR #1454, opened 2021-12-01
NameOverHeadHandlerGump now saves the last screen position when client is restarted

Tranq1:


src/Configuration/Profile.cs
```diff
@@ -344,6 +345,12 @@ private void SaveGumps(string path)
                     }
                 }
 
+                // Save NameOverHeadHandlerGump's last position even if it's not opened on save unless it's at default position
+                if (!gumps.Any(g => g.GumpType == GumpType.NameOverHeadHandler) && NameOverHeadHandlerGump.LastPosition.HasValue && NameOverHeadHandlerGump.LastPosition != new Point(100, 100))
+                {
+                    var nameOverHeadHandlerGump = new NameOverHeadHandlerGump();
+                    gumps.AddLast(nameOverHeadHandlerGump);
+                }
```
andreakarasho: I got the reason behind these lines of code, but i really do not like this workaroud.
I think we can spawn the gump one time per game session and disable/enable it when necessary,

Tranq1: Yeah that could work, so instead of spawning/despawning the gump every time the user presses ctrl+shift, we just create it once and disable it? I can look into changing the OverHeadGumpManager so it works like that

andreakarasho: Exactly, a single persistent gump that is visible when pressing cltr shift

Tranq1: Okay, I pushed another commit that is now only instantiating one gump and setting disabled and visible to true/false based on open/close. The initial instantiation still only happens when ctrl+shift is pressed for the first time to save some resources if a player never uses it.

src/Game/Managers/NameOverHeadManager.cs
```diff
@@ -95,22 +95,28 @@ public static bool IsAllowed(Entity serial)
...
+                _gump = new NameOverHeadHandlerGump();
             }
 
-            _gump = new NameOverHeadHandlerGump();
+            if (!_gump.IsEnabled)
+                _gump.IsEnabled = true;
+            if (!_gump.IsVisible)
+                _gump.IsVisible = true;
+
             UIManager.Add(_gump);
```
andreakarasho: the gump must be added when instanciated and not everytime the `Open` function is called.

src/Game/Managers/NameOverHeadManager.cs
```diff
@@ -95,22 +95,28 @@ public static bool IsAllowed(Entity serial)
...
         {
-            if (_gump != null)
+            if (_gump == null)
             {
-                return;
+                _gump = new NameOverHeadHandlerGump();
             }
 
-            _gump = new NameOverHeadHandlerGump();
+            if (!_gump.IsEnabled)
```
andreakarasho: These kind of checks are not necessary, just set the variable directly to `true` or `false`

src/Game/Managers/NameOverHeadManager.cs
```diff
@@ -95,22 +95,28 @@ public static bool IsAllowed(Entity serial)
...
+                _gump.IsEnabled = true;
+            if (!_gump.IsVisible)
```
andreakarasho: same ^

src/Game/Managers/NameOverHeadManager.cs
```diff
@@ -95,22 +95,28 @@ public static bool IsAllowed(Entity serial)
...
         {
-            if (_gump != null)
-            {
-                _gump.Dispose();
-                _gump = null;
-            }
+            if (_gump == null)
+                return;
+
+            if (_gump.IsEnabled)
```
andreakarasho: same ^

src/Game/Managers/NameOverHeadManager.cs
```diff
@@ -95,22 +95,28 @@ public static bool IsAllowed(Entity serial)
...
+                _gump.IsEnabled = false;
+            if (_gump.IsVisible)
```
andreakarasho: same ^

Tranq1:
Resolved all your comments hopefully, thanks for the review! 🙂

PR #1457, opened 2021-12-07
Add Markers Manager

anethus:
Add possibility to add markers on click for users

userMarkersResize

And after rework
markers


andreakarasho:
damn this file `src/Resources/ResGumps.resx` has conflicts :\

PR #1461, opened 2021-12-09
GPU-based circle of transparency

jaedan:


andreakarasho:
implemented in https://github.com/ClassicUO/ClassicUO/pull/1873

PR #1468, opened 2021-12-18
Lock Overhead Name Gumps Options

dreadlordchase:
Currently gumps will block players from moving, even if you can't close the gump. This can cause interference with movement when in area a with many mobiles/items and overhead names enabled.

This PR add the ability to "lock" overhead names and not interfere with player movement.

This adds a new control property, AllowPlayerWorldMovement, that can prevent controls from blocking the world movement of the player. If this property is true for a control, the game cursor icon will not change when hovering the gump and the gump will allow the game cursor to behave as if it's over the world.

The AllowPlayerWorldMovement property should be implemented is such a way that it doesn't impact any existing gumps and the current behavior of all gump should remain unchanged.

If there was already a way to achieve this without creating a new control property I was not able to find it. 
If you believe that this could also be achieved in another manner, please let me know so I can look into implementing that instead.


roxya:
This PR does two separate things - lock overhead gumps, and change the behaviour of the target cursor range indicator. I don't think it should be a single PR.

I also think it should be squashed to avoid unnecessary commits to launchsettings.json which ends up being untouched.

Not my decision if it gets merged, just noting my view

dreadlordchase:
tried to squash commits, didn't work out too well... going to just redo the changes in a new branch.

PR #1472, opened 2021-12-30
Add macros CloseCorpses and CloseInactiveHealthBars

anethus:
- CloseCorpses - close corpses gump (and gridLoot if enabled)
- CloseInactiveHealthBars - Close all inactive (out of range and isDead) healthbars

### DIFF
--- a/src/Game/Managers/MacroManager.cs
+++ b/src/Game/Managers/MacroManager.cs
@@ -2077,13 +2077,13 @@ namespace ClassicUO.Game.Managers
         UseItemInHand,
         UsePotion,
         CloseAllHealthBars,
-        CloseInactiveHealthBars,
-        CloseCorpses,
         RazorMacro,
         ToggleDrawRoofs,
         ToggleTreeStumps,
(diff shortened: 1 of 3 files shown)


Tranq1:
Wouldn't IsDead also close health bars from bonded pets that died even though they're still on screen? Tamers would probably want to keep those

anethus:
Yes, they will be disposed.
From perspective of tamers yes, but on the other hands, as you use your summoned ones, you want their bars to disappear.
I don't want to add multiple macros so i will do as you suggest

andreakarasho:
Macros are stored in XML and the `enum MacroType` is stored as an `int`. 
Adding more enum entries in the middle of the struct shifts next entries by N [where N = number of entries added to the enum].
Just an example:

```csharp
enum MacroType
{
    None ,
    One,
    Two,
    ADDED, // here will be 3, so MacroType.Three is equals to 4
    Three ,
    Four,
    Five
}
```

anethus:
done

PR #1486, opened 2022-01-17
Add option to lock nameplates.

dreadlordchase:
https://streamable.com/ha0988
(the cursor in the recording is slightly offset from where it actually is ingame. Ingame the cursor is over the nameplate to demoenstrate how it doesn't block movement and can be closed.)

Add option to lock nameplates.

Locked nameplates will not block player movement and can be closed by holding ALT (similar to anchored healthbars). They otherwise function the same as nameplates current do.
When a locked nameplate has been closed it will not reappear until the user has toggled nameplates.

To have a gump that will still accept input while not blocking player movement a new Control property was added; AllowPlayerWorldMovement. Also CanCloseWithRightClick needed to be made a virtual property.

Comments on a previous PR for the same feature can be found [here](https://github.com/ClassicUO/ClassicUO/pull/1468).

### DIFF
--- a/src/Configuration/Profile.cs
+++ b/src/Configuration/Profile.cs
@@ -170,6 +170,7 @@ namespace ClassicUO.Configuration
         public bool HoldDownKeyAltToCloseAnchored { get; set; } = true;
         public bool CloseAllAnchoredGumpsInGroupWithRightClick { get; set; } = false;
         public bool HoldAltToMoveGumps { get; set; }
+        public bool LockOverheadNameGumps { get; set; }
 
         public bool HideScreenshotStoredInMessage { get; set; }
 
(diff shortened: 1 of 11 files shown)


andreakarasho:
I wonder if it's better to avoid more options to setup by the end user and just hardcode this feature for the nameoverhead gump

Tranq1:
Yeah I'd be fine if it worked like that by default, but I'm sure some people would complain because they liked it better before :/

andreakarasho:
> Yeah I'd be fine if it worked like that by default, but I'm sure some people would complain because they liked it better before :/

You right, i forget to add an important part:
adding a check to the Mouse.RClickOffset [i don't remember the name exaclty but you get it] like:

``` when right is pressed
   if  Mouse.RClickOffset > (4, 4)  ---> move the player   ELSE   close the gump
```

dreadlordchase:
Ya I added this ask an options so since it is a pretty large change from how they currently work and didn't want to force it upon people.

I'm not sure what you mean by with the right click offset though.

PR #1495, opened 2022-02-07
Show an 8x8 grid on world map if zoomed in far enough

rsuperior:
The player can disable or re-enable this feature at will.

---
Full disclosure: I'm not actually sure whether all shard operators would want this.  Whether any shards are doing 8x8 skill gain or whatever significance 8x8 tiles have for resource spawning, I only know that some players on some shards find similar features in no-longer-maintained assistant apps to be useful for them.

I think it's really just a Quality of Life improvement, and I don't think it enables any player behavior that players can't achieve with other tools.  That said, I am not a shard operator, so I understand if there would be a different view.

Recording 2022-02-07 at 00 48 57


markdwags:
This is a standard feature on UOAM and makes sense to have it in the World Map, especially for shards that still use the RunUO code where mining or fishing nodes are in 8x8 squares.

PR #1497, opened 2022-02-09
Control light colours and light assignments with text files

roxya:
This PR is to allow shards to easily create new light colours and assign lights to items, which is currently all hardcoded.

- Allow custom RGB lights to be created by text file
- Allow RGB lights to be assigned to item IDs by text file
- Allow hue lights to be assigned to item IDs by text file
- Clean up of the code definition for RGB lights, so even hard coding new lights in a fork is easier

RGB lights are defined by an RGB colour and a set of curves, same as the original lights. https://roxya.github.io/CUOLightShaderTester/ shows how they work.

RGB light definition in `lightshaders.txt`:
```
# FORMAT
# ID RGB R_CURVE G_CURVE B_CURVE
# HARD LIMIT IS 63

1 00FF00 Standard A Standard
2 7F7FFF Standard Standard Standard
6 FF00FF B Standard A
10 3F3FFF Standard Standard Standard
```

The item assignment in `lights.txt`:

```
# FORMAT
# ITEM_ID LIGHT_SHADER_ID

0xE31 H2731       <-- Hue 2731
0xEFA 35          <--- RGB light 35
```

Example of a hue light:
huelight

Example of a custom RGB light `BDFFD6 A A E`
image

### DIFF
--- a/src/Game/Data/LightColors.cs
+++ b/src/Game/Data/LightColors.cs
@@ -30,21 +30,30 @@
 
 #endregion
 
+using ClassicUO.Utility;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
 namespace ClassicUO.Game.Data
 {
     internal static class LightColors
     {
-        private static readonly uint[][] _lightCurveTables = new uint[5][]
-        {
-            new uint[32] { 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,2,3,4,6,8,10,12,14,16,18,20,22,24,26,28},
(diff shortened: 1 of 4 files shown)


andreakarasho:
I see you create the default txt so nobody will notice the difference in game.
As you said it's more a shard-specific stuff.
From my side it's a good idea!

markdwags:
Agreed, this is a great addition and allows shards to add customization without needing to maintain their own version. Cool stuff!

roxya:
I changed it so the default shaders defined in code are always loaded first, rather than writing to text and loading back in again. Avoids the situation of telling everyone to delete the text file if we ever need to add another default light colour (imagine!)

Now lightshaders.txt is only for new colours / overrides.

roxya:
roxya slow. I fixed some things and I think this is ready now

PR #1499, opened 2022-02-14
Cleanup drag & drop UI

andreakarasho:
Drag & drop had an odd and tricky beahviour.
Under the hood the game allowed controls that cannot move [`CanMove = false;`] to invoke `DragBegin` & `DragEnd` methods into the `MouseOver` function.

The `DragBegin` & `DragEnd` methods are useful for moving controls and not for static ones. So they must be not called when the main gump is set to no move.

Although some control needs to get invoked their `DragBegin` and `DragEnd` methods like the `NameOverheadGump`.
In these particular cases I suggest to store the last left mouse position and trigger the `DragBegin` manually.

### DIFF
--- a/src/Game/Managers/UIManager.cs
+++ b/src/Game/Managers/UIManager.cs
@@ -650,12 +650,10 @@ namespace ClassicUO.Game.Managers
                     DraggingControl = dragTarget;
                     _dragOrigin = Mouse.LClickPosition;
 
-                    /* we need to keep the mouse down controls to invoke the mouseup
                      for (int i = 0; i < (int) MouseButtonType.Size; i++)
                      {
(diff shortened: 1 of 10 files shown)


roxya:
Dragging spell icons from spellbooks is broken

andreakarasho:
> Dragging spell icons from spellbooks is broken

Fixed:
spellbook, abilities and racial book.
They share the same `DragBegin` behaviour now

roxya:
Journal/etc scroll resizing isn't smooth, gets stuck often

andreakarasho:
> Journal/etc scroll resizing isn't smooth, gets stuck often

done

roxya:
I haven't found any further issues

PR #1504, opened 2022-03-10
Animation Loader cleanups

jaedan:
This is a series of code clean ups that doesn't (or shouldn't) change any behavior. This builds toward two goals in the long term, although the main ClassicUO project may never get to either of these:

1) Support compressed texture formats. In particular, BC1 is a great match for UO art and is very, very small. But BC1, like the native UO art, does not support alpha.
2) Support loading assets, such as animations, from alternate file formats

This PR can be reviewed as a set of individual patches, patch by patch. Each patch along the way is fully functional.

### DIFF
--- a/src/IO/Resources/AnimationsLoader.cs
+++ b/src/IO/Resources/AnimationsLoader.cs
@@ -1385,6 +1385,8 @@ namespace ClassicUO.IO.Resources
             return true;
         }
 
+        [ThreadStatic] private static byte[] _decompressedData = null;
+
         private bool ReadUOPAnimationFrame(ushort animID, byte animGroup, byte direction, ref AnimationDirection animDirection)
         {
             AnimationGroupUop animData = DataIndex[animID].GetUopGroup(ref animGroup);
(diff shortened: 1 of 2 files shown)


andreakarasho:
I just logged with latest client protocol 7.0.90+ and cuo crashed. 
I can't get the reason because there are too much changes. 
Then i highly suggest to maintain the UOP reading logic, because UO tends to use frame skipping as a "sentinel" to cut frames.
Another issue i noticed is the alpha that seems super additive.

jaedan:
I've reduced this to the first two commits. We can work in smaller increments.

src/IO/Resources/AnimationsLoader.cs
```diff
@@ -1385,6 +1385,8 @@ public bool LoadAnimationFrames(ushort animID, byte animGroup, byte direction, r
             return true;
         }
 
+        [ThreadStatic] private static byte[] _decompressedData = new byte[0x1000];
```
andreakarasho: `[ThreadStatic]` causes to create 2 instances of `_decompressedData`:
1 at loading time due to the `Task.Run`
1 from the main thread

src/IO/Resources/ArtLoader.cs
```diff
@@ -139,102 +153,126 @@ private void AddSpriteToAtlas(TextureAtlas atlas, int g, bool isTerrain)
...
+                        );
                     }
                 }
             }
+
+            return true;
         }
-      
+
+        [ThreadStatic] private static uint[] _data = new uint[0x10000];
```
andreakarasho: `[ThreadStatic]` causes to create 2 instances of `_data`:
1 at loading time due to the `Task.Run`
1 from the main thread

benlwalker: Yes - that's by design so that both threads can load sprites. I can make it start off as null though so if a thread never attempts to read anything this doesn't get allocated.

jaedan:
I've made it so _data starts off as null and only gets allocated if a thread tries to read textures at least once.

PR #1507, opened 2022-03-12
Simplify UOP animation reading

jaedan:
Use a nested loop first over direction, then over the frames, to simplify some of the logic. Also, combine the two separate loops into one to avoid repeated error checking and skipped frame logic.

### DIFF
--- a/src/IO/Resources/AnimationsLoader.cs
+++ b/src/IO/Resources/AnimationsLoader.cs
@@ -1438,77 +1438,68 @@ namespace ClassicUO.IO.Resources
             animDirection.FrameCount = (byte)(type < ANIMATION_GROUPS_TYPE.EQUIPMENT ? Math.Round(fc / 5f) : 10);
             animDirection.SpriteInfos = new SpriteInfo[animDirection.FrameCount];
 
-            int headerSize = sizeof(UOPAnimationHeader);
-            int count = 0;
-
+            /* If the UOP files didn't omit frames, we could just do this:
+             * reader.Skip(sizeof(UOPAnimationHeader) * direction * animDirection.FrameCount);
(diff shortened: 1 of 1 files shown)


src/IO/Resources/AnimationsLoader.cs
```diff
@@ -1438,77 +1438,62 @@ private bool ReadUOPAnimationFrame(ushort animID, byte animGroup, byte direction
             animDirection.FrameCount = (byte)(type < ANIMATION_GROUPS_TYPE.EQUIPMENT ? Math.Round(fc / 5f) : 10);
             animDirection.SpriteInfos = new SpriteInfo[animDirection.FrameCount];
 
-            int headerSize = sizeof(UOPAnimationHeader);
-            int count = 0;
-
+            /* If the UOP packers weren't a bunch of idiots that randomly omit frames, we could
```
andreakarasho: 🤣

src/IO/Resources/AnimationsLoader.cs
```diff
@@ -1438,77 +1438,62 @@ private bool ReadUOPAnimationFrame(ushort animID, byte animGroup, byte direction
...
+                for (ushort frameNum = 0; frameNum < animDirection.FrameCount; frameNum++)
                 {
-                    if (currentDir != direction)
+                    long start = reader.Position;
+                    animHeaderInfo = (UOPAnimationHeader*)reader.PositionAddress;
+
+                    if (animHeaderInfo->Group != animGroup || start + animHeaderInfo->DataOffset >= reader.Length)
                     {
-                        ++currentDir;
+                        throw new Exception("Corrupt UOP file!");
```
andreakarasho: UOP are imprevedible. I know. 
Instead of ruining the user experience just for an animation with an exception, we can go ahead i guess

PR #1508, opened 2022-03-13
Animation Loading Refactoring

jaedan:
Begin to separate the function that loads in the raw animation data from the mul/uop files from the function that constructs a graphics texture in the texture atlas.

### DIFF
--- a/src/IO/Resources/AnimationsLoader.cs
+++ b/src/IO/Resources/AnimationsLoader.cs
@@ -1363,6 +1363,9 @@ namespace ClassicUO.IO.Resources
                 return false;
             }
 
+            Span<FrameInfo> frames;
+            int uopFlag = 0;
+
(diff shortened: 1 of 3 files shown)


src/IO/Resources/AnimationsLoader.cs
```diff
@@ -1372,35 +1334,77 @@ public bool LoadAnimationFrames(ushort animID, byte animGroup, byte direction, r
...
+        {
+            public int Num;
+            public short CenterX;
+            public short CenterY;
+            public short Width;
+            public short Height;
+            public uint[] Pixels;
+        }
+
+        [ThreadStatic] private static FrameInfo[] _frames = new FrameInfo[20];
```
andreakarasho: `[ThreadStatic]` causes an extra allocation during the data loading

PR #1515, opened 2022-03-25
Added: packet handler for the "close vendor buy list" packet.

mythril:
This should implement the packet I described in the feature request ( https://github.com/ClassicUO/ClassicUO/issues/1509 ). I ran the change locally and it seemed to work as I expected.

I'm not super familiar with C# (about 150 hours of experience with it). If this code isn't up to your standards I will gladly accept input / take direction on how I should present this to you.


src/Network/PacketHandlers.cs
```diff
@@ -1953,6 +1954,18 @@ private static void UpdateContainedItems(ref StackDataReader p)
             }
         }
 
+        public static void CloseVendorInterface(ref StackDataReader p)
```
andreakarasho: `public` to `private`

PR #1523, opened 2022-04-07
AnimationLoader simplifications

jaedan:
Quick simplifications.

### DIFF
--- a/src/IO/Resources/AnimationsLoader.cs
+++ b/src/IO/Resources/AnimationsLoader.cs
@@ -509,7 +509,7 @@ namespace ClassicUO.IO.Resources
                                     DataIndex[index].MountedHeightOffset = mountedHeightOffset;
                                 }
 
-                                DataIndex[index].GraphicConversion = (ushort)(realAnimID | 0x8000);
+                                DataIndex[index].HasBodyConversion = false;
                                 DataIndex[index].FileIndex = (byte)animFile;
 
                                 addressOffset += currentIdxFile.StartAddress.ToInt64();
@@ -1041,9 +1041,9 @@ namespace ClassicUO.IO.Resources
 
                 if (replace)
                 {
-                    if (!DataIndex[i].HasBodyConversion)
+                    if (DataIndex[i].BodyConvGroups != null)
                     {
-                        DataIndex[i].GraphicConversion = (ushort) (DataIndex[i].GraphicConversion & ~0x8000);
+                        DataIndex[i].HasBodyConversion = true;
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/IO/Resources/AnimationsLoader.cs
```diff
@@ -886,7 +886,7 @@ public void ConvertBodyIfNeeded(ref ushort graphic, bool isParent = false, bool
 
                 do
                 {
-                    if ((DataIndex[newGraphic].HasBodyConversion || !dataIndex.HasBodyConversion) && !(DataIndex[newGraphic].HasBodyConversion && dataIndex.HasBodyConversion))
```
andreakarasho, line 889: i think this will cause issues because iirc we need to be sure that `new` and `old` graphics don't have a conversion.
So the conversion will happen when:
new graphic index has a conversion OR the old one doesn't
AND
both doesn't have a conversion.

src/IO/Resources/AnimationsLoader.cs
```diff
@@ -944,7 +944,7 @@ public void ReplaceAnimationValues(ref ushort graphic, ref byte action, ref usho
 
                 do
                 {
-                    if ((DataIndex[newGraphic].HasBodyConversion || !index.HasBodyConversion) && !(DataIndex[newGraphic].HasBodyConversion && index.HasBodyConversion))
```
andreakarasho, line 947: same as above

src/IO/Resources/AnimationsLoader.cs
```diff
@@ -1937,7 +1934,7 @@ internal class IndexAnimation
         private byte[] _uopReplaceGroupIndex;
         public bool IsUOP => (Flags & ANIMATION_FLAGS.AF_USE_UOP_ANIMATION) != 0;
 
-        public bool HasBodyConversion => (GraphicConversion & 0x8000) == 0 && BodyConvGroups != null;
```
andreakarasho, line 1940: Does the code check for `BodyConvGroups != null` when call it?

src/IO/Resources/AnimationsLoader.cs
jaedan: Ok so if we define
A = DataIndex[newGraphic].HasBodyConversion
B = dataIndex.HasBodyConversion

Then this statement is
(A || !B) && !(A &&B)

If I then build a table of inputs (A, B):

(T, T) -> F
(T, F) -> T
(F, T) -> F
(F, F) -> T

So it looks like it's actually equivalent to !B. In other words, the condition is

if (! dataIndex.HasBodyConversion)

Does that look correct to you? I'll update you that if so.

andreakarasho: Im from my mobilephone so it's a little bit hard to review.
Yeah right, but in the commit i see:
`if (DataIndex[newGraphic].HasBodyConversion)` which should be wrong

jaedan: Yes I have it wrong in the patch. I'll update.

src/IO/Resources/AnimationsLoader.cs
jaedan: I added a check that should be equivalent to this on line 1044

jaedan:
Ok I think I have this straightened out now

PR #1524, opened 2022-04-10
Customizable name plate/name overhead options

Tranq1:
This PR is meant to give players the ability to fine tune how they use the name overhead stuff by giving them more detailed options.

**THIS PR IS STILL A DRAFT because implementation is not finished and it's not fully tested, but please feel free to give feedback regardless!**

### Main Features

* Fine granular options as to which items/mobiles to display name plates for (✅ implemented)
* Detailed options page to configure those options (✅ implemented)
* Ability to switch through configurations by assigning each one a hotkey (✅ implemented)

## Planned additional features

* Switch between configs by pressing shift+ctrl+arrow up/down
* Add button on the overhead handler gump to allow it to stay up (similar to the macro toggle)
* Add option to enable health bar below name plate
* Add option to have background fill act as health bar
* Add option to show health % in the name plate
* Add option to allow players to customize min/max name plate width

### Options

The available options are currently as follows:

image

### How will this change affect players (if they don't like it?)

When the PR is accepted, a default config will be created that mirrors the current functionality and options exactly. This way, players who don't like the new options are not forced to adapt and can simply use the name overhead functionality as they always have.

### Demo

overhead_options_settings_demo

### DIFF
--- a/src/Game/Managers/NameOverHeadManager.cs
+++ b/src/Game/Managers/NameOverHeadManager.cs
@@ -33,9 +33,11 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Xml;
 using ClassicUO.Configuration;
+using ClassicUO.Game.Data;
 using ClassicUO.Game.GameObjects;
 using ClassicUO.Game.UI.Gumps;
 using ClassicUO.Utility.Logging;
@@ -53,6 +55,38 @@ namespace ClassicUO.Game.Managers
         MobilesCorpses = Mobiles | Corpses
     }
 
+
+    [Flags]
+    internal enum NameOverheadOptions
+    {
+        None = 0,
+
+        // Items
+        Containers = 1 << 0,
+        Gold = 1 << 1,
+        Stackable = 1 << 2,
+        Other = 1 << 3,
+
+        // Corpses
+        MonsterCorpses = 1 << 4,
+        HumanoidCorpses = 1 << 5,
+        OwnCorpses = 1 << 6,
+
+        // Mobiles (type)
+        Humanoid = 1 << 7,
+        Monster = 1 << 8,
+        OwnFollowers = 1 << 9,
+
+        // Mobiles (notoriety)
+        Innocent = 1 << 10,
+        Ally = 1 << 11,
+        Gray = 1 << 12,
+        Criminal = 1 << 13,
+        Enemy = 1 << 14,
+        Murderer = 1 << 15,
+        Invulnerable = 1 << 16,
+    }
+
     internal static class NameOverHeadManager
     {
         private static NameOverHeadHandlerGump _gump;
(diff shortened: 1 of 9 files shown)


src/Game/Managers/NameOverHeadManager.cs
```diff
@@ -116,7 +226,230 @@ public static void Close()
...
+
+                Options.Clear();
+                CreateDefaultEntries();
+                Save();
+
+                return;
+            }
+
+            Options.Clear();
+            XmlDocument doc = new();
```
andreakarasho, line 258: `new XmlDocument()` would be better to keep in line with the old net472

src/Game/Managers/NameOverHeadManager.cs
```diff
@@ -116,7 +226,230 @@ public static void Close()
...
+            }
+        }
+
+        public static void Save()
+        {
+            var list = Options;
+
+            string path = Path.Combine(ProfileManager.ProfilePath, "nameoverhead.xml");
+
+            using XmlTextWriter xml = new(path, Encoding.UTF8)
```
andreakarasho, line 291: No `new()` for net472

src/Game/Managers/NameOverHeadManager.cs
```diff
@@ -31,65 +31,175 @@
...
         {
             get => ProfileManager.CurrentProfile.NameOverheadToggled;
-            set => ProfileManager.CurrentProfile.NameOverheadToggled = value;
+            private set => ProfileManager.CurrentProfile.NameOverheadToggled = value;
         }
 
+        public static bool IsTemporarilyShowing { get; private set; }
+        public static bool IsShowing => IsPermaToggled || IsTemporarilyShowing || Keyboard.Ctrl && Keyboard.Shift;
+
+        private static List<NameOverheadOption> Options { get; set; } = new();
```
andreakarasho, line 107: No `new()` for net472

src/Game/UI/Controls/NameOverheadAssignControl.cs
```diff
@@ -0,0 +1,280 @@
...
+using ClassicUO.IO.Resources;
+using ClassicUO.Resources;
+using SDL2;
+
+namespace ClassicUO.Game.UI.Controls
+{
+    internal class NameOverheadAssignControl : Control
+    {
+        private readonly HotkeyBox _hotkeyBox;
+        private readonly Dictionary<NameOverheadOptions, Checkbox> checkboxDict = new();
```
andreakarasho, line 47: No `new()` for net472

src/Game/UI/Gumps/NameOverHeadHandlerGump.cs
```diff
@@ -44,6 +45,9 @@ internal class NameOverHeadHandlerGump : Gump
 
         public override GumpType GumpType => GumpType.NameOverHeadHandler;
 
+        private readonly List<RadioButton> _overheadButtons = new();
```
andreakarasho, line 48: No `new()` for net472

roxya:
I think there are too many options. Not sure why there should be special settings for gold, stackables, or different types of corpses. I'd say "own corpse" is going too far, if it can even be done without a hack.

I'm not so sure about having a bunch of game scene input related stuff in the name overhead classes. I don't think we've used this pattern of notifying other objects about keypresses to have their own handling. Would feel better to assign slots and have them in the existing macro system.

About the planned features, I would say hold off on them because the arrows thing doesn't fit with the currently hardcoded arrow key movement, and for the health stuff we already have healthbars AND HP lines AND HP% texts (and quite frankly too many implementations of healthbar handling which could use a good refactor).

The one I would think about is "Add button on the overhead handler gump to allow it to stay up (similar to the macro toggle)". I assume you mean if you press ctrl+shift, that you get the option to keep the names showing when you let go? It will look strange if it was enabled from the macro but it sounds like an idea.

This is all just opinion, at the end of the day it's Karasho who makes the decisions.

2dchaos (approved):

PR #1527, opened 2022-04-22
Delay loading of bodyconv.def until expansion flags are known

jaedan:
By delaying this, we can eliminate BodyConvGroups entirely.


andreakarasho:
I think there are cases where the server doesn't send the animation packet

jaedan:
Ok, I added a check that if we haven't received any flags by the time it goes to log into the world, it parses all the files as if the flags are 0.

PR #1528, opened 2022-04-24
Use ISO-8859-1 instead of ASCII

brndd:
Since proper CP-1252 support is still pending, this is probably a good compromise. I don't know why I didn't think of this sooner. ISO-8859-1 is only missing 27 or so characters that CP-1252 has, which is a big improvement compared to the ~96 characters ASCII is missing.

### DIFF
--- a/src/Utility/StringHelper.cs
+++ b/src/Utility/StringHelper.cs
@@ -52,7 +52,7 @@ namespace ClassicUO.Utility
                 {
                     //Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
                     //_cp1252Encoding = Encoding.GetEncoding(1252);
-                    _cp1252Encoding = Encoding.ASCII;
+                    _cp1252Encoding = Encoding.GetEncoding("iso-8859-1");
                 }
                 return _cp1252Encoding;
             }


andreakarasho:
Is not better to go with `CP-1252`?

brndd:
CP-1252 is not supported in .NET 5 and above without using the [System.Text.Encoding.CodePages](https://www.nuget.org/packages/System.Text.Encoding.CodePages/) NuGet package. We could not use that because of some dependency hell nonsense; including it would break builds on Windows, and fixing the build on Windows would break builds on Linux and Mac.

andreakarasho:
You removed the nuget package and used a custom cp1252 implementation iirc

brndd:
I tried a custom implementation, yeah, but the code is very smelly and I didn't have time to make it less smelly back then, so it's still not merged.

I don't really have time to improve it right now either, so this one-line change PR would be a good stop-gap solution until CP-1252 is properly supported. It's certainly better than ASCII because it doesn't mangle special characters like é, ó, ä, ö, etc.

andreakarasho:
fixed in https://github.com/ClassicUO/ClassicUO/pull/1439

PR #1529, opened 2022-04-26
Fixes sparse buffer filling by using an exclusive array pool

kamronbatman:
In future versions of .NET (Core, 5/6/7+) the C# core uses `ArrayPool<T>.Shared` without clearing buffers on return. _It is not recommended that ArrayPool.Shared is used with buffers that are sparsely filled since they might be dirty when rented_.

To get around this, I created another ArrayPool that is exclusive for our use and will always be cleared on rent for .NET Core.

### DIFF
--- a/src/Utility/Collections/CleanArrayPool.cs
+++ b/src/Utility/Collections/CleanArrayPool.cs
@@ -1,0 +1,19 @@
+using System.Buffers;
+
+namespace ClassicUO.Utility.Collections
+{
+    public class CleanArrayPool<T>
+    {
+        private static CleanArrayPool<T> _sharedArrayPool = new CleanArrayPool<T>();
+        public static CleanArrayPool<T> Shared => _sharedArrayPool;
+
+        private ArrayPool<T> _pool = ArrayPool<T>.Create(0x100000, 8);
+
+        public T[] Rent(int minimumLength) => _pool.Rent(minimumLength);
+
+        public void Return(T[] array)
+        {
+            _pool.Return(array, true);
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


kamronbatman:
There are other places where we need an exclusive array pool. We can combine them. I'll update this PR accordingly before we merge it.

src/Utility/Collections/CleanArrayPool.cs
```diff
@@ -0,0 +1,19 @@
+using System.Buffers;
+
+namespace ClassicUO.Utility.Collections
+{
+    public class CleanArrayPool<T>
+    {
+        private static CleanArrayPool<T> _sharedArrayPool = new CleanArrayPool<T>();
+        public static CleanArrayPool<T> Shared => _sharedArrayPool;
+
+        private ArrayPool<T> _pool = ArrayPool<T>.Create(0x100000, 8);
```
kamronbatman, line 10: These are default settings for ArrayPool.

src/Utility/Collections/CleanArrayPool.cs
```diff
@@ -0,0 +1,19 @@
+using System.Buffers;
+
+namespace ClassicUO.Utility.Collections
+{
+    public class CleanArrayPool<T>
```
kamronbatman, line 5: If desired, we could implement `ArrayPool<T>` abstract class and ignore the `clear` parameter in `Return`. Let me know if you prefer that.

src/Utility/Collections/CleanArrayPool.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            var array = _pool.Rent(minimumLength);
+#if NETCOREAPP3_1_OR_GREATER
+            Array.Clear(array, 0, minimumLength);
+#endif
+
+            return array;
+        }
+
+        public void Return(T[] array) => _pool.Return(array);
```
kamronbatman, line 22: TODO: Add a check for non-coreapp to clear on return.

PR #1531, opened 2022-04-27
BodyConv removal

jaedan:
By delaying this, we can eliminate BodyConvGroups entirely.

Updated based on our conversations on discord earlier.

This needs a lot more testing so don't merge it yet.

### DIFF
--- a/src/IO/Resources/AnimationsLoader.cs
+++ b/src/IO/Resources/AnimationsLoader.cs
@@ -90,6 +90,65 @@ namespace ClassicUO.IO.Resources
             new List<Tuple<ushort, byte>>(), new List<Tuple<ushort, byte>>()
         };
 
+        /* Load from the given file at the given index into the slot at body */
+        private void LoadMulBody(ushort body, ushort index, ANIMATION_GROUPS_TYPE type, int fileNum)
+        {
+            UOFile idxFile = _files[fileNum].IdxFile;
+            long offsetToData = DataIndex[body].CalculateOffset(index, type, out int count);
+            long maxaddress = idxFile.StartAddress.ToInt64() + idxFile.Length;
+
+            if (offsetToData >= idxFile.Length)
+            {
+                return;
+            }
+
+            DataIndex[body].Type = type;
+
+            bool isValid = false;
+            long address = idxFile.StartAddress.ToInt64() + offsetToData;
+
+            DataIndex[body].Groups = new AnimationGroup[100];
+
+            int offset = 0;
+
+            for (byte j = 0; j < 100; j++)
+            {
+                DataIndex[body].Groups[j] = new AnimationGroup
+                {
+                    Direction = new AnimationDirection[5]
+                };
+
+                if (j >= count)
+                {
+                    continue;
+                }
+
+                for (byte d = 0; d < 5; d++)
+                {
+                    if (DataIndex[body].Groups[j].Direction[d] == null)
+                    {
+                        DataIndex[body].Groups[j].Direction[d] = new AnimationDirection();
+                    }
+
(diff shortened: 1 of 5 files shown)


src/IO/Resources/AnimationsLoader.cs
```diff
@@ -90,6 +90,61 @@ public void CreateAtlas(GraphicsDevice device)
...
+                    AnimIdxBlock* aidx = (AnimIdxBlock*)(address + offset * sizeof(AnimIdxBlock));
+                    ++offset;
+
+                    if ((long)aidx < maxaddress && aidx->Size != 0 && aidx->Position != 0xFFFFFFFF && aidx->Size != 0xFFFFFFFF)
+                    {
+                        DataIndex[body].Groups[j].Direction[d].Address = aidx->Position;
+                        DataIndex[body].Groups[j].Direction[d].Size = aidx->Size;
+
+                        isValid = true;
+                    }
```
andreakarasho: for `bodyconv.def` the code must be this:

```csharp
if ((long)aidx < maxaddress && aidx->Position != 0xFFFFFFFF && aidx->Size != 0xFFFFFFFF)
{
    ref var direction = ref DataIndex[index].Groups[j].Direction[d];
  
    direction.Address = aidx->Position;
    direction.Size = Math.Max(1, aidx->Size);
  
    isValid = true;
}                                     
```

In particular the `aidx->Size != 0` must be removed. Time ago i found an animation with size == 0 but with 1 valid frame. But this is not true when reading `anim.mul`, which wants your implementation.

andreakarasho: just found the commit:
https://github.com/ClassicUO/ClassicUO/commit/4eb3f9f6e6e9e67765ab28e1e7adfa9bdee551da

Probably we can apply this logic when fileindex > 0

jaedan: Ok. I didn't mean to change this logic so I'll fix it up.

src/IO/Resources/AnimationsLoader.cs
```diff
@@ -982,42 +904,27 @@ public bool PixelCheck(ushort animID, byte group, byte direction, bool uop, int
...
-                if (replace)
-                {
-                    if (DataIndex[i].BodyConvGroups != null)
-                    {
-                        DataIndex[i].HasBodyConversion = true;
-                    }
-                }
+                ProcessBodyConvDef((LockedFeatureFlags)flags);
+                ProcessBodyDef();
+                ProcessCorpseDef();
```
andreakarasho, line 924: Not sure if it's volutary, but from what we said on discord `Body.def` & `Corpse.def` should be loaded before `Bodyconv.def` [?]

jaedan: I thought so initially but that's not the case. You ran a test with the following body.def:

```
11 {28} 1401
```

And bodyconv.def
```
28    206  -1    -1    -1    -1
```

Then, drawing body 11 draws a spider (anim.mul+28) not a swamp dragon (anim2.mul+206). That must mean the body.def overwrites the bodyconv.def, so it is loaded second.

jaedan: I think the main thing we were all missing is that there's a "logical" set of body values and then physical offsets into files. Body.def first column is logical body value and the number inside {} is the physical offset into anim.mul, not a logical body value. So there is never any sort of "chaining" going on in these files and hence why my patch removes the loops.

kamronbatman: Which body is used also depends on feature flags (for the EA client at least).

andreakarasho:
is this a thing?

jaedan:
This isn't up to date.

PR #1534, opened 2022-05-06
Implement OSI UseObject macro

roxya:
Marking as a draft, maybe it's done but I'll check again on the weekend.

OSI's UseObject internally checks the cliloc of the item's name to find the correct items, I suppose because using the graphic wasn't feasible for some of them (e.g. with the potions it tries to use the best type available, and also the two petal items are the same graphic). I've confirmed the exact behaviour.

To this end, I've stored the cliloc number on the items. For stacked items the cliloc comes in as an argument of another cliloc (e.g. 2#1041317), so I've had to account for this.

It all seems a bit hacky and I don't really like it but it's what OSI does.

### DIFF
--- a/src/Game/GameObjects/Entity.cs
+++ b/src/Game/GameObjects/Entity.cs
@@ -70,6 +70,7 @@ namespace ClassicUO.Game.GameObjects
         public bool IsClicked;
         public uint LastStepTime;
         public string Name;
+        public int NameCliloc;
         public uint Serial;
         public HitsRequestStatus HitsRequest;
 
--- a/src/Game/GameObjects/Entity.cs
+++ b/src/Game/GameObjects/Entity.cs
@@ -70,7 +70,6 @@ namespace ClassicUO.Game.GameObjects
         public bool IsClicked;
         public uint LastStepTime;
         public string Name;
-        public int NameCliloc;
         public uint Serial;
         public HitsRequestStatus HitsRequest;
 
--- a/src/Game/GameObjects/PlayerMobile.cs
+++ b/src/Game/GameObjects/PlayerMobile.cs
@@ -157,6 +157,18 @@ namespace ClassicUO.Game.GameObjects
             return null;
         }
 
+        public Item FindItemByCliloc(int cliloc)
+        {
+            Item backpack = FindItemByLayer(Layer.Backpack);
+
+            if (backpack != null)
+            {
+                return FindItemByClilocInContainerRecursive(backpack, cliloc);
+            }
+
+            return null;
+        }
+
         private Item FindItemInContainerRecursive(Item container, ushort graphic)
         {
             Item found = null;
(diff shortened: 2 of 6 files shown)


src/Game/GameObjects/PlayerMobile.cs
```diff
@@ -187,6 +199,53 @@ private Item FindItemInContainerRecursive(Item container, ushort graphic)
...
+                            return found;
+                        }
+                    }
+                }
+            }
+
+            return found;
+        }
+
+        public Item FindPreferredItemByCliloc(params int[] clilocs)
```
andreakarasho: `params int[]` creates an array internally every time you call the function.
So I sugest to replace it with `Span<int>`.
The function will be called using `Span<int> span = stackalloc int[] { 123, 456, 789 }`

src/Game/GameObjects/Entity.cs
```diff
@@ -70,6 +70,7 @@ protected Entity(uint serial)
         public bool IsClicked;
         public uint LastStepTime;
         public string Name;
+        public int NameCliloc;
```
andreakarasho: missing the cleanup when pull an entity from the Item/mobile pools

roxya: 👍 should be OK now, only for items since I'm not setting it for mobiles (amended original commit)

src/Game/GameObjects/PlayerMobile.cs
roxya: Memory management isn't something I've dealt with so thanks for the suggestion. I just implemented this in the new commit https://github.com/ClassicUO/ClassicUO/pull/1534/commits/036166676ff573b7b743ea463bf255a457a41382 but I'm not sure about it.

I can't do `Span<int> clilocs = stackalloc int[] { 1, 2, 3 };` for each macro since `clilocs` is already defined after the first time. Using `clilocs1`, and then `clilocs2`, feels bad but I guess it would work.

I tried this pattern
```csharp
case MacroType.UseObject:
    Item obj;
    Span<int> clilocs;

    switch (macro.SubCode)
    {
        case MacroSubType.BestHealPotion:
            clilocs = stackalloc int[3];
```
Which gives me "A result of a stackalloc expression of type 'Span<int>' cannot be used in this context because it may be exposed outside of the containing method"

So I settled on what I have, hope it's OK.

andreakarasho: It’s fine allocating fixed local array.
Or you can also try to put the body of eaxh case into a new scope using: 

{ 
// code
}

roxya: Yeah I was overthinking it with the local definitions, it's nicer now. All ready I think, commits squashed

src/Network/PacketHandlers.cs
```diff
@@ -4849,6 +4858,7 @@ private static void MegaCliloc(ref StackDataReader p)
                         if (entity != null && !SerialHelper.IsMobile(serial))
                         {
                             entity.Name = str;
+                            entity.NameCliloc = s.Item3 > 0 ? s.Item3 : s.Item1;
```
andreakarasho: Just back in mind about the OPL object.
It could contain the `NameCliloc` property instead of add it to each item.

src/Network/PacketHandlers.cs
```diff
@@ -4790,6 +4790,15 @@ private static void MegaCliloc(ref StackDataReader p)
                     continue;
                 }
 
+                int argcliloc = 0;
+
+                string[] argcheck = argument.Split('#');
```
andreakarasho: probably better to add a `SplitOptions.RemoveEmptyEntries` ?

src/Network/PacketHandlers.cs
roxya: Like in https://github.com/ClassicUO/ClassicUO/pull/1534/commits/ab13dd6e24eda6da9264fe680498497ae3e52b43?

src/Network/PacketHandlers.cs
roxya: Sure, done

src/Network/PacketHandlers.cs
andreakarasho: yeah!

PR #1536, opened 2022-05-26
Grid loot items order

GajuszZOrod:
The order in which items are shown in grid-loot will now depend on item
    type.
    Motivation: some items are likely to be always looted (e.g. gold, gems)
    and when looting is performed automatically (e.g. by Razor macros) it
    makes items to move in a grid making it harder to browse their
    properties. Hence, items like gold should be at the end of the grid.

### DIFF
--- a/src/Game/UI/Gumps/GridLootGump.cs
+++ b/src/Game/UI/Gumps/GridLootGump.cs
@@ -306,7 +306,8 @@ namespace ClassicUO.Game.UI.Gumps
         private bool ItemBelongsToGroup(Item it, int group)
         {
             // Note: items must be assigned to groups in a mutually-exclusive manner, so that each item occurs only once in the grid
-            if (it.IsMulti || it.Amount > 1)
+
+            if (it.ItemData.IsStackable)
                 return group > 0;
             else
                 return group == 0;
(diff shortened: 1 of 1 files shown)


roxya:
If you want something merged, you need to target the dev branch. Though I honestly don't think it's for CUO to prioritise different types of loot, nor make a decision based on an assumption about what an assistant plugin is doing.

GajuszZOrod:
> If you want something merged, you need to target the dev branch. Though I honestly don't think it's for CUO to prioritise different types of loot, nor make a decision based on an assumption about what an assistant plugin is doing.

Thanks for the advice about dev branch.

Since the current order of items in grid loot is kinda random, then I think introducing any change in ordering shouldn't really bother anyone...? Might help some people, without bothering others.

As for plugins usage, that was only one of the use case. Another is when multiple players are looting corpse at the same time: I think it is a fair assumption that stackable items (resources) are likely to be always picked up, while non-stackable items (like armour, weapon) which have custom properties are the ones people would like to take a look at before looting them (hence, we want them on the top of the list).

Anyways, I implemented it for myself and few friends, we are using that feature. I just thought I will share :)

PR #1541, opened 2022-07-29
Fix startup crashes

Sghirate:
- Crash when using client files < 3.0.5d: ClassicUO assumed gump.def and TexTerr.def to be there, otherwise the _spriteInfos arrays won't be correctly initialized. However ancient clients did not ship with those files => Fix: Make the parsing optional.
- Cash when World name is empty: Some Shards skip the server list and redirect the players to a default instance. The empty server name leads to a crash when generating the profile-save-path => Fix: fall back to a default world/server name: "_"


andreakarasho:
can you point to the `dev` branch?

andreakarasho:
I think you need to add your fix based on `dev` branch and not to the `master`. Otherwise github doesnt allow me to merge

Sghirate:
Changing to dev required merging in changes. While the Web-Atom still says 'Comitting...' it looks like that might already be through.

Sghirate:
Cherry picking into dev...

PR #1545, opened 2022-08-04
Fix Paperdoll & Prompt font on old clients

Sghirate:
The automatic font override of RenderedText would automatically switch the font to id 0 for ClientVersion < 3.0.5d, making it appear fat. Meaning: incorrect.
Switched those two gumps to explicitly use font 1.

Without fix:
image

With fix:
image


andreakarasho:
Thanks for all these PRs!

PR #1552, opened 2022-08-12
processes a Mounts.def file in the client directory (if it exists) and adjusts the internal CUO _mounts and animation table

credzba:
This is example modifies a bodyID of 1179 to use the 59 (red dragon) animation and the mount height of -30
Mounts.def  file looks like this:
#						
#  Fields:						
#  #bodyType	#animationId	#mountHeight
#
1179  59  -30	# Ridable dragon


credzba:
It didnt work well in a lot of cases, so never mind

PR #1566, opened 2022-10-21
Support mouse buttons/wheel in macros

roxya:
- Wheel/middle click
- Wheel Up/Down
- X1/X2

### DIFF
--- a/src/Game/Scenes/GameSceneInputHandler.cs
+++ b/src/Game/Scenes/GameSceneInputHandler.cs
@@ -304,6 +304,10 @@ namespace ClassicUO.Game.Scenes
             {
                 case MouseButtonType.Left: return OnLeftMouseDown();
                 case MouseButtonType.Right: return OnRightMouseDown();
+                case MouseButtonType.Middle:
+                case MouseButtonType.XButton1:
+                case MouseButtonType.XButton2:
+                    return OnExtraMouseDown(button);
             }
 
             return false;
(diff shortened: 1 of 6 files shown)


src/Game/Scenes/GameSceneInputHandler.cs
```diff
@@ -921,6 +945,21 @@ internal override bool OnMouseWheel(bool up)
                 }
             }
 
+            if (CanExecuteMacro())
+            {
+                Macro macro = Macros.FindMacro(up, Keyboard.Alt, Keyboard.Ctrl, Keyboard.Shift);
+
+                if (macro != null)
+                {
+                    if (macro.Items != null && macro.Items is MacroObject mac)
```
andreakarasho: `if (macro.Items is MacroObject mac)`
is enough, implicitly check for nullptr

src/Game/Scenes/GameSceneInputHandler.cs
```diff
@@ -902,6 +906,26 @@ private bool OnRightMouseDoubleClick()
...
 
+        private bool OnExtraMouseDown(MouseButtonType button)
+        {
+            if (CanExecuteMacro())
+            {
+                Macro macro = Macros.FindMacro(button, Keyboard.Alt, Keyboard.Ctrl, Keyboard.Shift);
+
+                if (macro != null && button != MouseButtonType.None)
+                {
+                    if (macro.Items != null && macro.Items is MacroObject mac)
```
andreakarasho: `if (macro.Items is MacroObject mac)`
is enough, implicitly check for nullptr

src/Game/Scenes/GameSceneInputHandler.cs
roxya: that's what I get for copying the existing code! fixed, thanks

src/Game/Scenes/GameSceneInputHandler.cs
roxya: fixed, thanks

PR #1568, opened 2022-11-02
Special Key Support

Ultima-Tony:
image


Ultima-Tony:
The code to add the keys might be overly broad, but I don't actually speak other languages, so I have no idea what keys are possible. This will hopefully cover all of them.

PR #1569, opened 2022-11-07
Fix bad dragging on Shop menu

Ultima-Tony:
Fixed the expander from being able to be dragging from off the button.


Ultima-Tony:
Previous Functionality:

bad drag shopmenu

My left mouse button is down the entire time, notice when it leaves the expander (from side to side) it will stop dragging.

Ultima-Tony:
Enhanced Functionality:

good drag shopmenu

PR #1572, opened 2022-12-11
don't update counter bar items when counter bar is disabled

roxya:
Minor thing, the counters continue to update even after you disable them, as the gump isn't disposed when disabling (in order to keep the counter bar items intact). I cannot push to the main branch


andreakarasho (approved):

andreakarasho:
@roxya FYI: I can't push to the `main` branch too because it's protected [unless forced]. It requires a review before the merge

PR #1576, opened 2022-12-24
Option to hue container gumps by the item hue

roxya:
image


andreakarasho:
I wonder if i'ts worth making this as a default feature. What do you think about?

roxya:
It's very nice for containers you have control over. Good enough to be the default for new profiles in my opinion.

People may find it annoying if friends/guild houses choose ugly colours. And there are people who customise their container gump art. But as long as it can be disabled it should be fine.

Let me know if you want me to update the PR with it as a default setting.

andreakarasho:
I think we can set it enabled by default.
Also there are conflicts with `ResGumps.resx` file. Probably due to the previous PRs
Thank you!

roxya:
Ok, set as default for new profiles/default button in the options page, and fixed the conflict.

andreakarasho (approved):

PR #1581, opened 2023-01-15
Update BoatMovingManager.cs

noobEone:
A byte can only hold a number up to 255.  Need to multiply by 10.

### DIFF
--- a/src/Game/Managers/BoatMovingManager.cs
+++ b/src/Game/Managers/BoatMovingManager.cs
@@ -64,7 +64,7 @@ namespace ClassicUO.Game.Managers
 
                 case 0x04: return FAST_INTERVAL;
 
-                case > 0x04: return speed * 10;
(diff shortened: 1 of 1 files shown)


src/Game/Managers/BoatMovingManager.cs
```diff
@@ -64,7 +64,7 @@ private static int GetVelocity(byte speed)
 
                 case 0x04: return FAST_INTERVAL;
 
-                case > 0x04: return speed;
+                case > 0x04: int Custom_Interval = speed * 10; return Custom_Interval;
```
andreakarasho: speed * 10 implicity creates an integer, you don't need to created a new variable.

noobEone: I originally just had it at speed * 10, but it threw errors in the build (macos-latest) check, creating a new variable gave it the green light

andreakarasho (approved):

PR #1591, opened 2023-02-15
NameOverhead item text changes

roxya:
- The OPL tends to have the quantity already so it's redundant when using tooltips.
- Evaluate plural strings in itemdata names like Diamond%s%

I don't know if it's worth using ValueStringBuilder instead of += concatenation here.


andreakarasho:
Concatenation is fine here, thank you for the PR!

andreakarasho (approved):

PR #1596, opened 2023-02-19
picinpic support for gumps sent by the server.

SaschaKP:
This should add the picinpic support for the gumps that are sent by the server.

This follows the issue: https://github.com/ClassicUO/ClassicUO/issues/1595

### DIFF
--- a/src/Game/UI/Controls/GumpPic.cs
+++ b/src/Game/UI/Controls/GumpPic.cs
@@ -182,4 +182,61 @@ namespace ClassicUO.Game.UI.Controls
             return base.Draw(batcher, x, y);
         }
     }
+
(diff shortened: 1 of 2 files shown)


src/Game/UI/Controls/GumpPic.cs
```diff
@@ -182,4 +182,61 @@ public override bool Draw(UltimaBatcher2D batcher, int x, int y)
             return base.Draw(batcher, x, y);
         }
     }
+
+    internal class GumpPicInPic : GumpPicBase
+    {
+        private int _SX, _SY;
```
andreakarasho: `private readonly int _sX, _sY;`

andreakarasho:
@SaschaKP Anything wrong with the implementation?

PR #1598, opened 2023-02-28
Decoupling, continued

jaedan:
1) Create the renderer library
2) Split Assets into Assets and IO

Neither of these libraries have the correct boundaries/APIs. Instead, this is split in the way that was the absolute smallest code change. I'll do another pull request with refactoring to get the boundaries/APIs/responsibilities right later on.


andreakarasho (commented):
I recently added https://github.com/ClassicUO/ClassicUO/blob/main/Directory.Build.props#L3-L14 file to replace commons properties between csproj. So you don't need anymore to specify some of these props

jaedan:
Updated to use Directory.Build.props

Issue #1608, opened 2023-04-03
Zones Feature Broken on Linux


ghost:
**Describe the bug**

Hi there, running ClassicUO using Mono with Linux, I observed the Zones feature doesn't work.

**To Reproduce**

1. Followed [instructions on settings up a .zones.json file in the Wiki](https://github.com/ClassicUO/ClassicUO/wiki/World-Map#zones-feature), using the [guardlines-to-polygons](https://github.com/rsuperior/guardlines-to-polygons) tool.

2. Feature doesn't work, tried clicking "Reload map zones" and restarting client.

3. Running ClassicUO from the command line, I get the following log message that seems relevant,

```
04/03/2023 17:00:27 |   Trace    | LoadZones()...
04/03/2023 17:00:27 |   Error    | System.IO.FileNotFoundException: Could not load file or assembly 'System.ServiceModel.Internals, Version=0.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' or one of its dependencies.
File name: 'System.ServiceModel.Internals, Version=0.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
  at System.Runtime.Serialization.XmlObjectSerializer.ReadObjectHandleExceptions (System.Runtime.Serialization.XmlReaderDelegator reader, System.Boolean verifyObjectName) [0x00000] in <98435247ecb949a6a66aa1e21386e749>:0 
  at System.Runtime.Serialization.Json.DataContractJsonSerializer.ReadObject (System.Xml.XmlDictionaryReader reader) [0x0000d] in <98435247ecb949a6a66aa1e21386e749>:0 
  at System.Runtime.Serialization.Json.DataContractJsonSerializer.ReadObject (System.IO.Stream stream) [0x00017] in <98435247ecb949a6a66aa1e21386e749>:0 
  at ClassicUO.Game.UI.Gumps.WorldMapGump+ZoneSets.AddZoneSetByFileName (System.String filename, System.Boolean hidden) [0x0001b] in <654f96af802149d183fc0a198064e7e8>:0 
```

As a workaround, I was able to monkey patch a fix by searching my machine for a file called `System.ServiceModel.Internals.dll` in my Mono installation, with the string `b77a5c561934e089` in its path as well. (That hex number taken from the `PublicKeyToken` part of the exception). I really have no idea what I'm doing, but it seems to fix the problem.

```
# Copy library I found to ClassicUO folder
cp -v \
[...]

PR #1609, opened 2023-04-03
Fix Map Zones on Linux (#1608)

ghost:
Here is a **naive** attempt to fix bug #1608, map zones on Linux.

From my investigation, it appears a library called `System.ServiceModel.Internals` is _not_  ncluded with the build. In Windows, perhaps the global system library is used, but that appears not to be the case in Linux. I'm not sure where the build decides to bundle DDLs, but I _think_ I've found the spot, in `tools/ManifestCreator/Program.cs`.

I unfortunately have no way of testing this bug. Thanks!


ghost:
Withdrawn, since it was fixed in a different way by @andreakarasho

PR #1613, opened 2023-04-10
The order in which items are shown in grid-loot will now depend on item type or its parameters.

GajuszZOrod:
Motivation: some items are likely to be always looted (e.g. gold, gems) and when looting is performed (usually automatically e.g. by Razor macros, but also manually by other players) it makes other items to move in a grid making it harder to browse their properties. Hence, items like gold should be at the end of the grid. Stackable items have no custom properties, so displaying them at the bottom of the grid is beneficial for players: it is more convenient to browse the loot and allows saving the time.

### DIFF
--- a/src/ClassicUO.Client/Game/UI/Gumps/GridLootGump.cs
+++ b/src/ClassicUO.Client/Game/UI/Gumps/GridLootGump.cs
@@ -222,36 +222,44 @@ namespace ClassicUO.Game.UI.Gumps
             int line = 1;
             int row = 0;
 
-            for (LinkedObject i = _corpse.Items; i != null; i = i.Next)
+            for (int displayGroup = 0; displayGroup < 2; displayGroup++)
             {
-                Item it = (Item) i;
-
-                if (it.IsLootable)
+                for (LinkedObject i = _corpse.Items; i != null; i = i.Next)
                 {
-                    GridLootItem gridItem = new GridLootItem(it, GRID_ITEM_SIZE);
+                    Item it = (Item)i;
 
-                    if (x >= MAX_WIDTH - 20)
+                    if (!ItemBelongsToGroup(it, displayGroup))
                     {
-                        x = 20;
-                        ++line;
-
-                        y += gridItem.Height + 20;
-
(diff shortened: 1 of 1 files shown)


GajuszZOrod:
Please let me know in case if I need to raise an issue ticket first, in order to agree upon the feature implemented in PR.

src/ClassicUO.Client/Game/UI/Gumps/GridLootGump.cs
```diff
@@ -222,36 +222,44 @@ protected override void UpdateContents()
...
-                    if (x >= MAX_WIDTH - 20)
+                    if (!ItemBelongsToGroup(it, displayGroup))
                     {
-                        x = 20;
-                        ++line;
+                        continue;
+                    }
 
-                        y += gridItem.Height + 20;
+                    if (it.IsLootable)
```
andreakarasho: probably better to invert the `if` statement and continue [readability stuff]

src/ClassicUO.Client/Game/UI/Gumps/GridLootGump.cs
```diff
@@ -295,6 +303,14 @@ protected override void UpdateContents()
...
         }
 
+        private bool ItemBelongsToGroup(Item it, int group)
+        {
+            // Note: items must be assigned to groups in a mutually-exclusive manner, so that each item occurs only once in the grid
+            if (it.ItemData.IsStackable)
+                return group > 0;
+            else
+                return group == 0;
+        }
```
andreakarasho, line 310: Add an empty line as functions separator

src/ClassicUO.Client/Game/UI/Gumps/GridLootGump.cs
GajuszZOrod: Done.

src/ClassicUO.Client/Game/UI/Gumps/GridLootGump.cs
GajuszZOrod: Done.

PR #1670, opened 2024-02-19
Fix race condition when handling plugins packets

### LINKED ISSUE #1618: System.ArgumentException: Destination is too short
######################## [START LOG] ########################
ClassicUO [STANDARD_BUILD] - 0.1.11.83 - 04/27/2023 23:38:01
OS: Win32NT x64
Thread: 

Shard: play.trueuo.com
ClientVersion: 7.0.98.1

Exception:
System.ArgumentException: Destination is too short.
   at System.ThrowHelper.ThrowArgumentException_DestinationTooShort()
   at ClassicUO.Network.CircularBuffer.Enqueue(Span`1 buffer, Int32 offset, Int32 size) in /home/runner/work/ClassicUO/ClassicUO/src/ClassicUO.Client/Network/CircularBuffer.cs:line 104
   at ClassicUO.Network.PacketHandlers.Append(Span`1 data, Boolean fromPlugins) in /home/runner/work/ClassicUO/ClassicUO/src/ClassicUO.Client/Network/PacketHandlers.cs:line 145
   at ClassicUO.Network.Plugin.OnPluginRecv(Byte[]& data, Int32& length) in /home/runner/work/ClassicUO/ClassicUO/src/ClassicUO.Client/Network/Plugin.cs:line 658
   at Assistant.Engine.SendPacketToClient(Byte[] packet, Int32 length, Boolean delay) in c:\Projects\ClassicAssist\ClassicAssist\Engine.cs:line 783
   at ClassicAssist.UO.Commands.SystemMessage(String text, Int32 hue, Boolean suppressInQuietMode, Boolean throttleRepeating) in c:\Projects\ClassicAssist\ClassicAssist\UO\Commands.cs:line 200
   at ClassicAssist.Data.Macros.MacroEntry.OnExceptionEvent(Exception exception) in c:\Projects\ClassicAssist\ClassicAssist\Data\Macros\MacroEntry.cs:line 317
   at ClassicAssist.Data.Macros.MacroInvoker.<>c__DisplayClass40_0.<Execute>b__0() in c:\Projects\ClassicAssist\ClassicAssist\Data\Macros\MacroInvoker.cs:line 259
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
######################## [END LOG] ########################

### LINKED ISSUE #1665: Race condition in handling plugins packets
I first noticed this with ClassicUO 1.0.0.0, I was developing an autolooter with Razor Enhanced 0.8.2, and for debugging reasons I was printing item properties both in System messages and Overhead messages (with basically no pause).

With that script it seems I can cause several different issues in ClassicUO:
1. 1-2 Messages are printed, then ClassicUO crashes due to an error in creating a Texture2D, basically the same error as: https://github.com/ClassicUO/ClassicUO/issues/1616

I looked a bit into this, I think the logic in ClassicUO is missing some safety checks to avoid the crash, but in theory the safety check would have to result in a disconnect anyway (because I think the packet is being read from corrupted buffer).

2. Few messages are printed, some are corrupted (random Japanese characters) and then a crash. This one was similar to: https://github.com/ClassicUO/ClassicUO/issues/1618, but it was actually inside SetCapacity. It was trying to copy the old buffer over the new, but with a Span of size bigger than the old buffer.

3. Corrupted characters, the script end up "working" (it moves some selected items from a container to another), but then looking at the console, it's spamming "need more data" for some random but fixed packet ID.

After the whole day behind this I've been able to catch ClassicUO in this state:
image

While it cannot be seen for the CUO_MAIN_THREAD which instance of a circular buffer is being Dequeued from, I can see from the debugger that it's using `_pluginsBuffer`, which is also being used by that other thread, which is Enqueueing data.

While there's this lock in `OnPluginRecv`:
https://github.com/ClassicUO/ClassicUO/blob/c209bfe38619c44c8dda141627773430cf49687a/src/ClassicUO.Client/Network/Plugin.cs#L788-L791

There is no lock in `Update`:
https://github.com/ClassicUO/ClassicUO/blob/c209bfe38619c44c8dda141627773430cf49687a/src/ClassicUO.Client/GameController.cs#L368

# Resolution
I'm still not sure yet, to be fair I just recently digged in to the code; slapping a lock in `GameController::Update` might work, but I see that there are other accesses to `PacketHandler.Handler` that are not protected, and I wonder if it's a bigger problem of having to restructure the flow of how handlers are accessed, because it's not always enough to lock the resource accessed concurrently; there might be other surrounding code that has to behave atomically.

Furthermore `ParsePackets(CircularBuffer stream, bool allowPlugins)` locks the CircularBuffer itself, but I'm not sure where else is a concurrent access that it's using or might use that lock to properly synchronize access.

Spasitjel:
There's a race condition in the access of the packets CircularBuffers, between the render thread, which reads the packets from the Server->Client buffer, the Plugin->Client buffer and the Plugins thread which can enqueue new packets to be read by the Client.

Add a lock of the Handler in the render thread.

Fixes #1665
Fixes #1618 

I've tested this change for a couple of weeks and honestly I only saw an improvement in stability when using scripts or hotkeys in Razor Enhanced.
I double checked again and it should be all its necessary.

NOTE: It should also fix the sporadic error in https://github.com/ClassicUO/ClassicUO/issues/1616, which seems to be caused by the Client reading garbage packet data and interpreting map sizes incorrectly.
I think there can be a change made to put a limit in what a packet can say the map size is, to avoid even trying to allocate such a big Texture, which is why I wasn't sure if it made sense to include that in the fixed issues.

NOTE2: While the fix works, I think in the future an improvement can be made to the API to avoid locking the CircularBuffer here:
https://github.com/ClassicUO/ClassicUO/blob/fe11d90b84d77420bc59a82c50a7c01c113d0df2/src/ClassicUO.Client/Network/PacketHandlers.cs#L98
and the Handler, because one of the two seems redundant if we reorder where the locks are put.
Ideally we would lock only the CircularBuffer, not the whole Handler.

### DIFF
--- a/src/ClassicUO.Client/GameController.cs
+++ b/src/ClassicUO.Client/GameController.cs
@@ -365,8 +365,10 @@ namespace ClassicUO
             Mouse.Update();
 
             var data = NetClient.Socket.CollectAvailableData();
-            var packetsCount = PacketHandlers.Handler.ParsePackets(UO.World, data);
-
+            int packetsCount = 0;
+            lock(PacketHandlers.Handler) {
+                packetsCount = PacketHandlers.Handler.ParsePackets(UO.World, data);
+            }
             NetClient.Socket.Statistics.TotalPacketsReceived += (uint)packetsCount;
             NetClient.Socket.Flush();
 


Spasitjel:
Unfortunately, this is not fully correct, the issue is more complex, more context in the original issue.

PR #1676, opened 2024-04-02
Support 16k hues

kaczy93:
This puts all the hues into a single texture, 16 per row, 1024 rows.
I tested it extensivly on a standard file copied over 5 times, so total of 15k hues and no shifting due to precision or other artifacts


andreakarasho (approved):

andreakarasho:
LGTM
Thank you!

PR #1741, opened 2025-01-07
Some NetClient robustness using async

MaxKellermann:
This PR makes NetClient more robust by making Connect() and Read() async, which gives more direct UI feedback on connect errors and allows cancellation at any time, without freezing the UI.

An exotic feature is the DNS resolver; I helped a new shard with their DDoS protection and wrote an UO login server + proxy (https://github.com/MaxKellermann/uologin/) based on code from my uoproxy project (https://github.com/MaxKellermann/uoproxy). They have multiple such proxies, and that certain bad actor who DDoSes all the shards in order to extort money from them (you may of heard of this guy) has so far failed to bring down the shard. Having multiple proxies needs client support, and this is the code I wrote for the (proprietary) ClassicUO fork of that shard, which I am hereby offering to the public.

Later, I'd like to implement transparent reconnect; i.e. if you lose the connection, I want the client to reconnect automatically and instantly with no visible effect (only lagging for a second or so); not going through LoginScene, not losing anything on the UI, all container gumps stay where they are. This, too, can help with making connections to DDoSed shards more reliable; if a proxy goes down, all users shall automatically switch to the next one. This code will be based on the "async" refactoring in this PR.

I modified the WebSocket code but havn't tested this - but how can I test this piece of code? What shard will allow me to test WebSocket support?

This PR is based on https://github.com/ClassicUO/ClassicUO/pull/1739 i.e. is BSD2-licensed. Merge only after relicensing has been approved.

It's a large PR - if you prefer to split it, let me know.


jabinb:
I'm against adding these "DDoS protection" changes, i.e. the DNS resolver and any future proxying features, they're too specific should be kept to a fork. 

Particularly the change in this PR adds an extra DNS request slowing down login for all users regardless of whether the shard uses this feature. Future changes for proxying support would add code complexity and maintenance overhead for what is so far a small audience (1 shard, uosagas), and create a feature-lock in to your project.

Additionally it's effectively useless, extra DNS addresses as you present it in this PR is not a protection for DDoS attacks. It's merely a gimmick as once the attacker learns how to use DNS to request all addresses it's over. This also ignores that the attacker can bypass needing to flood the login servers by logging in legitimately, inspecting the game login packet (0x8C) and retrieve the game server IP, where they can more effectively DDoS the shard directly.

After skimming your project I see you understand there's more needed to prevent DDoS attacks as you're also blocking traffic via a udp knock + nft rule, and likely run another proxy for the game server itself. However I see these setups as too shard specific and don't belong in upstream CUO.

MaxKellermann:
> Particularly the change in this PR adds an extra DNS request slowing down login for all users regardless of whether the shard uses this feature. Future changes for proxying support would add code complexity and maintenance overhead for what is so far a small audience (1 shard, uosagas), and create a feature-lock in to your project.

This is not true. There are no extra DNS requests and thus no overhead/slowdown due to that commit.

Without this PR, the DNS request (actually a request to the system resolver which may or not cause an actual DNS request) is done implicitly by `TcpClient.Connect(string)`.

With this PR, the DNS request (again: system resolver) is done explicitly via `Dns.GetHostAddressesAsync()` and then `TcpClient.Connect(IPAddress)` is called.

> Additionally it's effectively useless, extra DNS addresses as you present it in this PR is not a protection for DDoS attacks. It's merely a gimmick as once the attacker learns how to use DNS to request all addresses it's over. This also ignores that the attacker can bypass needing to flood the login servers by logging in legitimately, inspecting the game login packet (0x8C) and retrieve the game server IP, where they can more effectively DDoS the shard directly.

I think you misunderstand how uologin works. It proxies the whole connection through the initial AccountLogin; there is no GameLogin at all. The attacker doesn't know the game server IP address.

For an attacker, this multiplies the cost of DDoSing a shard by the number of login servers, which turned out to be prohibitive for that shard. The shard survived all the attacks without any problems.

gaechti:
you should probably remove the license change from the pr :)

MaxKellermann:
> you should probably remove the license change from the pr :)

That was an explicit decision I explained in the PR text.

jabinb:
> This is not true. There are no extra DNS requests and thus no overhead/slowdown due to that commit.
> Without this PR, the DNS request (actually a request to the system resolver which may or not cause an actual DNS request) is done implicitly by TcpClient.Connect(string).
> With this PR, the DNS request (again: system resolver) is done explicitly via Dns.GetHostAddressesAsync() and then TcpClient.Connect(IPAddress) is called.

You're right I can see that now, however the point below still stands.

> I think you misunderstand how uologin works. It proxies the whole connection through the initial AccountLogin; there is no GameLogin at all. The attacker doesn't know the game server IP address.
> For an attacker, this multiplies the cost of DDoSing a shard by the number of login servers, which turned out to be prohibitive for that shard. The shard survived all the attacks without any problems.

How your service works is irrelevant, we are speaking about the DNS changes in this PR and the changes here do not offer any protection therefore they are useless. Furthermore even if it did come with all the integration required to operate with your `uologin` project it wouldn't matter as DDoS protection is not a client concern, it is a responsibility for the shard to solve. It's well known that **Upstream CUO is not meant for features specific to one shard**.

Let me be clear, I'm not against CUO making life easier for shard owners by being friendly to DDoS mitigations but I see no benefit in these changes. If the shard owner needs to resort to forking CUO to implement their homegrown DDoS solution then that is their prerogative.

MaxKellermann:
> > This is not true. There are no extra DNS requests and thus no overhead/slowdown due to that commit.
> > Without this PR, the DNS request (actually a request to the system resolver which may or not cause an actual DNS request) is done implicitly by TcpClient.Connect(string).
> > With this PR, the DNS request (again: system resolver) is done explicitly via Dns.GetHostAddressesAsync() and then TcpClient.Connect(IPAddress) is called.
> 
> You're right I can see that now, however the point below still stands.
> 
> > I think you misunderstand how uologin works. It proxies the whole connection through the initial AccountLogin; there is no GameLogin at all. The attacker doesn't know the game server IP address.
> > For an attacker, this multiplies the cost of DDoSing a shard by the number of login servers, which turned out to be prohibitive for that shard. The shard survived all the attacks without any problems.
> 
> How your service works is irrelevant, we are speaking about the DNS changes in this PR and the changes here do not offer any protection therefore they are useless.

It is effectively useless for most shards as they usually have only one IP address. Useless as in "makes no difference". They have no advantage, but OTOH they have no disadvantage either.

Still it's not "useless" - just because most shards don't have it doesn't mean it's correct. The CUO code is actually incorrect. If you want to connect to a host name, you should try all the addresses. Proper network programs do that, but CUO does not, because of dotNET's bad defaults.

Shards using uologin (which other shards may adopt eventually because everybody is riddled by this bad actor) are one use case.
But there may be other reasons to announce multiple addresses via DNS: for example, if you host your shard in a datacenter that is connected to multiple backbones and your shard server has multiple IP addresses. If one route is down, CUO should just fall back to the other one.
Other example: IPv4+IPv6. Have both an A and an AAAA record, announce both addresses, and CUO just uses the one that works.

The browser you're currently using to read this sentence implements this just like this PR: having multiple addresses is very common for web servers, and thus browsers - of course! - implement this.

> Furthermore even if it did come with all the integration required to operate with your `uologin` project it wouldn't matter as DDoS protection is not a client concern, it is a responsibility for the shard to solve. It's well known that **Upstream CUO is not meant for features specific to one shard**.

I tried to explain that CUO's current implementation is bad, and doing it the way I suggest is common sense and good practice, and not just a feature for one specific shard. Even if most CUO users won't see a difference.

> Let me be clear, I'm not against CUO making life easier for shard owners by being friendly to DDoS mitigations but I see no benefit in these changes. If the shard owner needs to resort to forking CUO to implement their homegrown DDoS solution then that is their prerogative.

The usefulness of a software feature like this cannot be measured by "what percentage of users will see a benefit". What I suggest is more correct, independent of what one specific shard is doing.

Of course, every shard can fork CUO, and I guess most of them do. Which I find sad, because a central project that is universal for all shards would have a greater chance of concentrating improvements and distributing them to everybody. 
By rejecting that idea and forcing shard owners to fork CUO, chances are lower that improvements get fed back here.

Just imagine every web site would require to you use their Chrome fork. They need to make forks because Google refuses to integrate IPv6 support. And some websites cannot be used on Apple and Linux because the webmaster doesn't provide Apple/Linux builds. And if they do provide Linux builds, it's only x86-64, but not Arm64, so the one guy with an Arm64 workstation (i.e. me) cannot every play.
Sounds silly, doesn't it?
But that's how your CUO policy sounds to me.

(But yes, of course, if that commit stands in the way of merging this PR, I'll remove it.)

jabinb:
After learning about your previous [interactions with other projects in the UO community](https://github.com/modernuo/ModernUO/pull/2034#issuecomment-2566292877) combined with your tone and approach in this PR [and other issues](https://github.com/ClassicUO/ClassicUO/issues/1738) we've decided your interactions with the community present a hostile and abrasive pattern of behaviour that we are not interested in bringing into the project.

While we appreciate your technical skills and efforts, we will not be accepting these contributions or any others, please refrain from interacting with the project in the future. Thanks for your time we wish you the best.

Reetus:
For what it's worth...

IF dotnet/TcpClient/whatever doesn't correctly round robin a host with multiple IP addresses...

My quick test with TcpClient shows it did 1 out of 10 times...

image

Then I think it's a reasonable request, and definately less niche than WebSocket connect support, which I assume was added for Outlands' benefit, as I don't know any public UO server project that supports it...

There this small little UO shard that has multiple IP's on their hostname...

image

jaedan:
WebSocket support was added to reduce the difference between this version and webcuo. Outlands doesn't use it.

But as a DDoS defense strategy, I think tunneling everything through WebSockets and hiding behind cloudflare is the absolute best strategy from both cost and effectiveness standpoints for all shards. I would absolutely love to see first class websocket support show up in something like ModernUO that interoperates with the websocket support in classicuo. That's just really good for the whole community, and it's a much more robust solution than what this PR attempts to do by allowing shard admins to run their own sets of proxies. So this PR is already closed but I would have voted no and suggested the websocket route instead.

Reetus:
Maybe it's me, but testing the WebSocket stuff, it fails to login when the character is in a location with alot of mobiles...

I think it's too new, where this presumably ready to go...

Also, async connecting is desperately needed so when it doesn't connect, you don't have to do this all the time....

image

PR #1755, opened 2025-02-15
Add drag-select hostiles only option

### LINKED ISSUE #1746: Feature Request: Drag-select filter to only select hostile mob/humanoids
**Describe the feature**
It would be nice if there were an additional drag-select filter that only selected hostile mob/humanoids. This would be the opposite of "humanoids only" with an enhancement that it only gets health bar for hostiles. I don't want to select my horse, but I do want to select something attacking me.

It should also ignore my own character. I don't want to drag a health bar out of my  character, **ever**. I already have the status bar where I want it.

jabinb:
resolves #1746

https://github.com/user-attachments/assets/f2a3418b-612b-49a2-ade3-45cb7e094c5f


andreakarasho (approved):

E-Thang:
Thank you!

PR #1757, opened 2025-02-23
Fix map block 0,0 for mul format

kaczy93:
0 address is perfectly valid for for both map and statics in mul format.


andreakarasho:
closed because merged with https://github.com/ClassicUO/ClassicUO/pull/1758
Thanks btw

PR #1759, opened 2025-02-23
Add a simple dps meter to damage numbers

bittiez:
# DPS meter

This is a simple DPS meter that can be toggled on/off

This simply shows a DPS meter anytime damage numbers are shown, now because the client doesn't know who dealt the damage(You or your pets for example), it is technically a damage-take dps not a damage-dealt dps.

image
image


jabinb (changes requested):
First of all thank you, sorry to hear about the discord/repo being deleted, we appreciate the PR.

Can we make this use a circular array with a moving average function instead of the loops we're doing here? It might be useful in other places too, there should be a fair few drop-in implementations around on StackOverflow or elsewhere.

Also another improvement we can make is to reduce the calls to `UtcNow` and `TimeSpan.FromSeconds` each loop. I think using `Timer.Ticks` (or `Delta`) instead will be nicer as it's is computed per frame and just a simple `uint`.

bittiez:
> First of all thank you, sorry to hear about the discord/repo being deleted, we appreciate the PR.
> 
> Can we make this use a circular array with a moving average function instead of the loops we're doing here? It might be useful in other places too, there should be a fair few drop-in implementations around on StackOverflow or elsewhere.
> 
> Also another improvement we can make is to reduce the calls to `UtcNow` and `TimeSpan.FromSeconds` each loop. I think using `Timer.Ticks` (or `Delta`) instead will be nicer as it's is computed per frame and just a simple `uint`.

good points, I will rework this when I get some time next thank you

andreakarasho:
There are conflicts due to other PRs

andreakarasho:
There are conflicts due to other PRs

bittiez:
> There are conflicts due to other PRs

Should be good now, not sure what happen but it looks like my ResGumps got a little wonky switching between branches and the other branch already had the entry needed for this, my bad I will have to pay closer attention to that.

jabinb (approved):

andreakarasho (approved):

PR #1760, opened 2025-02-23
Allow setting your current profile as the default for new characters

### LINKED ISSUE #1675: Default Profile for all the UO settings
This is not a bug, but more of a feature request, would it be possible to make a Configurable Default Profile, which will be used for all New accounts and Characters,
Right now Profiles are only per character, it would be a huge improvement, especially for new Players, if It would be possible to Configure Default Profile, what then would be used for all characters and Accounts, this would give really nice option for shard owners

There are many things that could be defaulted, to Improve, new player experience
- resolution
- video settings
- suggested macros
- Cooldown bars
- Container settings
- various options like grid inventory and other CUO-specific things

This should not be Hard Coded, but be optional and configurable.

bittiez:
image

This allows players to set their current profile as the default for new characters so they don't need to redo all the settings they usually set up.

Resolves #1675

### DIFF
--- a/src/ClassicUO.Client/Resources/ResGumps.resx
+++ b/src/ClassicUO.Client/Resources/ResGumps.resx
@@ -1610,7 +1610,7 @@ Client version: '{0}'</value>
     <value>Show DPS with damage numbers</value>
   </data>
   <data name="SetAsNewDefault" xml:space="preserve">
-    <value>Set current profile as default for new characters</value>
+    <value>Save current settings as defaults for new users</value>
   </data>
   <data name="UseAlternateJournal" xml:space="preserve">
     <value>Use alternative journal</value>
--- a/src/ClassicUO.Client/Configuration/Profile.cs
+++ b/src/ClassicUO.Client/Configuration/Profile.cs
@@ -360,6 +360,11 @@ namespace ClassicUO.Configuration
             Log.Trace("Saving done!");
         }
 
+        public void SaveAs(string path, string filename = "default.json")
+        {
+            ConfigurationResolver.Save(this, Path.Combine(path, filename), ProfileJsonContext.DefaultToUse.Profile);
+        }
+
         private void SaveGumps(World world, string path)
         {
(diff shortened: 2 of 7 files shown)


src/ClassicUO.Client/Resources/ResGumps.resx
```diff
@@ -1606,4 +1606,10 @@ Client version: '{0}'</value>
   <data name="DragHostileOnly" xml:space="preserve">
     <value>Select hostiles only</value>
   </data>
+  <data name="ShowDPSWithDamage" xml:space="preserve">
+    <value>Show DPS with damage numbers</value>
```
jabinb, line 1610: Leftover from #1759 might want to remove to prevent a conflict later.

src/ClassicUO.Client/Resources/ResGumps.resx
```diff
@@ -1606,4 +1606,10 @@ Client version: '{0}'</value>
...
+  </data>
+  <data name="SetAsNewDefault" xml:space="preserve">
+    <value>Set current profile as default for new characters</value>
```
jabinb: This initially confused me as I first thought it meant this profile will be copied to a new character (at creation). But I see it gets saved as a copy when you press the button, then no changes after that are persisted.

I think if we use `Save` instead of `Set` it will help? Also I think using `settings` not `profile` as the rest of the profile data like gumps/macros etc aren't copied too.

`Save current settings as defaults for new users`, then they know its whatever the settings are now is what gets saved.

jabinb (approved):
Looks good just minor things

src/ClassicUO.Client/Resources/ResGumps.resx
bittiez: Whoops, will fix when I have time next

src/ClassicUO.Client/Resources/ResGumps.resx
bittiez: Good points, I like your wording, I'll update it when I get some time

PR #1761, opened 2025-02-23
Add an alternative journal

bittiez:
image
image
image

This PR would bring an alternative journal for players to choose from. It has customizable tabs based on message types.


src/ClassicUO.Client/Configuration/Profile.cs
```diff
@@ -313,6 +314,39 @@ internal sealed class Profile
 
         public static uint GumpsVersion { get; private set; }
 
+        //Alternate Journal
+        public bool UseAlternateJournal { get; set; } = false;
+        public Dictionary<string, MessageType[]> JournalTabs { get; set; } = new Dictionary<string, MessageType[]>()
```
andreakarasho: `set;` is not necessary

src/ClassicUO.Client/Configuration/Profile.cs
```diff
@@ -313,6 +314,39 @@ internal sealed class Profile
 
         public static uint GumpsVersion { get; private set; }
 
+        //Alternate Journal
+        public bool UseAlternateJournal { get; set; } = false;
```
andreakarasho: `= false` it's not necessary

src/ClassicUO.Client/Game/Managers/MacroManager.cs
```diff
@@ -736,6 +736,19 @@ private int Process(MacroObject macro)
...
 
                                 case MacroSubType.Journal:
+                                    if(ProfileManager.CurrentProfile.UseAlternateJournal)
+                                    {
+                                        ResizableJournal rjournal = UIManager.GetGump<ResizableJournal>();
+                                        if (macro.Code == MacroType.Close)
+                                        {
+                                            rjournal?.Dispose();
+                                        }
+                                        else if (macro.Code == MacroType.Open)
```
andreakarasho: The `MacroType.Open` should be handled by the `GameActions.OpenJournal` function like for std journal:
https://github.com/ClassicUO/ClassicUO/blob/a244efff3822c9186918b70cc83a289a21cd242c/src/ClassicUO.Client/Game/Managers/MacroManager.cs#L518

andreakarasho (changes requested):
Thanks for your PR(s).
There are few small things than needs to be addressed before merging.

bittiez:
Thanks Karasho, I'll address these when I get some time next

src/ClassicUO.Client/Game/GameActions.cs
```diff
@@ -107,6 +107,12 @@ public static void OpenStatusBar(World world)
 
         public static void OpenJournal(World world)
         {
+            if (ProfileManager.CurrentProfile.UseAlternateJournal)
+            {
+                UIManager.Add(new ResizableJournal(world));
```
andreakarasho, line 112: I missed that:
I think that part should take in account when the journal gump already exists, otherwise you can spawn more than 1 instance of that journal and I don't think that's the intended behaviour

andreakarasho (changes requested):

src/ClassicUO.Client/Game/GameActions.cs
bittiez: It was intended actually, allowing people to have multiple journals showing different tabs, let me know if you don't want that behaviour

andreakarasho: That's fine

andreakarasho (approved):

PR #1764, opened 2025-02-26
Add basic controller support

bittiez:
## This PR aims to add some basic controller support(That can be expanded later on)

Currently this PR provides:  
1. Macro hotkeys can be controller button(s)  
2. Mouse movement via joystick(right joystick)  
3. Mouse left and right click via clicking thumbsticks down  
4. UI button presses(Currently only used for setting hotkeys, but can be built upon to navigate various gumps with controllers potentially)  
5. Character movement via left joystick  

## Media
image

https://github.com/user-attachments/assets/d5638a0a-5fe4-44ef-9f54-9cb054ebebad

### DIFF
--- a/src/ClassicUO.Client/Game/Managers/MacroManager.cs
+++ b/src/ClassicUO.Client/Game/Managers/MacroManager.cs
@@ -17,6 +17,7 @@ using ClassicUO.Resources;
 using ClassicUO.Utility;
 using ClassicUO.Utility.Logging;
 using SDL2;
+using static SDL2.SDL;
 
 namespace ClassicUO.Game.Managers
 {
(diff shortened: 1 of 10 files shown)


src/ClassicUO.Client/Game/Managers/MacroManager.cs
```diff
@@ -2016,6 +2050,24 @@ public void Load(XmlElement xml)
...
+            XmlElement buttons = xml["controllerbuttons"];
+
+            if (buttons != null)
+            {
+                List<SDL.SDL_GameControllerButton> savedButtons = new List<SDL_GameControllerButton>();
+                foreach (XmlElement buttonNum in buttons.GetElementsByTagName("button"))
+                {
+                    if (int.TryParse(buttonNum.InnerText, out int b))
+                    {
+                        if (Enum.IsDefined(typeof(SDL_GameControllerButton), b))
```
andreakarasho: Use this https://learn.microsoft.com/en-us/dotnet/api/system.enum.isdefined?view=net-9.0#system-enum-isdefined-1(-0) to avoid reflection

andreakarasho (changes requested):

src/ClassicUO.Client/Game/Managers/MacroManager.cs
bittiez: Fixed

bittiez:
Closing, seems no interest here, code is still there if someone wanted to reopen down the road

PR #1765, opened 2025-02-26
Overhead party messages

bittiez:
## This PR adds an option to display **party** messages over head

Since party messages don't display overhead, and usually for parties you are working with each other fairly closely this is helpful to easily see messages similar to if they were chatting normally.

Default is off  

image

### DIFF
--- a/src/ClassicUO.Client/Game/Managers/MessageManager.cs
+++ b/src/ClassicUO.Client/Game/Managers/MessageManager.cs
@@ -103,7 +103,7 @@ namespace ClassicUO.Game.Managers
                                 }
                     }
 
-                    if (type != MessageType.Spell && _world.IgnoreManager.IgnoredCharsList.Contains(parent.Name))
+                    if (type != MessageType.Spell && parent != null && _world.IgnoreManager.IgnoredCharsList.Contains(parent.Name))
                         break;
 
                     //Add null check in case parent was not found above.
--- a/src/ClassicUO.Client/Game/Managers/MessageManager.cs
+++ b/src/ClassicUO.Client/Game/Managers/MessageManager.cs
@@ -85,6 +85,40 @@ namespace ClassicUO.Game.Managers
                 case MessageType.Encoded:
                 case MessageType.System:
                 case MessageType.Party:
+                    if (!currentProfile.OverheadPartyMessages)
+                        break;
+
+                    if (parent == null)
+                    {
+                        bool handled = false;
+                        foreach (PartyMember member in _world.Party.Members)
+                            if (member != null)
+                                if (member.Name == name)
+                                {
+                                    Mobile m = _world.Mobiles.Get(member.Serial);
+                                    if (m != null)
+                                    {
+                                        parent = m;
+                                        handled = true;
+                                        break;
+                                    }
+
+                                }
+                        if (!handled) break;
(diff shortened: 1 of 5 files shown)


src/ClassicUO.Client/Game/Managers/MessageManager.cs
```diff
@@ -85,6 +85,40 @@ public void HandleMessage
...
+                                        parent = m;
+                                        handled = true;
+                                        break;
+                                    }
+
+                                }
+                        if (!handled) break;
+                    }
+
+                    if (_world.IgnoreManager.IgnoredCharsList.Contains(parent.Name) && type != MessageType.Spell)
```
andreakarasho: put the `type != MessageType.Spell` before so the `.Contains` doesn't get triggered too much

src/ClassicUO.Client/Game/Managers/MessageManager.cs
```diff
@@ -85,6 +85,40 @@ public void HandleMessage
...
+                                    Mobile m = _world.Mobiles.Get(member.Serial);
+                                    if (m != null)
+                                    {
+                                        parent = m;
+                                        handled = true;
+                                        break;
+                                    }
+
+                                }
+                        if (!handled) break;
```
andreakarasho: the `handled` variable is not necessary here because you already have `parent` as a valid check.
So replacing `if (!handled) break;` with `if (parent == null) break;` would be enough

andreakarasho (changes requested):

src/ClassicUO.Client/Game/Managers/MessageManager.cs
bittiez: I could rework it to do another parent check, but the handled is there because if there is no matching party member that is in world.Mobiles it needs to break early before adding message to the parent later

parent is assigned if there is a party member whos name matches the name in the party message from the server

So I either need the handled bool for after the foreach, or I need to recheck if parent still == null

Or I can move the ignore check and AddMessage parts directly in the switch which may be better

src/ClassicUO.Client/Game/Managers/MessageManager.cs
bittiez: Done 👍

src/ClassicUO.Client/Game/Managers/MessageManager.cs
```diff
@@ -85,6 +85,37 @@ public void HandleMessage
...
+                                    Mobile m = _world.Mobiles.Get(member.Serial);
+                                    if (m != null) //Mobile exists
+                                    {
+                                        parent = m;
+                                        break;
+                                    }
+                                }
+                    }
+
+                    if (type != MessageType.Spell && _world.IgnoreManager.IgnoredCharsList.Contains(parent.Name))
```
andreakarasho: missing a null check for the `parent`

andreakarasho (changes requested):
LGTM. Just last one thing to fix

src/ClassicUO.Client/Game/Managers/MessageManager.cs
bittiez: Good catch

andreakarasho (approved):

PR #1774, opened 2025-03-06
refactor/GameController-constructor

Hyeoncheol-Kang:
After the CUO "net8 porting," some commands in ClassicAssist started to throw errors. Upon investigation, I found that the `GameController.cs` class, which was previously defined as:

```csharp
public GameController()
```

was changed to:

```csharp
public GameController(IPluginHost pluginHost)
```

After this change, it seems that ClassicAssist is no longer able to access it via reflection. So, I modified the structure there.

### DIFF
--- a/src/ClassicUO.Client/GameController.cs
+++ b/src/ClassicUO.Client/GameController.cs
@@ -64,7 +64,24 @@ namespace ClassicUO
         
         public GameController(IPluginHost pluginHost)
         {
-            GameController();
+            GraphicManager = new GraphicsDeviceManager(this);
+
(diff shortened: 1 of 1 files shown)


Hyeoncheol-Kang:
```
// Default constructor
public GameController() 
{
    // Functionality is the same as the other constructor,
    // except for this line being omitted.
}

// Constructor with IPluginHost parameter
public GameController(IPluginHost pluginHost) 
{
    // This constructor includes the assignment:
    PluginHost = pluginHost;
}
```

two constructors...

Sorry about that, I accidentally committed multiple times. :)

PR #1788, opened 2025-06-22
Fixed a crashbug related to the server sending item IDs that are not found

Lichtblitz:
Instead the client now degrades more gracefully by loading idx '0' instead and logging an error


Lichtblitz:
Almost the entire PR is just indentation of the remaining code after introducing the short-circuit

andreakarasho (approved):

PR #1791, opened 2025-07-21
Picked up stale branch new-counterbar

### LINKED ISSUE #1216: Counter Bar Enhancements

Lichtblitz:
This replaces #1245

Previous features:

1. doubleclick on empty space will enable/disable resizing
2. drag & drop items over empty space to add items in sequence [replace of the awful & useless grid system]
3. auto item arrange

Added Features:

4. Placeholders can be added/removed to help with organizing the items in a certain way for those who wish to replicate the previous grid layout

https://github.com/user-attachments/assets/52e9225e-c3d2-4911-b378-491c0fe81fdf

6. it is now possible to show each value as a relative amount instead of an absolute; just set it before a crafting session or a dungeon run and continuously see how the amount in reagents, bandages, gold, etc. changed compared to the start

https://github.com/user-attachments/assets/a94413c8-9565-40a0-ad94-68e8bb090ca5

7. allow rearranging items through drag and drop; if that's too complex to implement, there needs to be some other option

https://github.com/user-attachments/assets/75db3430-3210-4963-9278-cd72b100230e

8. allowing an entry to ignore hues; on most shards bandages are the same, no matter the hue, for example

https://github.com/user-attachments/assets/e2bfbaf9-16fa-4865-858b-73bb91fea7df

9. Adding the ability to set macros to use slots in the counter bar, no matter what item it's currently set to

image

10. Making sure that restoring from the previous counter bar XML will produce the exact same layout including placeholders in the new one

11. When resizing is disabled, adding and reordering entries should also be disabled. Basically the bar should be in a read-only and use-only mode.

https://github.com/user-attachments/assets/44cd3283-b795-4e2f-8751-006c518bf2b9

12. Fix highlight on use (closes #1216 )

https://github.com/user-attachments/assets/ab10a1e6-8e8e-45f5-8d6d-e44462482100

13. Counter Bar should snap to grid, so that it always only takes up the optimal space and does not waste any

https://github.com/user-attachments/assets/fcfda158-b225-4ad9-b2e6-f1abba094bc7

14. Testing and Polishing :-)

### DIFF
--- a/src/ClassicUO.Client/Game/Managers/MacroManager.cs
+++ b/src/ClassicUO.Client/Game/Managers/MacroManager.cs
@@ -1778,6 +1778,16 @@ namespace ClassicUO.Game.Managers
                 case MacroType.LookAtMouse:
                     // handle in gamesceneinput
                     break;
+
+                case MacroType.UseCounterBarSlot:
+                    {
+                        MacroObjectString objectString = (MacroObjectString)macro;
+                        string slotString = objectString.Text;
+
+                        CounterBarGump gump = UIManager.Gumps.First(gump => gump is CounterBarGump) as CounterBarGump;
+                        gump?.UseSlot(slotString);
+                        break;
+                    }
             }
 
 
@@ -2033,6 +2043,7 @@ namespace ClassicUO.Game.Managers
                 case MacroType.SetUpdateRange:
                 case MacroType.ModifyUpdateRange:
                 case MacroType.RazorMacro:
+                case MacroType.UseCounterBarSlot:
                     obj = new MacroObjectString(code, MacroSubType.MSC_NONE);
 
                     break;
@@ -2207,6 +2218,7 @@ namespace ClassicUO.Game.Managers
                 case MacroType.SetUpdateRange:
                 case MacroType.ModifyUpdateRange:
                 case MacroType.RazorMacro:
+                case MacroType.UseCounterBarSlot:
                     SubMenuType = 2;
 
                     break;
@@ -2323,7 +2335,8 @@ namespace ClassicUO.Game.Managers
         CloseInactiveHealthBars,
         CloseCorpses,
         UseObject,
-        LookAtMouse
+        LookAtMouse,
+        UseCounterBarSlot
     }
 
     internal enum MacroSubType
--- a/src/ClassicUO.Client/Game/Managers/MacroManager.cs
+++ b/src/ClassicUO.Client/Game/Managers/MacroManager.cs
@@ -1784,8 +1784,10 @@ namespace ClassicUO.Game.Managers
                         MacroObjectString objectString = (MacroObjectString)macro;
                         string slotString = objectString.Text;
 
-                        CounterBarGump gump = UIManager.Gumps.First(gump => gump is CounterBarGump) as CounterBarGump;
-                        gump?.UseSlot(slotString);
+                        if (UIManager.GetGump<CounterBarGump>() is { } bar)
+                        {
+                            bar.UseSlot(slotString);
+                        }
                         break;
                     }
             }
--- a/src/ClassicUO.Client/Game/Managers/UIManager.cs
+++ b/src/ClassicUO.Client/Game/Managers/UIManager.cs
@@ -523,7 +523,7 @@ namespace ClassicUO.Game.Managers
         /// </summary>
         /// <param name="allowedGumps">Gumps for which hovered controls should be returned</param>
         /// <returns>Read-only enumerable with the desired controls.</returns>
-        public static IEnumerable<Control> GetAllMouseOverControls(IEnumerable<Type> allowedGumps)
+        public static IEnumerable<Control> GetAllMouseOverControlsOfType<T>() where T : Control
         {
             List<Control> results = new List<Control>();
 
@@ -537,7 +537,7 @@ namespace ClassicUO.Game.Managers
             {
                 Control c = first.Value;
 
-                if (IsModalOpen && !c.IsModal || !c.IsVisible || !c.IsEnabled || !allowedGumps.Contains(c.GetType()))
+                if (IsModalOpen && !c.IsModal || !c.IsVisible || !c.IsEnabled || c is not T)
                 {
                     continue;
                 }
--- a/src/ClassicUO.Client/Game/Managers/UIManager.cs
+++ b/src/ClassicUO.Client/Game/Managers/UIManager.cs
@@ -1,7 +1,11 @@
 // SPDX-License-Identifier: BSD-2-Clause
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Collections.Specialized;
+using System.Linq;
 using ClassicUO.Configuration;
 using ClassicUO.Game.Scenes;
 using ClassicUO.Game.UI.Controls;
@@ -510,6 +514,46 @@ namespace ClassicUO.Game.Managers
 
             return null;
         }
+
+        /// <summary>
+        ///     Returns all controls which are part of a Gump from the provided list,
+        ///     as long as they are below the current mouse cursor position.
+        ///     
+        ///     Never returns null, but an empty enumerable instead.
+        /// </summary>
+        /// <param name="allowedGumps">Gumps for which hovered controls should be returned</param>
+        /// <returns>Read-only enumerable with the desired controls.</returns>
+        public static IEnumerable<Control> GetAllMouseOverControls(IEnumerable<Type> allowedGumps)
+        {
+            List<Control> results = new List<Control>();
+
+            Point position = Mouse.Position;
+
+            Control control = null;
+
+            IsModalOpen = IsModalControlOpen();
+
+            for (LinkedListNode<Gump> first = Gumps.First; first != null; first = first.Next)
+            {
+                Control c = first.Value;
+
+                if (IsModalOpen && !c.IsModal || !c.IsVisible || !c.IsEnabled || !allowedGumps.Contains(c.GetType()))
+                {
+                    continue;
+                }
+
+                c.HitTest(position, ref control);
+
+                if (control != null)
+                {
+                    results.Add(control);
+                }
+            }
+
+            return results.AsReadOnly();
+        }
+
+
 
         public static void MakeTopMostGump(Control control)
         {
--- a/src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
+++ b/src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
@@ -22,6 +22,11 @@ namespace ClassicUO.Game.UI.Gumps
     {
         private class CounterItem : Control
         {
+            private const int UPDATE_INTERVAL = 100;
+            private const int DRAG_OFFSET = 22;
+            private const float HIGHLIGHT_AMOUNT_CHANGED_DURATION = 5000f;
+            private const int HIGHLIGHT_AMOUNT_CHANGED_UP_HUE = 1165; //icelight
+            private const int HIGHLIGHT_AMOUNT_CHANGED_DOWN_HUE = 1166; //firelight
             private int _amount;
             private uint _lastChangeTime;
             private readonly ImageWithText _image;
@@ -174,16 +179,15 @@ namespace ClassicUO.Game.UI.Gumps
 
             protected override void OnDragBegin(int x, int y)
             {
-                if (!_gump.ShowBorder)
+                if (_gump.ReadOnly)
                 {
-                    // in read-only mode
                     return;
                 }
 
                 DraggableGump gump = new DraggableGump(_gump.World)
                 {
-                    X = Mouse.LClickPosition.X - 22,
-                    Y = Mouse.LClickPosition.Y - 22
+                    X = Mouse.LClickPosition.X - DRAG_OFFSET,
+                    Y = Mouse.LClickPosition.Y - DRAG_OFFSET
                 };
                 gump.Add(this);
                 X = 0;
@@ -198,16 +202,15 @@ namespace ClassicUO.Game.UI.Gumps
 
             protected override void OnDragEnd(int x, int y)
             {
-                if (!_gump.ShowBorder)
+                if (_gump.ReadOnly)
                 {
-                    // in read-only mode
                     return;
                 }
 
                 Control oldParent = this.Parent;
                 if (oldParent is DraggableGump)
                 {
-                    FinalizeDragDrop(oldParent, UIManager.GetAllMouseOverControls([typeof(CounterBarGump)]));
+                    FinalizeDragDrop(oldParent, UIManager.GetAllMouseOverControlsOfType<CounterBarGump>());
                 }
 
                 base.OnDragEnd(x, y);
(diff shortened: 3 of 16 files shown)


Lichtblitz:
@jabinb I believe one or two more programming sessions and I will have completed the features in the pipeline. How would you prefer the PR to be handed over to you? Are the commits in the current form okay, should I squash them into one, etc.?

jabinb:
> @jabinb I believe one or two more programming sessions and I will have completed the features in the pipeline. How would you prefer the PR to be handed over to you? Are the commits in the current form okay, should I squash them into one, etc.?

The current form should be okay for review. If there's any specific areas you want me to comment on feel free to point them out.

One area to focus on for testing is the upgrade path from the previous version, users will get upset if their counter bars get bugged or trashed in the update.

Lichtblitz:
> One area to focus on for testing is the upgrade path from the previous version, users will get upset if their counter bars get bugged or trashed in the update.

I fully agree. It's point 10 on the feature list.

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
```diff
@@ -0,0 +1,539 @@
...
+                            {
+                                GetAmount(item, Graphic, Hue, ref _amount);
+                            }
+                        }
+
+                        if (ProfileManager.CurrentProfile.CounterBarHighlightOnUse)
+                        {
+                            if (_amount > previousAmount)
+                            {
+                                _background.Hue = 1165; //icelight
```
Lichtblitz: I'm a bit unhappy with using ingame Hues that may look different on different shards. However, the AlphaBlendControl only takes Hues and not colors. Is there a better way?

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
```diff
@@ -0,0 +1,539 @@
...
+                                _lastChangeTime = Time.Ticks;
+                            }
+                            else if (_amount < previousAmount)
+                            {
+                                _background.Hue = 1166; //firelight
+                                _lastChangeTime = Time.Ticks;
+                            }
+
+                            _background.Alpha = Math.Min(1, 1 - (Time.Ticks - _lastChangeTime) / 5000f);
```
Lichtblitz: Yeah... very clunky way of adding an animation. Is there a better way?

Lichtblitz:
The PR is now ready for review. I'm happy to take suggestions and review findings, especially regarding the code style in case I'm too far off. I have also added two comments to the code (see above) where I'm still a bit unhappy but didn't find a better solution.

Apart from that all features as outlined in the PR above have been tested. Just note that I have created the videos as I have finished one feature after the other, so the earlier videos don't show features that were implemented further down the line, for example the snap to grid feature when resizing.

Thanks for offering your review, @jabinb, it's greatly appreciated :-)

src/ClassicUO.Client/Game/Managers/MacroManager.cs
```diff
@@ -1778,6 +1778,16 @@ private int Process(MacroObject macro)
                 case MacroType.LookAtMouse:
                     // handle in gamesceneinput
                     break;
+
+                case MacroType.UseCounterBarSlot:
+                    {
+                        MacroObjectString objectString = (MacroObjectString)macro;
+                        string slotString = objectString.Text;
+
+                        CounterBarGump gump = UIManager.Gumps.First(gump => gump is CounterBarGump) as CounterBarGump;
```
jabinb: Can use the `UIManager.GetGump<>` helper here:

```cs
if (UIManager.GetGump<CounterBarGump>() is { } bar)
    bar.UseSlot(slotString);
```

src/ClassicUO.Client/Game/Managers/UIManager.cs
```diff
@@ -511,6 +515,46 @@ private static Control GetMouseOverControl(Point position)
...
 
+        /// <summary>
+        ///     Returns all controls which are part of a Gump from the provided list,
+        ///     as long as they are below the current mouse cursor position.
+        ///     
+        ///     Never returns null, but an empty enumerable instead.
+        /// </summary>
+        /// <param name="allowedGumps">Gumps for which hovered controls should be returned</param>
+        /// <returns>Read-only enumerable with the desired controls.</returns>
+        public static IEnumerable<Control> GetAllMouseOverControls(IEnumerable<Type> allowedGumps)
```
jabinb: If I understand the usage correctly `UIManager.MouseOverControl` might be a better option for this, as we can traverse up the parents to find if its hovered rather than hit-testing all gumps.

Also since ClassicUO runs via nativeaot for the production builds we generally avoid using any dynamic typing (i.e. passing `Type`) due to reflection being unsupported. If you keep the method it's best to use generics and be more specific, i.e. `GetAllMouseOverControlsOfType<T>() where T : Control`

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
```diff
@@ -0,0 +1,539 @@
...
+                {
+                    FinalizeDragDrop(oldParent, UIManager.GetAllMouseOverControls([typeof(CounterBarGump)]));
+                }
+
+                base.OnDragEnd(x, y);
+            }
+
+            private void FinalizeDragDrop(Control oldParent, IEnumerable<Control> hoveredControls)
+            {
+                if (hoveredControls.Any())
```
jabinb: I recommend "short-circuiting" where possible to reduce nesting as it will help others read the code easier.

This should be okay? I don't see `oldParent` referenced in the nested if
```cs
if(!hoveredControls.Any()) {
  oldParent.Dispose();
  return;
}
```

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
```diff
@@ -0,0 +1,539 @@
...
+                            
+                    }
+
+                }
+                oldParent.Dispose();
+            }
+
+            protected override void OnMouseUp(int x, int y, MouseButtonType button)
+            {
+                if (button == MouseButtonType.Left)
```
jabinb: Would be nicer using a switch statement with pattern matching here

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
jabinb: I think using the game hues is fine for that reason (they can change it), but I recommend using the base colours like `0x44` (green) and `x20` (red) as they're typically left alone unless the shard has some wacky styling.

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
```diff
@@ -0,0 +1,539 @@
...
+                                return;
+                            }
+                        }
+
+                        _image.SetAmount(prefix + displayAmount.ToString());
+                    }
+                }
+            }
+
+            private static void GetAmount(Item parent, ushort graphic, ushort? hue, ref int amount)
```
jabinb: I know this is from the previous gump but could be useful to hoist this as a utility function in `Entity.cs` or `PlayerMobile.cs` as `GetTotalAmount` while we're here

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
```diff
@@ -0,0 +1,539 @@
...
+
+                return true;
+            }
+
+            private int CalculateDisplayAmount()
+            {
+                return _amount - CompareTo;
+            }
+
+            public override void Update()
```
jabinb, line 314: We're getting a bit complex here, maybe less cognitive load if we broke it down into smaller pieces?

e.g. could break it into a few methods like
`UpdateAmount()`
`UpdateOnUseAnimation()`
`UpdateLayout()`

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
```diff
@@ -0,0 +1,539 @@
...
+
+            protected override void OnMouseExit(int x, int y)
+            {
+                base.OnMouseExit(x, y);
+                ClearTooltip();
+            }
+
+            protected override void OnDragBegin(int x, int y)
+            {
+                if (!_gump.ShowBorder)
```
jabinb: Lets just rename it to `ReadOnly` as it looks like from the usages we use it more for checking that than knowing whether to draw the border.

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
```diff
@@ -0,0 +1,539 @@
...
+using Microsoft.Xna.Framework.Graphics;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ClassicUO.Game.UI.Gumps
+{
+    internal partial class CounterBarGump
+    {
+        private class CounterItem : Control
```
jabinb, line 23: Lets hoist some of the magic numbers as constants so its easier to tweak later

```cs
private const int UPDATE_INTERVAL = 100;
private const int DRAG_OFFSET = 22;
```

jabinb (commented):
Overall looks great, thanks for spending the effort adding the videos too.
I played around with it locally, seems to work really well nice work :+1:

Some feature changes:
- `Highlight On Use` is a bit of a misnomer as it's really based on the amount changing, so maybe `Highlight On Consume` might be more explicit? 
  `On Use` made me think of the item being double-clicked. Alternatively you could make it also highlight on the item being used (i.e. macro or double-click) and keep the `On Use` nomenclature. 
- Small UX change could be to not show `1` when there's a singular item? For instance adding your spellbook would always show the amount when its not useful.

Left some comments/feedback for improvements but feel free to disagree.
\@ me again when you're set to merge

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
Lichtblitz: I'll double-check later but I believe I have tried it and they were very dark. Looks like the background uses the first color in the hue. That's why I settled for `firelight` and `icelight` because they start with a very light color already and thus _pop_ more.

Lichtblitz: https://github.com/user-attachments/assets/0a42a7c3-2648-440e-a84c-09e469f0111b

This is too dark for my taste. What do you think, @jabinb ?

src/ClassicUO.Client/Game/Managers/MacroManager.cs
Lichtblitz: Thanks, implementation changed with the next commit.

src/ClassicUO.Client/Game/Managers/UIManager.cs
Lichtblitz: The method `GetAllMouseOverControls` was implemented the same way as `GetMouseOverControl`. The latter isn't helpful in this case because we are dragging a gump. So `GetMouseOverControl` and in turn the cached `MouseOverControl` will always return the dragged gump. No the one we are hovering over **with** the dragged gump.

I will happily follow your suggestion of implementing generics here, which is part of the next commit.

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
Lichtblitz: Fixed in upcoming commit

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
Lichtblitz: Will be fixed in upcoming commit, thanks

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
Lichtblitz: Good idea!

I've split up the code into `Item` for searching in Containers and the items itself. `Entity` does not have amounts, so it doesn't make sense at that level.

And into `PlayerMobile` for finding the appropriate containers on the player and calling the method on `Item`.

Will be part of the upcoming commit.

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
Lichtblitz: Method refactored in upcoming commit

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
Lichtblitz: Makes perfect sense, refactoring in upcoming commit

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
Lichtblitz: Refactored all magic numbers I found into constants (where reasonable). Part of the upcoming commit.

Lichtblitz:
- Singular items don't show an amount now.
- "Highlight On Use" renamed to "Highlight On Change"
image


Commit is pushed and the PR is ready for round two, @jabinb :-)

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
jabinb: I agree its not ideal but I think safer than using more commonly edited hues which will cause some very strange behaviour. Sadly there's currently no support to use a RGB value which would solve it too.

src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.CounterItem.cs
jabinb: FWIW this seems correct, animating `Update()` seems to be the common way from what I understand

jabinb (approved):
Looks good :+1: 

I think do a final test that users local settings safely upgrade then lets merge.

Lichtblitz:
Sure thing. :-)

1. Login with the current ClassicUO client. This is what I see:

image

2. Copy the Profiles folder over to my Visual Studio workspace; staring the head of the branch and this is what I see:

image

Looks identical to me :-)

Lichtblitz:
btw. thanks for you valuable code review @jabinb 
I enjoyed the process and was happy to receive such a thorough feedback :-)

PR #1794, opened 2025-07-30
Player Feedback: Additional features/improvements for the counter bar

Lichtblitz:
1. If a dagger was moved into the paperdoll, it disappeared from the counter bar. Now it is still counted and can be used through the macro or the double click function. This applies to all worn/carried items.
2. Automatically adds a helpful text to get people started with the counter bar while it is empty

https://github.com/user-attachments/assets/55e4e389-4b9a-42f9-8a5c-0c275656cf6f

### DIFF
--- a/src/ClassicUO.Bootstrap/src/Program.cs
+++ b/src/ClassicUO.Bootstrap/src/Program.cs
@@ -39,7 +39,7 @@ AppDomain.CurrentDomain.UnhandledException += (s, e) =>
     if (!Directory.Exists(path))
         Directory.CreateDirectory(path);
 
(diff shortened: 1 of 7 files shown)


Lichtblitz:
@jabinb Since you're already familiar with the counter bar code, would you do the code review for this as well?

My tests were successful and I won't be adding more features this round, so it's ready for code review.

Please wait before merging it, though. I'm waiting on another player, that they were unable to break anything even when trying :-D

Lichtblitz:
This PR is now ready for review and merge @jabinb  🙂

jabinb (approved):
Looks good, thanks for the follow up!

PR #1796, opened 2025-08-11
SystemChatControl extended to 500 characters. It will send the first …

Lichtblitz:
…100 characters, breaking at the last possible space character and keep the mode and remaining text to be sent with the next press of the enter button until the text box is exhausted

Thirty years! Thirty years players were annoyed by the character limit of only 100 characters. Especially in the heat of role playing, one may not notice that the limit was exhausted and type a few words into the void, never to be seen again. Today we enter the 21st century. The text box now takes up to 500 characters of text. On each successive press of the `enter` key it will send the first 100 characters to the server (breaking at the latest possible space character), keeping the remaining text and the text mode. Only after all the text is exhausted, will the text box reset its mode.

### DIFF
--- a/src/ClassicUO.Client/Game/Scenes/GameSceneInputHandler.cs
+++ b/src/ClassicUO.Client/Game/Scenes/GameSceneInputHandler.cs
@@ -1223,19 +1223,20 @@ namespace ClassicUO.Game.Scenes
                     {
                         if (ProfileManager.CurrentProfile.ActivateChatAfterEnter)
                         {
-                            UIManager.SystemChat.Mode = ChatMode.Default;
-
-                            if (
-                                !(
-                                    Keyboard.Shift
-                                    && ProfileManager.CurrentProfile.ActivateChatShiftEnterSupport
-                                )
-                            )
+                            if (UIManager.SystemChat.IsActive && Keyboard.Shift && ProfileManager.CurrentProfile.ActivateChatShiftEnterSupport)
                             {
-                                UIManager.SystemChat.ToggleChatVisibility();
+                                // Shift+Enter keeps the chat the way it is
+                                return;
                             }
-                        }
 
+                            if (UIManager.SystemChat.IsComposing)
+                            {
+                                // still text left to be sent, do nothing
+                                return;
+                            }
+                            UIManager.SystemChat.ToggleChatVisibility();
+                        }
+                        // everything else is handled in the system chat control
                         return;
                     }
 
--- a/src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.cs
+++ b/src/ClassicUO.Client/Game/UI/Gumps/CounterBarGump.cs
@@ -370,9 +370,6 @@ namespace ClassicUO.Game.UI.Gumps
                 ReadOnly = isReadOnly;
             }
 
-            Width = width;
-            Height = height;
-
             BuildGump();
 
             XmlElement controlsXml = xml["controls"];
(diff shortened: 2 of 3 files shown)


jabinb (changes requested):
Great feature idea I like it!

Couple things we can improve in the code:

- We have quite a bit of "business logic" for the feature spread across various functions which makes it a bit complex to maintain (or for others to modify). 
  Lets try to consolidate it. 
- We should try to stick to using the class methods wherever possible, 
  i.e. `TextBoxControl.SetText(value)` rather than `TextBoxControl.Text = value` etc
- The uses of `ref` makes things hard to follow due to its bi-directionality. There's no perf gain from using `ref` in this instance as we're alloc'ing new strings anyway (we'd need to use `ReadOnlySpan<char>` to avoid some of those). 
  How about we use the try/out pattern that's more C# idiomatic?

Here's my attempt (only briefly tested this), I've repeated the party/prompt checks but I think that's worth it to keep things in one place.

```cs
//...
public override void OnKeyboardReturn(int textID, string text)
{
    if (!IsActive && ProfileManager.CurrentProfile.ActivateChatAfterEnter || Mode != ChatMode.Default && string.IsNullOrEmpty(text))
    {
        TextBoxControl.ClearText();
        Mode = ChatMode.Default;
        return;
    }

    if (string.IsNullOrEmpty(text))
        return;

    var isPrompted = _gump.World.MessageManager.PromptData.Prompt != ConsolePrompt.None;

    if (TryHandleMessageMultipartSend(text, Mode, isPrompted, out var remainder))
    {
        TextBoxControl.SetText(remainder);
    }
    else
    {
        HandleMessageSend(text, Mode);
        TextBoxControl.ClearText();
    }
    
    if (TextBoxControl.Length == 0)
    {
        Mode = ChatMode.Default;
        DisposeChatModePrefix();
    }
}

private bool TryHandleMessageMultipartSend(string text, ChatMode mode, bool isPrompted, out string remainder)
{
    // Prompt response messages cannot be multiple parts
    if (isPrompted || text.Length <= MAX_MESSAGE_LENGTH)
    {
        remainder = text;
        return false;
    }
    
    int lastSpaceIndex = text.LastIndexOf(' ', MAX_MESSAGE_LENGTH);
    if (lastSpaceIndex < 0) lastSpaceIndex = MAX_MESSAGE_LENGTH;
    
    var message = text[..lastSpaceIndex];
    remainder = text[lastSpaceIndex..].TrimStart();
    
    HandleMessageSend(message, mode);
    
    // Preserve the party index we're messaging
    if (mode == ChatMode.Party && int.TryParse(text[0..2], out var partyCharIndex) && partyCharIndex is > 0 and < 11)
        remainder = $"{partyCharIndex} {remainder}";

    return true;
}

private void HandleMessageSend(string text, ChatMode sentMode)
//...
```

Lichtblitz:
Thanks for the feedback @jabinb 
I've mostly implemented it plus a missed edge case when the text box still has content but "activate chat on enter" is also active.
Also, I've fixed a bug in the counter bar when it is restored on login where the minimum width for the help text would be enforced even if items would be deserialized a moment later.

I'm usually very hesitant to refactor the existing code because of missing unit tests. I believe it would make sense to think about how to introduce them in the build process and project files, so I could start writing them moving forward. But that's a separate topic.

Lichtblitz:
Hey @jabinb 
Is there anything I could improve where you are still hesitant to make a decision? I'm happy to keep the conversation going even if it means more adjustments 🙂

jabinb (approved):
Sorry for the delay, I'll merge it now!

PR #1798, opened 2025-08-13
One of the reflection shims seem to have been forgotten in the bootstrapper

Lichtblitz:
ClassicAssist was broken for quite some time but degraded gracefully by sending the pathfind packet to the client instead. That packet does not allow setting the distance however and breaks when trying to pathfind to a location that is impassable (for example pathfinding to a blocking item, like a loom or a spinning wheel). This makes pathfinding from ClassicAssist in those cases impossible

### DIFF
--- a/src/ClassicUO.Bootstrap/src/Program.cs
+++ b/src/ClassicUO.Bootstrap/src/Program.cs
@@ -3,15 +3,15 @@ using CUO_API;
 using System;
 using System.Buffers;
 using System.Collections.Generic;
-using System.IO;
-using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
-using System.Threading;
-
 
 #if !DEBUG
+using System.IO;
+using System.Reflection;
+using System.Threading;
+
 AppDomain.CurrentDomain.UnhandledException += (s, e) =>
 {
     var dt = DateTime.Now;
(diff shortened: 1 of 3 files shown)


Lichtblitz:
@jabinb And another PR - this one should be pretty straight-forward to review
Thanks in advance :-)

src/ClassicUO.Bootstrap/src/Program.cs
```diff
@@ -419,6 +419,15 @@ public unsafe bool ReflectionAutowalking(sbyte walking)
         return toBool;
     }
 
+    public unsafe bool ReflectionWalkTo(int x, int y, int z, int distance)
+    {
+        var f = (4, x, y, z, distance);
+        var result = SendReflectionCmd((IntPtr)(&f));
+        var toBool = Unsafe.AsRef<bool>(result.ToPointer());
+        Console.WriteLine("bool: {0} [{1}]", toBool, result);
```
andreakarasho: maybe this is no needed anymore

andreakarasho (changes requested):

src/ClassicUO.Bootstrap/src/Program.cs
Lichtblitz: I'm happy to remove it. I just didn't want to break the pattern from the other methods that have the log output. 

Will fix it later today.

Lichtblitz: @andreakarasho fixed

andreakarasho (approved):
LGTM!
Thank you

PR #1800, opened 2025-08-19
The client now allows pathfinding to an object that is not a surface

Lichtblitz:
We are now catching up to assistants, which have always allowed this but froze the client for a second or two on an impossible pathfinding quest if the distance was left to 0

We still try and navigate to the target as usual but it the field is impassable and the distance is set to 0, we set it to 1, which is very, very simple but should catch the majority of cases that were broken previously.

https://github.com/user-attachments/assets/2a98588a-2a5e-4bed-9331-7c45d7e71c8b


andreakarasho (approved):
LGTM
Thank you!

PR #1801, opened 2025-08-22
The system chat will now break into multiple lines to show how the text will be sent as multiple parts

Lichtblitz:
also, Ctrl + Backspace now handles the caret not being at the end of the line correctly

https://github.com/user-attachments/assets/8063f2d1-0dcf-4603-8a9f-b3cc150d4a97

> This is a particularly long text. It will cover more than the allowed one hundred characters and
thus will be split up into separate lines at the last possible space character. With each press of
the ENTER key, one of the lines will be sent to the server. That way you're always in full control
over how quickly you wish to send the messages or how much time you would like to give others to
read first!

### DIFF
--- a/src/ClassicUO.Client/Game/UI/Controls/StbTextBox.cs
+++ b/src/ClassicUO.Client/Game/UI/Controls/StbTextBox.cs
@@ -22,6 +22,19 @@ namespace ClassicUO.Game.UI.Controls
 
         private readonly int _maxCharCount = -1;
 
+        public class BeforeTextChangedEventArgs : EventArgs
+        {
+            public BeforeTextChangedEventArgs(string previousText, string newText, int newCaretIndex)
+            {
+                PreviousText = previousText;
+                NewText = newText;
+                NewCaretIndex = newCaretIndex;
+
+            }
+            public string PreviousText { get; }
+            public string NewText { get; set; }
+            public int NewCaretIndex { get; set; }
+        }
 
         public StbTextBox
         (
@@ -133,6 +146,7 @@ namespace ClassicUO.Game.UI.Controls
 
         public bool AllowTAB { get; set; }
         public bool NoSelection { get; set; }
+        public bool PassEnterToParent { get; set; } = false;
 
         public int CaretIndex
         {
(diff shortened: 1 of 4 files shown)


src/ClassicUO.Client/Game/UI/Controls/StbTextBox.cs
```diff
@@ -133,6 +146,7 @@ public byte Font
 
         public bool AllowTAB { get; set; }
         public bool NoSelection { get; set; }
+        public bool PassEnterToParent { get; set; } = false;
```
andreakarasho: The `= false;` is not needed here because it's `false` by default

src/ClassicUO.Client/Game/UI/Gumps/SystemChatControl.cs
```diff
@@ -118,6 +122,59 @@ public SystemChatControl(WorldViewportGump gump, int x, int y, int w, int h)
...
+
+                relativeCursorIndex -= message.Length;
+                text = remainder;
+            }
+
+            result = AppendMultilinePart(result, message);
+            e.NewText = result;
+        }
+
+        private static string AppendMultilinePart(string result, string message)
```
andreakarasho, line 164: Probably here it's better to use a `StringBuilder result` instead of a raw string.

```csharp
private static void AppendMultilinePart(StringBuilder result, string message)
{
    if (result.Length > 0)
    {
        result.AppendLine();
    }
    else
    {
        result.Clear();
    }

    result.Append(message);
}
```

andreakarasho (changes requested):

src/ClassicUO.Client/Game/UI/Controls/StbTextBox.cs
Lichtblitz: Fair. I wanted to communicate that it was a conscious decision to have it turned off. Will fix.

src/ClassicUO.Client/Game/UI/Gumps/SystemChatControl.cs
Lichtblitz: I believe this is unnecessary complexity. We will have at most six lines with many seconds between operations and for such few concatenations the StringBilder will likely perform worse.

Lichtblitz:
Your finding is fixed in the second commit @andreakarasho

andreakarasho (approved):

andreakarasho:
Thank you!

PR #1802, opened 2025-08-22
Custom House Fixes

Voxpire:
Custom House Design Interface
- Roof tiles were not being properly parsed because column `EDent` was missing
- The columns `Extra` and `Piece` are actually a single column with a space separator; `Extra Piece`
- Minor fixes for potential out-of-range exceptions


andreakarasho (approved):
LGTM.
Thank you!!

PR #1803, opened 2025-08-23
Fixed a bug where the chat mode and the text box were overlapping

Lichtblitz:
The code path for resizing did not consider it and I was reusing it for the multiline functionality


andreakarasho (approved):

andreakarasho:
Thanks

PR #1804, opened 2025-08-23
Found and fixed a few more inconsistencies in the system chat control

Lichtblitz:
The text box now shows the correct color related to the chat mode and the corresponding label is positioned correctly in several different edge cases

I will do some more thorough testing before undrafting the PR. Maybe I find more inconsistencies.

They don't amount to a level of necessitating a hotfix.

Also, the half implemented `Prompt` chat mode is now fully implemented and has been refactored to make it more robust and a little less distributed between classes
image

### DIFF
--- a/src/ClassicUO.Client/Game/Data/PromptData.cs
+++ b/src/ClassicUO.Client/Game/Data/PromptData.cs
@@ -9,9 +9,9 @@ namespace ClassicUO.Game.Data
         Unicode
     }
 
-    internal struct PromptData
+    internal readonly struct PromptData(ConsolePrompt prompt, ulong data)
     {
-        public ConsolePrompt Prompt;
-        public ulong Data;
+        public readonly ConsolePrompt Prompt = prompt;
+        public readonly ulong Data = data;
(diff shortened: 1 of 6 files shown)


Lichtblitz:
@andreakarasho I can't find any more inconsistencies for now. I'll sleep over it another night to get a fresh mind, as it is (and has been) perfectly usable the way it is. Do you spot something else that is behaving weirdly in the system chat control?

Lichtblitz:
@andreakarasho I have refactored the server prompt handling as well. I was now unable to break the system chat and am happy with how consistently it now behaves, especially when resizing the main game gump.

I'm happy for additional feedback :-)

Lichtblitz:
Found another bug. When the client is being closed while a prompt is open, an Exception gets thrown. Will fix tomorrow.

Lichtblitz:
Event subscriptions are now properly cleaned up on destroy so we don't try and read an unset profile

andreakarasho (approved):
LGTM. Thank you !

PR #1805, opened 2025-08-26
Update world map when receiving terrain or static packets from Ultima…

DavideRei:
… Live

The only issue that I've found is the following. 
If the WorldMapGump is updated from UltimaLive, you logout and login again, without closing the client, the WorldMapGump loads the old map from the cached PNG image.
Instead if you close the client and launch it again, the WorldMapGump is reloaded correctly.

### DIFF
--- a/src/ClassicUO.Client/Game/Map/Chunk.cs
+++ b/src/ClassicUO.Client/Game/Map/Chunk.cs
@@ -111,7 +111,8 @@ namespace ClassicUO.Game.Map
                 }
             }
 
-            if (im.StaticAddress != 0 && im.StaticCount > 0)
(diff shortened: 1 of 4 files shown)


DavideRei:
I made some changes:
- null check on _mapTexture before updating it
- fixed the algorithm to update the block height colors. I tried to make it as close as possibile to the one on the world map gump load method but there you have all the map available, in the chunk file only the map block to update, so you can't read heights from other blocks around. Therefore there is no way to apply the right color to the last line of the block. On that line I left the original tile color.
- added the world map gump offset to the block coordinates
- clean the map cache on logout. This is required to load correctly the world map gump after an UL update and logout without closing the client. If there were no changes the map is still loaded from the .png image.
- when loading a new .png image for the current map, delete all the previous cached .png files for the same map number. I saw there is a command in the world map gump to delete the cache but I think is better this way, to avoid too many chached images.

PR #1806, opened 2025-08-30
Allow to set externaly IsUOPInstallation

kaczy93:
I need this for CentrED#. 
I dont call UOFileManager.Load() and instead I directly call Load on only these loaders that are needed.
Due to this I also manually set if client is UOP


andreakarasho:
That property should remains readonly to avoid errors while reading files.
A workaround would be to move these lines:
https://github.com/ClassicUO/ClassicUO/blob/main/src/ClassicUO.Assets/UOFileManager.cs#L142-L144
to the .ctor. But no sure if there is some side effect due to the overridemap class.

kaczy93:
I will create another PR as it have to be done different way

PR #1807, opened 2025-08-31
The character subfolder for profiles now contains the serial of the loaded character

Lichtblitz:
That way even if the name changes temporarily (incognito, morph, etc.) we can still load the correct profile. Legacy folders are automatically renamed when the modern folder does not exist, but a folder with only the character name is found.


Lichtblitz:
Is being fixed server-side, which makes much more sense: 

https://github.com/TrueUO/TrueUO/pull/1471
https://github.com/ServUO/ServUO/pull/5113
https://github.com/Sphereserver/Source-X/pull/1452

Lichtblitz:
ServUO is the final straggler to merge their PR. The others have done so and I feel good about closing this PR now, since we don't need a client-side solution.

PR #1811, opened 2025-09-03
More flexibility in UOFileManager 

kaczy93:
Housekeeping around UOFilesOverrideMap
UOFilesOverrideMap will be loaded outside of UOFileManager.
UOFileManager.IsUOPInstallation will be initialized in constructor

My only concern is if 2nd ctor of UOFilesOverrideMap shouldn't be in fact a factory method

### DIFF
--- a/src/ClassicUO.Assets/UOFileManager.cs
+++ b/src/ClassicUO.Assets/UOFileManager.cs
@@ -26,7 +26,6 @@ namespace ClassicUO.Assets
             BasePath = uoPath;
 
             _overrideMap = overrideMap;
-            _overrideMap.Load();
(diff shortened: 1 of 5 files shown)


andreakarasho:
> My only concern is if 2nd ctor of UOFilesOverrideMap shouldn't be in fact a factory method

That's right. Probably just store the `string overrideFile` .ctor parameter in a member variable. Then move the factory logic to a `Load` method (like it was in precedence). At the end call that `Load` function before loading the files

kaczy93:
Alright, I finally found some time to fix this.
I also didnt like that now overrides map was loaded in the file manager ctor, so I changed it to Load explicitly before we create file manager.
I think now it is clean and it should allow me to do what I need :)

andreakarasho (approved):
LGTM. Thank you!

PR #1813, opened 2025-09-07
Add missing Verdata.Load call

### LINKED ISSUE #1787: ClassicUO in its latest versions has become incompatible with older versions of Ultima Online
Describe the bug
ClassicUO in its latest versions has become incompatible with older versions of Ultima Online (1.26.4b), preventing successful login or gameplay on shards using this version.
To Reproduce
Launch ClassicUO with force verdata true
i can share my ultima online file game to test

Screenshots

Image

Image

darcymiranda:
I'm trying to get an old shard to work that uses `verdata.mul` and noticed that `Verdata.Load()` is never actually called. Without this, the file is never processed, even with `use_verdata=true`. I scoured the commit history but couldn't find any reason why this is missing. 

Feel free to close if there's a reason, but after this change, it works great for the old shard I'm working with.

Might also fix https://github.com/ClassicUO/ClassicUO/issues/1787.

### DIFF
--- a/src/ClassicUO.Assets/UOFileManager.cs
+++ b/src/ClassicUO.Assets/UOFileManager.cs
@@ -164,6 +164,7 @@ namespace ClassicUO.Assets
             MultiMaps.Load();
             TileArt.Load();
             StringDictionary.Load();
+            Verdata.Load();
 
             ReadArtDefFile();
 


andreakarasho (approved):
Probably a unintended deletion.
Thank you!

PR #1818, opened 2025-09-23
Updated FNA to 25.9 and SDL to v3

Lichtblitz:
At a quick glance everything seems to be working as before.

Except for adhering the DPI settings. Discussion here:
https://discord.com/channels/458277173208547350/674900239421014016/1419739719239073852


Lichtblitz:
Improving the existing code will take more time. So I'm setting this for "Ready for Review" to get it merged. This switches the framework to the new version without adding new features or fixing pre-existing bugs/issues.

Lichtblitz:
Please test this locally to make sure that I haven't missed any submodule changes in the PR. Those need to be updated as well after checking out.

andreakarasho:
My only concern is about that DPI setting.

Lichtblitz:
The behavior has not changed with the switch to version 3. DPI settings were ignored before and are ignored after. Or rather, there is no implementation that took them into consideration, since it's up to the application to handle it, instead of an automatism.

We have no pressure to get this merged. I simply didn't want this to become a long-lived branch that requires continued rebasing with the potential for merge conflicts while I figure out ways to rework the rendering stack.

andreakarasho (approved):
LGTM, thank you

PR #1819, opened 2025-09-26
Sound def issue

dagid4:
Sound.def is not respected when Sound.mul returns a length of -1. As a result, some sounds don’t work (for example, the Eagle mob and the Invisible spell, tested in version 4.0.4b).

This simple fix changes the skip condition from “not equal to zero” to “greater than zero.”


andreakarasho (approved):
LGTM. Thank you!

PR #1820, opened 2025-09-30
Missing clilocs do not return `null` any more

Lichtblitz:
Instead they return a more helpful and less crash-prone text of `"MegaCliloc: missing {clilocNum} [~1_val~] [~2_val~]"`

This can help shard owners find and fix missing translations instead of hiding them or crashing the client.

### DIFF
--- a/src/ClassicUO.Assets/ClilocLoader.cs
+++ b/src/ClassicUO.Assets/ClilocLoader.cs
@@ -5,14 +5,14 @@ using ClassicUO.Utility;
 using ClassicUO.Utility.Logging;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace ClassicUO.Assets
 {
     public sealed class ClilocLoader : UOFileLoader
     {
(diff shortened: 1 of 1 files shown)


src/ClassicUO.Assets/ClilocLoader.cs
```diff
@@ -96,51 +96,32 @@ public override void ClearResources()
...
-        {
-            string s = GetString(number);
-
-            if (string.IsNullOrEmpty(s))
-            {
-                s = replace;
-            }
-
-            return s;
+            return text ?? fallback ?? string.Format(MISSING_CLILOC_TEXT, clilocNum);
```
andreakarasho, line 104: Should not check if `fallback` is `NullOrEmpty`?

Lichtblitz: The default value is null. So if fallback is an empty string, then a dev must have specifically passed it as a parameter. 

So I'd say that they want it to be empty. On the other hand I don't feel strongly about it. So if you prefer the check for empty as well, I'll change it.

andreakarasho: Makes sense

andreakarasho (approved):
LGTM. Thank you

PR #1828, opened 2025-11-01
PointClamp looks better for pixel art compared to linear or anisotropic

Lichtblitz:


andreakarasho:
Nearest-neighbor upscaling means each original pixel becomes a 1.25×1.25 block, so it doesnt fit into a whole pixel. That's why `PointClamp` causes cripsy and jagging artifacts.
On Windows, when changing the DPI settings of an application, the OS applies the bilinear filter to compensate the graphics artifacts. https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows#dpi-unaware
So that `Linear` is doing the same of Windows.
We should provide a way to disable (or swap) the Filter from Linear to Pointer when the DpiScale prop is `1.0`.

PR #1835, opened 2025-11-10
Add new ability mapping for Spiked Whip

tr1age:
Fixes Whirlwind attack for whip.

### DIFF
--- a/src/ClassicUO.Client/Game/Data/Ability.cs
+++ b/src/ClassicUO.Client/Game/Data/Ability.cs
@@ -305,11 +305,9 @@ namespace ClassicUO.Game.Data
             { 0x48D1, new (0x48D1, Ability.Feint, Ability.DoubleStrike) },
             { 0x48D0, new (0x48D0, Ability.Feint, Ability.DoubleStrike) },             // Gargish Daisho
             { 0xA289, new (0xA289, Ability.ConcussionBlow, Ability.WhirlwindAttack) }, // Barbed Whip
-            { 0xA291, new (0xA291, Ability.ConcussionBlow, Ability.WhirlwindAttack) }, // Barbed Whip (second ID)
             { 0xA28A, new (0xA28A, Ability.ArmorPierce, Ability.WhirlwindAttack) },    // Spiked Whip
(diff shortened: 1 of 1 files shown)


src/ClassicUO.Client/Game/Data/Ability.cs
```diff
@@ -306,6 +306,7 @@ internal static class AbilityData
             { 0x48D0, new (0x48D0, Ability.Feint, Ability.DoubleStrike) },             // Gargish Daisho
             { 0xA289, new (0xA289, Ability.ConcussionBlow, Ability.WhirlwindAttack) }, // Barbed Whip
             { 0xA28A, new (0xA28A, Ability.ArmorPierce, Ability.WhirlwindAttack) },    // Spiked Whip
+            { 0xA292, new (0xA292, Ability.ArmorPierce, Ability.WhirlwindAttack) },    // Spiked Whip
```
Lichtblitz: Image

There is also a Barbed and Bladed Whip at 0xA291 and 0xA293. If you know the abilities, it would be great to add those as well.

tr1age: They are the same so added them as secondary IDs Although these are just FLIPPED ids. So, I don't think this actually matters. I am removing it.

tr1age (commented):
This doesn't actually matter since these are JUST the flipped ID for the items.

src/ClassicUO.Client/Game/Data/Ability.cs
Lichtblitz, line 143: Broadswords for example have their flipped counterparts added as well. It makes sense since it's the server, not the client that knows about flipped IDs. From the client's perspective those graphics could as well be different items.

tr1age: Ahh didn't know they did it that way, Adding.

tr1age:
Ok added (Flipped ID) adhere to the way OSI does properties.

Lichtblitz (approved):
LGTM

andreakarasho (approved):

PR #1850, opened 2025-11-26
prevent closing containers when changing facets

tr1age:
There are two behavior changes:

```
if (item != null && item.Container != containerSerial && (container.Graphic != 0x2006 || item.Layer == Layer.Invalid)) {
    world.RemoveItem(item, true);
}
```

Previously this would remove the item from the world any time it existed, which could blow away items when containers are being reused or when changing facets. Now it only removes the item if it currently belongs to a different container. So if the item is already associated with this container, it is left alone.

```
if (item.Container != containerSerial)
{
    world.RemoveItemFromContainer(item);
    item.Container = containerSerial;
}

```

Previously you always called RemoveItemFromContainer and then reassigned the container. Now you only do that if the item is actually moving to a different container. If it is already in containerSerial, you skip the remove and readd cycle, which helps avoid unnecessary container state changes and the side effects that go with them, such as open gumps closing when you cross facets.

### DIFF
--- a/src/ClassicUO.Client/Network/PacketHandlers.cs
+++ b/src/ClassicUO.Client/Network/PacketHandlers.cs
@@ -6034,38 +6034,39 @@ namespace ClassicUO.Network
                     Console.WriteLine("ADD ITEM TO CONTAINER -- CLEAR HOLD");
                     Client.Game.UO.GameCursor.ItemHold.Clear();
                 }
-
+        
                 //else if (ItemHold.Graphic == graphic && ItemHold.Amount == amount &&
                 //         ItemHold.Container == containerSerial)
                 //{
-                //    ItemHold.Enabled = false;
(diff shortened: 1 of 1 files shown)


andreakarasho:
This might create phantom items because the server (this needs to be verified) will emit a deletion packet for the container once you change  facet

Lichtblitz:
After review and some testing I believe this change to be harmless @andreakarasho 

Old behavior:
1. ClassicUO receives packets to add an item to a container
2. The item is removed from the current container if it exists (if it is a container, this will close them)
3. The item is added to the new container

New behavior:
1. ClassicUO receives packets to add an item to a container
2. The item is removed from the current container if it exists **and is not already inside that container**
3. The item is added to the new container **if it is not already inside that container**

The end result is exactly the same before and after.

PR #1852, opened 2025-12-13
Proposal: Fix Weather Particle Following Player Bug & Add Immersive Weather Enhancements

birdinforest:
## Description

This PR addresses a weather particle positioning bug and proposals a configurable weather rendering enhancements (configirations haven't been implemented)

[Demo Video on Youtube](https://youtu.be/wIwxsSMqvAc?si=PUjoajL4rxuJVSMD)

---

## Scope & PR Structure Note

I understand that best practice is to separate a core bug fix and visual/immersive enhancements into two independent PRs. In this case, the immersive weather changes are included only as a proposal for discussion and review, not as a requirement for merging.

If maintainers prefer, I am very happy to:

+ Extract the weather particle positioning fix into a separate, bug-fix-only PR

+ Submit the immersive weather enhancements as a follow-up PR, or keep them purely conceptual until scope is confirmed

Please feel free to advise on the preferred approach.

---

### Issue

Rain and snow particles were incorrectly following the player around. It looked like a raining/snowing buble was stuck to the player instead of being part of the actual game world. 

### Solution

Fixed it by:
- Implementing a proper world-space coordinate system using isometric world coordinates
- Adding viewport offset calculations to convert between world space and viewport space
- Making sure particles maintain their position in the game world regardless of player movement
- Fixing particle respawn logic to use world coordinates instead of viewport-relative positions

### Impact

✅ Weather particles now behave correctly and exist independently in world space  

---

#### Three-Layer Depth System

It adds a depth system that makes rain look way more realistic:
- **Background Layer** (0-30% viewport height): Distant particles that fade out early - gives that atmospheric depth feel
- **Middle Layer** (31-70% viewport height): Mid-range particles that fade out in middle range of the screen
- **Foreground Layer** (71-100% viewport height): Close particles that have a 50% chance to never fade - for the realistic raining view on screen perspective

#### Dynamic Splash Effects

Rain now creates water splashes when it hits the ground.

Splash sizing is also depth-aware - distant splashes are smaller (3 droplets), middle ones are medium (6 droplets), and close ones are larger (10 droplets)

- ~~[ ] I might remove this feature later if it doesn't add much value (To be honestly the splash sizing is not really noticeable in practice)~~
- [x] Avoid splash when rain drops on building / floor tiles.

#### Ripple Effects on Water

When rain hits water tiles, it creates ripples on the surface. 

#### Density-Based Rendering Styles

The visual style adapts to how heavy the rain is:
- **Small Dots** (0-10): Light drizzle
- **Large Dots** (11-30): Moderate rain
- **Short Lines** (31-50): Heavy rain streaks
- **Long Bolts** (50-70): Full-on storm conditions

Speed also scales with density

#### Density-Based Raining Sound Effects
0-50: 0x011
50-70: 0x010

#### UI Toggles:
Options -> Video -> Misc -> Weather effects (Splash & Ripples)
Opitons -> Sounds -> Play Rain Sound

#### Three-Layer Depth System

Same depth layer system as rain, so everything stays consistent. Particles fade naturally as they reach ground level, which creates a nice snow accumulation visual effect.

#### Vertical Fall with Wind Influence

Snow now falls down. Wind adds horizontal drift with a 0.8x strength multiplier. Removed the old oscillating angle system that made snow drift around weirdly - now it stays mostly vertical with just a subtle wind influence. 

#### Density-Based Rendering

Snow scales with intensity just like rain:
- **Light Flakes** (0-17): Small, slow particles
- **Medium Flakes** (18-35): Moderate size and speed
- **Heavy Flakes** (36-52): Larger, faster particles
- **Blizzard** (53-70): Biggest, fastest particles

Both size and speed multiplier go from 1.0x to 2.0x based on density.

#### Semi-Transparent Soft Edges

Each snowflake gets rendered with a four-layer gradient to create soft, circular edges. The alpha goes from 25% (transparent outer edge) → 45% → 70% → 100% (opaque core). This makes the edges blend naturally into the background and gives it a much more polished look. 

---

### Immersive Enhancement Scope

**Question for Review**: These enhancements look great and really improve immersion, but I’m not sure whether they are within the scope of the ClassicUO project. If they are in scope, should they be placed behind configuration flags in the Settings? I think they should probably be optional, especially for web client users. Another concern is that some users may be annoyed by the particle effects and therefore prefer to disable them.

### Performance Impact

Here's what to expect:
- Additional rendering layers per particle (3-4 draw calls per snowflake for the soft edge effect - can be toggled off if necessary)
- Splash and ripple effect systems add some overhead (though particle counts limited in 256)
- Depth-based calculations are pretty minimal - shouldn't be a big deal for modern CPU
- Modern desktop hardware should handle this fine, but **I have a bit concerned about web client performance**

~~### Fallback Behavior~~

~~When enhanced effects are disabled, it could apply simplified rendering:~~
~~- Single-layer particles (no depth system)~~
~~- No splash/ripple effects~~
~~- Solid snow particles (no gradient)~~

### DIFF
--- a/src/ClassicUO.Client/Game/Weather.cs
+++ b/src/ClassicUO.Client/Game/Weather.cs
@@ -9,6 +9,8 @@ using Microsoft.Xna.Framework;
 using System;
 using MathHelper = Microsoft.Xna.Framework.MathHelper;
 using ClassicUO.Game.Map;
+using ClassicUO.IO.Audio;
+using ClassicUO.Configuration;
 
 namespace ClassicUO.Game
 {
@@ -29,10 +31,10 @@ namespace ClassicUO.Game
         private const float SIMULATION_TIME = 37.0f;
 
         // Density thresholds for rain visual styles
-        private const int DENSITY_SMALL_DOTS = 17;
-        private const int DENSITY_LARGE_DOTS = 35;
-        private const int DENSITY_SHORT_LINES = 52;
-        // Above 52 = long bolts
+        private const int DENSITY_SMALL_DOTS = 10;
+        private const int DENSITY_LARGE_DOTS = 30;
+        private const int DENSITY_SHORT_LINES = 50;
+        // Above 50 = long bolts
 
         private const float BASE_RAIN_SPEED_Y = 15.0f;
         private const float BASE_RAIN_SPEED_X = -2.5f;
@@ -54,27 +56,32 @@ namespace ClassicUO.Game
         // Foreground:  81-100% (19% fade zone) - appears closest, reaches near bottom
 
         // --- Splash Enable/Disable by Density ---
-        private const bool SPLASH_ENABLED_SMALL_DOTS = false;
-        // Enable splash effects for small dots (density 0-17)
+        private const bool SPLASH_ENABLED_SMALL_DOTS = true;
+        // Enable splash effects for small dots (density 0-10)
         // - false: No splashes for light drizzle (subtle, minimal effect)
         // - true: Splashes even at lowest density
-        
+
         private const bool SPLASH_ENABLED_LARGE_DOTS = true;
-        // Enable splash effects for large dots (density 18-35)
+        // Enable splash effects for large dots (density 11-30)
         // - false: No splashes for moderate rain
         // - true: Splashes appear at moderate density ✓ RECOMMENDED
-        
+
         private const bool SPLASH_ENABLED_SHORT_LINES = true;
-        // Enable splash effects for short lines (density 36-52)
+        // Enable splash effects for short lines (density 30-50)
         // - Always recommended to be true for heavy rain
-        
+
         private const bool SPLASH_ENABLED_LONG_BOLTS = true;
-        // Enable splash effects for long bolts (density 53-70)
+        // Enable splash effects for long bolts (density 50-70)
         // - Always recommended to be true for storm conditions
+
+        private const float RAIN_VOLUME_MULTIPLIER = 0.65f;
+        private const int MINOR_RAIN_SOUND_ID = 0x011;
+        private const int HEAVY_RAIN_SOUND_ID = 0x010;
 
         private readonly WeatherEffect[] _effects = new WeatherEffect[byte.MaxValue];
         private uint _timer, _windTimer, _lastTick;
         private readonly World _world;
+        private UOSound _currentRainSound;
 
         public Weather(World world)
         {
@@ -107,6 +114,7 @@ namespace ClassicUO.Game
             Wind = 0;
             _windTimer = _timer = 0;
             CurrentWeather = null;
+            StopRainSound();
         }
 
         public void Generate(WeatherType type, byte count, byte temp)
@@ -152,6 +160,9 @@ namespace ClassicUO.Game
 
                         CurrentWeather = type;
                     }
+
+                    // Start looping rain sound
+                    UpdateRainSound();
 
                     break;
 
@@ -439,6 +450,122 @@ namespace ClassicUO.Game
            PlaySound(RandomHelper.RandomList(0x028, 0x206));
         }
 
+        private void PlayMinorRain()
+        {
+            PlayRainSoundLoop(MINOR_RAIN_SOUND_ID);
+        }
+
+        private void PlayHeavyRain()
+        {
+            PlayRainSoundLoop(HEAVY_RAIN_SOUND_ID);
+        }
+
+        private void PlayRainSoundLoop(int soundId)
+        {
+            if (!_world.InGame || _world.Player == null)
+            {
+                return;
+            }
+
+            StopRainSound();
+
+            _currentRainSound = (UOSound)Client.Game.UO.Sounds.GetSound(soundId);
+            if (_currentRainSound != null)
+            {
+                Profile currentProfile = ProfileManager.CurrentProfile;
+                if (currentProfile == null || !currentProfile.EnableSound)
+                {
+                    return;
+                }
+
+                const float SOUND_DELTA = 250.0f;
+                float volume = currentProfile.SoundVolume / SOUND_DELTA;
+
+                if (!Client.Game.IsActive && !currentProfile.ReproduceSoundsInBackground)
+                {
+                    volume = 0;
+                }
+
+                // Rain sound is quieter than normal sound effects
+                volume *= RAIN_VOLUME_MULTIPLIER;
+
+                if (volume > 0 && volume <= 1.0f)
+                {
+                    _currentRainSound.IsLooping = true;
+
+                    if (_currentRainSound.Play(Time.Ticks, volume, 0.0f))
+                    {
+                        // Submit additional buffers upfront for seamless playback (prevents gaps)
+                        // DynamicSoundEffectInstance needs at least 3 buffers for smooth playback
+                        _currentRainSound.SubmitAdditionalBuffers(2);
+
+                        _currentRainSound.X = _world.Player.X;
+                        _currentRainSound.Y = _world.Player.Y;
+                        _currentRainSound.CalculateByDistance = false;
+                    }
+                    else
+                    {
+                        // If sound failed to play, clear the reference
+                        _currentRainSound.IsLooping = false;
+                        _currentRainSound = null;
+                    }
+                }
+            }
+        }
+
+        private void StopRainSound()
+        {
+            if (_currentRainSound != null)
+            {
+                _currentRainSound.IsLooping = false;
+                _currentRainSound.Stop();
+                _currentRainSound = null;
+            }
+        }
+
+        private void UpdateRainSound()
+        {
+            if (Type != WeatherType.WT_RAIN && Type != WeatherType.WT_STORM_APPROACH)
+            {
+                StopRainSound();
+                return;
+            }
+
+
+            // Check if we need to start or restart the rain sound
+            bool shouldPlaySound = Count > 0 && CurrentCount > 0;
+
+            if (!shouldPlaySound)
+            {
+                StopRainSound();
+                return;
+            }
+
+            // Determine if rain is minor or heavy based on density
+            // Minor: Count <= DENSITY_LARGE_DOTS (SmallDots + LargeDots)
+            // Heavy: Count > DENSITY_LARGE_DOTS (ShortLines + LongBolts)
+            bool shouldPlayHeavyRain = Count > DENSITY_LARGE_DOTS;
+
+            // Determine what sound should be playing
+            int expectedSoundId = shouldPlayHeavyRain ? HEAVY_RAIN_SOUND_ID : MINOR_RAIN_SOUND_ID;
+
+            // Only restart if we need to switch between minor/heavy, or if sound is not currently playing
+            // When looping is enabled, sound will continue automatically, so we only need to check for type changes
+            bool needsRestart = _currentRainSound == null || _currentRainSound.Index != expectedSoundId;
+
+            if (needsRestart)
+            {
+                if (shouldPlayHeavyRain)
+                {
+                    PlayHeavyRain();
+                }
+                else
+                {
+                    PlayMinorRain();
+                }
+            }
+        }
+
         private void PlaySound(int sound)
         {
             // randomize the sound of the weather around the player
@@ -492,6 +619,12 @@ namespace ClassicUO.Game
             {
                 _lastTick = Time.Ticks;
                 passed = 25;
+            }
+
+            // Update rain sound looping for rain weather types
+            if (Type == WeatherType.WT_RAIN || Type == WeatherType.WT_STORM_APPROACH)
+            {
+                UpdateRainSound();
             }
 
             bool windChanged = false;
--- a/src/ClassicUO.Client/Game/Weather.cs
+++ b/src/ClassicUO.Client/Game/Weather.cs
@@ -75,7 +75,7 @@ namespace ClassicUO.Game
         // Enable splash effects for long bolts (density 50-70)
         // - Always recommended to be true for storm conditions
 
-        private const float RAIN_VOLUME_MULTIPLIER = 0.30f;
+        private const float RAIN_VOLUME_MULTIPLIER = 0.20f;
         private const int MINOR_RAIN_SOUND_ID = 0x011;
         private const int HEAVY_RAIN_SOUND_ID = 0x010;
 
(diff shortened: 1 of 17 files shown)


markdwags:
I was testing this locally, it looks great. To note, water isn't rippling when raindrops hit static tiles that are wet (ie: the water tiles around all the coastlines).

birdinforest:
Thanks for the note, @markdwags — I’ll create a list of tiles that should generate ripples.

By the way, I’d appreciate any advice on whether there is an existing rain sound effect in the official client assets, or a similar sound effect. I’m not very familiar with the audio assets. Having a rain sound could improve immersion. I can also try tuning the wind sound effect, though I’m not sure how well that would work.

DragonSlayer62:
There is sound most time server sends the sound when the weather starts wind sounds are 0x15 to 0x17 thunder is 0x029 - 0x02A but I can't find any rain or snow sounds. they have water sounds. You can use Uofiddler to look at soundsBy the way this looks really good and think this gives the client some cool stuff

markdwags:
> Thanks for the note, @markdwags — I’ll create a list of tiles that should generate ripples.
> 
> By the way, I’d appreciate any advice on whether there is an existing rain sound effect in the official client assets, or a similar sound effect. I’m not very familiar with the audio assets. Having a rain sound could improve immersion. I can also try tuning the wind sound effect, though I’m not sure how well that would work.

You shouldn't need a list, those static titles also have "water" in the title, and are flagged wet.

DragonSlayer62:
anything flagged wet will get the animation that is the issue with the water waves animation you have barrels and stuff flagged as wet it be better to just list the ids out and not depend on a flag wet

markdwags:
I changed the logic a bit in `IsWaterTileAtPosition`

It works both on static and land tiles, but doesn't show ripples on non-wet static items over wet static tiles (ie: docks over water)

```csharp
// Get the chunk
Chunk chunk = _world.Map.GetChunk(targetTileX, targetTileY, load: false);
if (chunk == null) return false;

// Get the first object in the tile's linked list
GameObject obj = chunk.Tiles[targetTileX % 8, targetTileY % 8];
// Find the highest Z-level object (the one that's actually visible)
GameObject topMostObject = null;
sbyte highestZ = sbyte.MinValue;

while (obj != null)
{
    // Track the object with the highest Z coordinate
    if (obj.Z > highestZ)
    {
        highestZ = obj.Z;
        topMostObject = obj;
    }
    
    obj = obj.TNext;
}

// Now check only the top-most visible object
if (topMostObject != null)
{
    switch (topMostObject)
    {
        case Land land:
            return land.TileData.IsWet && 
                   (land.TileData.Name?.ToLower().Contains("water") == true);
        case Static staticTile:
            return staticTile.ItemData.IsWet && 
                   (staticTile.ItemData.Name?.ToLower().Contains("water") == true);
    }
}

return false;
```

birdinforest:
> I changed the logic a bit in `IsWaterTileAtPosition`
> 
> It works both on static and land tiles, but doesn't show ripples on non-wet static items over wet static tiles (ie: docks over water)
> 
Thanks @markdwags I will use it to check the roof tile avoiding splash

birdinforest:
Update:
+ Avoid generate ripple on tiles over water (deck, bridge...)
+ Avoid generate splash on floor while player is in house.
+ Refactor and add unit tests for ripple and splash for properly particile managment.

DragonSlayer62:
just wander if it's possibly make wind effects near trees with leaves blowing
like in some games have not tree its self but like the leafs falling effect

birdinforest:
> just wander if it's possibly make wind effects near trees with leaves blowing like in some games have not tree its self but like the leafs falling effect

That's a good idea for falling leaves in strong wind or during lumbering. I initially thought representing individual leaf shapes would be challenging with particle effects without textures. However since trees are relatively small compared to real-world scale, the leaves could be represented as blurred particle effects like flying pieces in the wind. 
Could you share a screenshot or the name of the game you mentioned? That would help me understand the visual style.

DragonSlayer62:
here is a video using effects like you talked in unity
https://www.bing.com/videos/search?q=game+that+has+leafs+blowing+in+the+wind+that+is+isomatreic+2d+game&&view=detail&mid=040F0034574195DE00C5040F0034574195DE00C5&FORM=VRDGAR

could possibly do the same thing just he the effects green or brownish yellow if the season set to fall ect

birdinforest:
Add looped raining sound effects depends on density.
Add UI toggles for sound effects and splash & ripple particle animations.

Ready for review.

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                        Rectangle outerRect = new Rectangle(
+                                            newX - outerSize / 2,
+                                            newY - outerSize / 2,
+                                            outerSize,
+                                            outerSize
+                                        );
+
+                                        batcher.Draw
+                                        (
+                                            SolidColorTextureCache.GetTexture(outerColor),
```
andreakarasho: This function will create a lot of textures if the `outerColor` changes.
This might be a good opportunity to add the support for RGBA color drawing function instead of UO color.

Reference and requirements:
Function: https://github.com/ClassicUO/ClassicUO/blob/impl/ecs/src/ClassicUO.Renderer/Batcher2D.cs#L123
Shader: https://github.com/ClassicUO/ClassicUO/blob/impl/ecs/src/ClassicUO.Renderer/shaders/IsometricWorld.fx

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                        Rectangle midOuterRect = new Rectangle(
+                                            newX - midOuterSize / 2,
+                                            newY - midOuterSize / 2,
+                                            midOuterSize,
+                                            midOuterSize
+                                        );
+
+                                        batcher.Draw
+                                        (
+                                            SolidColorTextureCache.GetTexture(midOuterColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                        Rectangle middleRect = new Rectangle(
+                                            newX - middleSize / 2,
+                                            newY - middleSize / 2,
+                                            middleSize,
+                                            middleSize
+                                        );
+
+                                        batcher.Draw
+                                        (
+                                            SolidColorTextureCache.GetTexture(middleColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                    Rectangle smallCoreRect = new Rectangle(
+                                        newX - smallCoreSize / 2,
+                                        newY - smallCoreSize / 2,
+                                        smallCoreSize,
+                                        smallCoreSize
+                                    );
+
+                                    batcher.Draw
+                                    (
+                                        SolidColorTextureCache.GetTexture(smallCoreColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                        Rectangle outerRect = new Rectangle(
+                                            newX - outerSize / 2,
+                                            newY - outerSize / 2,
+                                            outerSize,
+                                            outerSize
+                                        );
+
+                                        batcher.Draw
+                                        (
+                                            SolidColorTextureCache.GetTexture(outerColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                        Rectangle midOuterRect = new Rectangle(
+                                            newX - midOuterSize / 2,
+                                            newY - midOuterSize / 2,
+                                            midOuterSize,
+                                            midOuterSize
+                                        );
+
+                                        batcher.Draw
+                                        (
+                                            SolidColorTextureCache.GetTexture(midOuterColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                        Rectangle middleRect = new Rectangle(
+                                            newX - middleSize / 2,
+                                            newY - middleSize / 2,
+                                            middleSize,
+                                            middleSize
+                                        );
+
+                                        batcher.Draw
+                                        (
+                                            SolidColorTextureCache.GetTexture(middleColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                    Rectangle largeCoreRect = new Rectangle(
+                                        newX - largeCoreSize / 2,
+                                        newY - largeCoreSize / 2,
+                                        largeCoreSize,
+                                        largeCoreSize
+                                    );
+
+                                    batcher.Draw
+                                    (
+                                        SolidColorTextureCache.GetTexture(largeCoreColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                    // Boost alpha for visibility
+                                    float shortLineAlpha = Math.Max(0.65f, rainDepthProps.AlphaMultiplier);
+                                    shortLineColor *= shortLineAlpha;
+
+                                    // Apply line width with depth
+                                    int shortLineWidth = Math.Max(1, (int)(2 * rainDepthProps.SizeMultiplier));
+
+                                    batcher.DrawLine
+                                    (
+                                        SolidColorTextureCache.GetTexture(shortLineColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                    // Boost alpha for dramatic effect
+                                    float boltAlpha = Math.Max(0.75f, rainDepthProps.AlphaMultiplier);
+                                    boltColor *= boltAlpha;
+
+                                    // Apply line width with depth
+                                    int boltLineWidth = Math.Max(1, (int)(3 * rainDepthProps.SizeMultiplier));
+
+                                    batcher.DrawLine
+                                    (
+                                        SolidColorTextureCache.GetTexture(boltColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                                oldY = (int)(newY - MAX_OFFSET_XY);
+                            else if (screenOfsy <= -MAX_OFFSET_XY)
+                                oldY = (int)(newY + MAX_OFFSET_XY);
+
+                            Vector2 start = new Vector2(oldX, oldY);
+                            Vector2 end = new Vector2(newX, newY);
+
+                            batcher.DrawLine
+                            (
+                                SolidColorTextureCache.GetTexture(Color.Blue),
```
andreakarasho: Worth to swap to RGBA new function when implemented

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                            Rectangle outerRect = new Rectangle(
+                                snowX - outerSize / 2,
+                                snowY - outerSize / 2,
+                                outerSize,
+                                outerSize
+                            );
+
+                            batcher.Draw
+                            (
+                                SolidColorTextureCache.GetTexture(outerColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                            Rectangle midOuterRect = new Rectangle(
+                                snowX - midOuterSize / 2,
+                                snowY - midOuterSize / 2,
+                                midOuterSize,
+                                midOuterSize
+                            );
+
+                            batcher.Draw
+                            (
+                                SolidColorTextureCache.GetTexture(midOuterColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                            Rectangle middleRect = new Rectangle(
+                                snowX - middleSize / 2,
+                                snowY - middleSize / 2,
+                                middleSize,
+                                middleSize
+                            );
+
+                            batcher.Draw
+                            (
+                                SolidColorTextureCache.GetTexture(middleColor),
```
andreakarasho: Same

src/ClassicUO.Client/Game/Weather.cs
```diff
@@ -409,78 +894,675 @@ public void Draw(UltimaBatcher2D batcher, int x, int y, float layerDepth)
...
+                            snowY - coreSize / 2,
+                            coreSize,
+                            coreSize
+                        );
 
                         batcher.Draw
                         (
-                            SolidColorTextureCache.GetTexture(Color.White),
-                            snowRect,
+                            SolidColorTextureCache.GetTexture(coreColor),
```
andreakarasho: Same

andreakarasho (changes requested):

birdinforest:
Hi @andreakarasho thanks for review. I just updated:
1. apply RGBA color to only create one texture
2. apply defensive programming, any fails won't crash game
3. update unit test to cover above changes

Please review when getting a chance.

PR #1854, opened 2026-01-08
Use a proper UnmanagedType for tiledata name so we can correctly get marshalled size

kaczy93:
https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshalasattribute.sizeconst?view=net-10.0#remarks
https://learn.microsoft.com/en-us/dotnet/framework/interop/default-marshalling-for-strings#strings-used-in-structures


andreakarasho (approved):
Thanks!

PR #1861, opened 2026-02-21
TargetSystem changes

canerksk:
The "on/off still incomplete" message for the new target system macro has been fixed.

The new target system wasn't working as it should in the OSI; the gumps designated for the target system were being painted, but this shouldn't have been the case. Only the top point gump should have changed according to the Noto color.
OSI;
image
image

CUO (current);
image

CUO (after the change);
image
image

Furthermore, the target system needed to be marked not only when attacking via warmod, but also when double-clicking normally.

### DIFF
--- a/src/ClassicUO.Client/Game/GameActions.cs
+++ b/src/ClassicUO.Client/Game/GameActions.cs
@@ -304,12 +304,16 @@ namespace ClassicUO.Game
                 Attack(world, serial);
             }
             else
-            {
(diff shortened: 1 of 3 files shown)


andreakarasho (approved):
LGTM, thank you!

canerksk:
> LGTM, thank you!

Thank you, that was a very quick accepted, I honestly wasn't prepared for it 😅

PR #1862, opened 2026-02-23
Stitchin support

andreakarasho:


Lichtblitz:
There is still a bug when unequipping an item:

https://github.com/user-attachments/assets/3901d176-e3a0-42f7-96d4-0edc02a4a9c3

And this is what it's supposed to look like:


https://github.com/user-attachments/assets/d3f34bb2-d985-4c92-afdd-2738c176935e

Lichtblitz:
This is what a female fancy shirt is supposed to do:
```
# 1435
coveredBy TORSO UPPER_ARMS_TOP UPPER_ARMS_BOTTOM LOWER_ARMS_TOP LOWER_ARMS_BOTTOM
covers TORSO UPPER_ARMS_TOP UPPER_ARMS_BOTTOM LOWER_ARMS_TOP LOWER_ARMS_BOTTOM
//--- Hide bustiers
remove 1924 1925
# enddef
```

I'm having trouble understanding the `coveredBy` and the `covers` statements. If it does both at once, how is it decided, which one it actually covers or is covered by?

Lichtblitz:
These are the leather sleeves:

```
//H_Female_Armor_Leather_Sleeves_V2
# 1544
coveredBy UPPER_ARMS_TOP UPPER_ARMS_BOTTOM LOWER_ARMS_TOP LOWER_ARMS_BOTTOM
covers UPPER_ARMS_TOP UPPER_ARMS_BOTTOM LOWER_ARMS_TOP LOWER_ARMS_BOTTOM
//--- Replace long sleeve shirts with tunic only models
replace 1435 with 3261
replace 1538 with 3251
replace 1911 with 3271
replace 1691 with 3278
replace 1734 with 3285
replace 1907 with 3930
replace 1908 with 3932
# enddef
```

They should have replaced `1435` with a different ID - which didn't happen. I have validated it with the original client and it does indeed replace the fancy shirt with one without arms:


https://github.com/user-attachments/assets/1596ec22-078a-4691-a6ce-cf329e47dca9

Lichtblitz:
This looks better but it's not quite there, yet:

https://github.com/user-attachments/assets/e0f51df0-b7a0-4ea2-85dd-78a638da2a50

Lichtblitz:
The female chest is still above the pants, but should be underneath.

Original Client:
image

Current State:
image

Apart from that it looks mostly right. The female chest is underneath the shirt but it should instead not be drawn in that combination. A shoulder piece is peeking through. But I would consider that a minor issue.

andreakarasho:
Stitchin seems to not be supported by the CC. This should cover all cases: https://github.com/ClassicUO/ClassicUO/pull/1901

PR #1867, opened 2026-03-06
Fixes offset for firing bow/xbow/heavy-xbow & directed spells

kamronbatman:
### Problem

When a character fires a bow/crossbow or casts a directed spell, the moving effect originates from the character's **tile anchor point (feet)**. In certain directions (especially South, SE, SW), this makes projectiles appear to fire from the ground rather than from the character's hand.

The 0xC7 (`HuedEffect`/`ParticleEffect`) packet includes a `layer` byte that tells the client which body part the effect originates from (e.g., `RightHand = 1`). ClassicUO reads this byte but **discards it** — it is never passed to the effect pipeline.

### Solution

**Two detection paths** for maximum compatibility:

1. **Explicit (0xC7 packet)**: When the server sends `layer = 1` (RightHand), apply hand offsets. The animation group distinguishes bow, crossbow, and directed spell.
2. **Heuristic (0xC0 packet)**: When no layer info is present, check if the source mobile is currently playing a ranged attack or directed cast animation. This enables the fix for legacy servers that only send 0xC0.

Pixel offsets are applied in `MovingEffect` as additions to `Offset.X`/`Offset.Y`, the same mechanism already used for the baseline `+22` half-tile centering.

#### Ranged weapons (bow/crossbow)

The offset table was produced by a purpose-built [AnimFrameAnalyzer](https://discord.com/channels/751317910504603701/1479366801916821628) (Can be found on the [ModernUO Discord](https://muo.gg/discord)) tool that composites body 400 (human male) with each weapon's equipment overlay from `anim.mul`–`anim5.mul` at the **release frame** of each attack animation.

Results can be downloaded here: [Averaged-Grip-Contacts.zip](https://github.com/user-attachments/files/25787747/Averaged-Grip-Contacts.zip)

**Method** (extremity-biased weighted centroid, averaged across weapons):

1. For each weapon x direction x mounted state, composite body + equipment at the release frame using ClassicUO's Y anchor formula: `screenY = pixelY - (Height + CenterY)`
2. Find **contact points** — equipment pixels adjacent (within radius 2) to body pixels in screen space (the grip zone)
3. Compute a **bias direction** from the body center toward the equipment's visual centroid
4. **Weight** each contact point by its projection along the bias direction using quadratic falloff — contacts toward the weapon tip (front hand / arrow release point) receive exponentially more weight:
   ```
   bias = normalize(equipmentCentroid - bodyCenter)
   projection = dot(contact, bias)
   t = (projection - min) / range       // 0 = body side, 1 = weapon tip
   weight = 0.1 + t^2
   ```
5. **Average** the weighted grip points across all weapons sharing the same action, producing 4 universal tables:
   - **ShootBow** (5 bows: Bow, Composite, Elven Composite, Magical Shortbow, Yumi)
   - **MountedShootBow** (same 5)
   - **ShootXBow** (3 crossbows: Crossbow, Heavy Crossbow, Repeating Crossbow)
   - **MountedShootXBow** (same 3)

The max deviation of any individual weapon from its group average is ~5px — imperceptible in-game. New weapons using the same animation groups get correct offsets automatically.

#### Example Heuristic Screenshots
AVG_Bow_ShootBow_dir6_West_f3_RELEASE
AVG_Bow_MountedShootBow_dir2_East_f2_RELEASE

#### Directed spells (Magic Arrow, Fireball, Energy Bolt)

The spell offset table was produced by a SpellFrameAnalyzer, a tool similar to the AnimFrameAnalyzer that renders body 400 at the **release frame** of the CastDirected animation (group 16), then **hand-tuned** by visually placing offset markers on the casting hand in each of the 8 directions.

Dirs 5–7 are horizontal mirrors of dirs 3–1 (X negated, Y identical), matching UO's animation storage format. Only unmounted offsets are needed — the server does not send cast animation while mounted (`Spell.cs:917`).

#### Example Tuned Screenshots
CastDirected_dir3_SE_f3_RELEASE
CastDirected_dir2_East_f3_RELEASE

### Design notes

- **No body-type check needed**: `PeopleAnimationGroup` values only match human animation groups, so non-human mobiles firing projectiles naturally skip the offset logic.
- **Default parameter (`0xFF`)**: All existing callers are unaffected — the layer defaults to "no info" which triggers the heuristic path only.
- **Explicit path is precise**: With `layer = 1`, each animation group is explicitly matched (`AttackBow`, `AttackCrossbow`, `CastDirected`). Unrecognized animations receive zero offset — no default-to-bow fallback.
- **Heuristic path detects all three types**: `AttackBow`/`OnmountAttackBow`, `AttackCrossbow`/`OnmountAttackCrossbow`, and `CastDirected` are all checked in the 0xC0 path for legacy server compatibility.
- **Other spells unaffected**: Spells using `CastArea` (group 17) or non-projectile effects (Lightning, Heal, etc.) don't match any detected animation group and receive zero offset.

### Test plan

- [ ] Fire bow/crossbow unmounted in all 8 directions — arrows originate near hand
- [ ] Fire bow/crossbow mounted in all 8 directions — arrows originate near hand
- [ ] Cast Magic Arrow in all 8 directions — projectile originates near casting hand
- [ ] Cast Fireball in all 8 directions — projectile originates near casting hand
- [ ] Cast Energy Bolt in all 8 directions — projectile originates near casting hand
- [ ] Spell projectiles track moving targets (homing preserved)
- [ ] Spell reflect — projectile originates from defender's hand
- [ ] NPC archers/casters — offsets apply to their projectiles too
- [ ] Non-projectile spell effects (Lightning, Heal, etc.) — no visual change
- [ ] Drag effects / other moving effects — no visual change
- [ ] Cast while mounted — no offset applied (expected, no animation sent)
- [ ] Test with 0xC0-only server (legacy) — heuristic detection still works

### DIFF
--- a/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
+++ b/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
@@ -351,6 +351,7 @@ namespace ClassicUO.Game.GameObjects
             bool fromServer = false
         )
         {
+            Console.WriteLine($"{this} is doing animation {id}");
             _animationGroup = id;
             AnimIndex = (byte)(forward ? 0 : frameCount);
             _animationInterval = interval;
--- a/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
+++ b/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
@@ -10,6 +10,7 @@ using ClassicUO.Utility;
 using ClassicUO.Utility.Collections;
 using Microsoft.Xna.Framework;
 using System;
+using ClassicUO.Renderer.Animations;
 
 namespace ClassicUO.Game.GameObjects
 {
@@ -351,7 +352,6 @@ namespace ClassicUO.Game.GameObjects
             bool fromServer = false
         )
         {
-            Console.WriteLine($"{this} is doing animation {id}");
             _animationGroup = id;
             AnimIndex = (byte)(forward ? 0 : frameCount);
             _animationInterval = interval;
@@ -579,7 +579,7 @@ namespace ClassicUO.Game.GameObjects
             byte action = GetGroupForAnimation(this, id, true);
 
             bool mirror = false;
-            animations.GetAnimDirection(ref dir, ref mirror);
+            Animations.GetAnimDirection(ref dir, ref mirror);
             int currentDelay = Constants.CHARACTER_ANIMATION_DELAY;
 
             if (id < animations.MaxAnimationCount && dir < 5)
--- a/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
+++ b/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
@@ -100,6 +100,7 @@ namespace ClassicUO.Game.GameObjects
         private long _lastAnimationIdleDelay;
         private bool _isAnimationForwardDirection;
         private byte _animationGroup = 0xFF;
+        public byte AnimationGroup => _animationGroup;
         private byte _animationInterval;
         private bool _animationRepeat;
         private ushort _animationRepeateMode = 1;
--- a/src/ClassicUO.Client/Game/GameObjects/Item.cs
+++ b/src/ClassicUO.Client/Game/GameObjects/Item.cs
@@ -7,6 +7,7 @@ using ClassicUO.Game.Scenes;
 using ClassicUO.Game.UI.Gumps;
 using Microsoft.Xna.Framework;
 using System;
+using ClassicUO.Renderer.Animations;
 
 namespace ClassicUO.Game.GameObjects
 {
@@ -498,7 +499,7 @@ namespace ClassicUO.Game.GameObjects
                 bool mirror = false;
 
                 var animations = Client.Game.UO.Animations;
-                animations.GetAnimDirection(ref dir, ref mirror);
+                Animations.GetAnimDirection(ref dir, ref mirror);
 
                 if (id < animations.MaxAnimationCount && dir < 5)
                 {
--- a/src/ClassicUO.Client/Game/GameObjects/MovingEffect.cs
+++ b/src/ClassicUO.Client/Game/GameObjects/MovingEffect.cs
@@ -1,7 +1,6 @@
 ﻿// SPDX-License-Identifier: BSD-2-Clause
 
 using System;
-using ClassicUO.Assets;
 using ClassicUO.Game.Data;
 using ClassicUO.Game.Managers;
 using Microsoft.Xna.Framework;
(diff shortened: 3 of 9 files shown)


DragonSlayer62:
The layer btye for the effects is only for Enhanced client 2d clients ignore this completely

kamronbatman:
> The layer btye for the effects is only for Enhanced client 2d clients ignore this completely

That's fine. We should support it regardless, doesn't hurt.

kamronbatman:
Upon testing, this won't work because RunUO/ServUO/ModernUO sends the following: Swing Packet, Effect Packet, Animation Packet. This order is technically wrong, but the old EA client didn't care cause it was hard coded to "do it right."

I'll fix this on ModernUO specifically so the order is correct. Meanwhile, the next best thing we can do for ClassicUO is to make inferences. If swing packet was just sent and we are on an effect packet, and the mobile is holding a specific weapon, assume it will accompany a specific animation even though we haven't received it.

src/ClassicUO.Client/Game/GameObjects/Mobile.cs
```diff
@@ -350,6 +351,7 @@ public void SetAnimation(
             bool fromServer = false
         )
         {
+            Console.WriteLine($"{this} is doing animation {id}");
```
kamronbatman: I'll remove this.

DragonSlayer62:
I’ve been reviewing the UOX3 packet order regarding projectile effects.

Previously, in UOX3, I fixed the visual issue where spells and arrows appeared to drag along the ground by adjusting the Z value of the effect between the attacker and defender. This corrected the visual trajectory and made projectiles appear properly elevated during flight.

However, since this topic came up again, it made me wonder if the underlying issue could instead be related to packet ordering.

That said, my assumption is that packet order alone should not affect the Z-axis of the visual effect, since the axis position should be determined by the effect data itself rather than the order in which packets arrive. I could be mistaken on that point, though.

I’ve also asked Xuri about this to get their perspective.

I would absolutely like to see this handled properly if there is a more correct solution. My only concern is that any packet changes might introduce side effects in other emulators, since different clients and emulator implementations may rely on slightly different packet behaviors.

That’s the main reason I wanted to bring this up for discussion.

kamronbatman:
Looking at how the EA client does it, which is "correct", it's hard coded. I'll get more info soon here and update this post.

EDIT: The EA client has an offset for the "head" which is the default effect layer, mounted and unmounted, and uses that as a hardcoded offset. That's it. Technically this change will be even better, because we can detect it by effect id, and I can default it back to the head accordingly.

Another Edit: I am working to tune the numbers and get it to look good. I'll get back to you guys on this soon.

kamronbatman:
I got the arrows/bolts adjusted almost perfectly, even better than the EA client. For spells I need to redo the calculation because the model is not perfectly centered in all directions (for example UP -> 17pixels off from center). By default this makes all effects look like they fire from the left foot, or behind the model depending on direction.

When I get some more time I can probably fix this fundamentally and then readjust arrows/bolts accordingly. The correct calculation (to match EA) is from the face. Not only that, but with the _back_ of the projectile/effect starting in front of the face.

DragonSlayer62:
I feel like this could possibly break people's shard even custom shards have custom spells because you're trying to force the client to do the animations instead just leaving the packet up to the emulator to do the work. Seems like this can go wrong on so many levels. instead, just simply changing the z correctly in the code of the emulator,  but just my thoughts.

kamronbatman:
> I feel like this could possibly break people's shard even custom shards have custom spells because you're trying to force the client to do the animations instead just leaving the packet up to the emulator to do the work. Seems like this can go wrong on so many levels. instead, just simply changing the z correctly in the code of the emulator, but just my thoughts.

I'm not sure I understand. This is how the EA client works. The anchor point is the back of the visual part of the effect against the front of the face after accounting for rotation. And arrows/bolts are manually manipulated from there to look even more realistic.

Emulators should never be manipulating the Z level based on the current anchor point in CUO (the foot), because that is wrong anchor point and has always been wrong.

Additionally, changing the Z coordinate disconnects the effect from the source, so moving whole sending the animation, depending on timing will break it, and that is after the fact that Z is too course for proper manipulation.

DragonSlayer62:
Can I ask where you are getting your source of information for the UO client code, I'm not trying to argue or anything but if you have the code that UO uses for the client it could help allot of people out. Even classic uo because you are saying this how works and I think if we had the source of this information, Lot of things we could get working in CUO.

kamronbatman:
I used EA Client 5.x, 6.x, and 7.x against RunUO/ModernUO, and newest classic client against OSI with packet captures. It's easy to see the source to destination just visually.

While I don't condone reverse engineering software, it's fairly easy to do in this case to gather more info because the fireball effect and the struct is easy to trace.

PR #1868, opened 2026-03-06
Fixes keep alive on server select / character creation

kamronbatman:
### Summary

The ping time for server selection, character creation etc is too long. ModernUO no longer waits 90s for DDoS/attack reasons and now only waits 30s, expecting a nominal ping every few seconds. While we don't need pings every 1s like in-game, 10s should suffice, otherwise players designing their character will get booted for inactivity after some time.

I can understand if we don't want to send pings during server selection, it shouldn't take that long to choose a server. Let me know what you think and I can change it, or you are welcome to update this PR.

### DIFF
--- a/src/ClassicUO.Client/Game/Scenes/LoginScene.cs
+++ b/src/ClassicUO.Client/Game/Scenes/LoginScene.cs
@@ -161,7 +161,7 @@ namespace ClassicUO.Game.Scenes
                 }
             }
 
-            if ((CurrentLoginStep == LoginSteps.CharacterCreation || CurrentLoginStep == LoginSteps.CharacterSelection) && Time.Ticks > _pingTime)
+            if ((CurrentLoginStep == LoginSteps.CharacterCreation || CurrentLoginStep == LoginSteps.CharacterSelection || CurrentLoginStep == LoginSteps.ServerSelection) && Time.Ticks > _pingTime)
             {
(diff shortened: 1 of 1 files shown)


DragonSlayer62:
Have you checked with other emus make sure this will not mess anything up for POL/Sphere/UOX3/Servuo
Also, I have never been booted for character creation so that has to be an emu Design in modernUO to boot on character selection because the server sends the timeout to the char when should be booted that is a setting each emu can change and the timer shouldn't start for the boot until the char logins into the world not less that is a different timer started like connection timer.

I will test this out in uox3 using your change

kamronbatman:
> Have you checked with other emus make sure this will not mess anything up for POL/Sphere/UOX3/Servuo Also, I have never been booted for character creation so that has to be an emu Design in modernUO to boot on character selection because the server sends the timeout to the char when should be booted that is a setting each emu can change and the timer shouldn't start for the boot until the char logins into the world not less that is a different timer started like connection timer.
> 
> I will test this out in uox3 using your change

RunUO/ServUO boots after 90 seconds, so the current 60s is enough. The server selection doesn't send pings so you get booted on all emus that have a keep alive. ModernUO will be fixing this with a "ping back" too.

DragonSlayer62:
Xuri Confirmed we already have the ping packet on uox3 emu implanted, I tested the change you made on uox3 seems to work fine and no issues, again just me giving feedback

kamronbatman:
I fixed this on MUO by adding an RTT, so I don't care anymore.

PR #1869, opened 2026-03-07
Updated Mount Visuals

DragonSlayer62:
Implemented a special rendering path for select mounted body IDs that were exhibiting rider clipping issues in ClassicUO. Normally, mounted creatures are drawn using a vertical slice-based depth layering system intended to improve front/back occlusion between the rider and mount. While this works for most standard mounts, some newer or nonstandard mounted bodies do not match the assumptions of that depth model. As a result, parts of the rider most notably legs could render incorrectly through the mount’s torso or ribs.

For the affected mounts, rendering is now forced to use a single-layer draw instead of the generic slice-based depth pass. This preserves proper visual composition for these bodies and eliminates obvious clipping artifacts without changing rendering behavior for normal mounts.

Affected mounted body IDs:

0x04E6 Tiger
0x04E7 Tiger (alternate)
0x0611  Manticore
0x05F7 Capybara
0x0666  Bear Zombie
0x05A1 Skeletal Cat


GladieUO:
I think that the issue with this now is that those mounts have feets inside the ground now...

DragonSlayer62:
Not sure this old pr since they updated mesh stuff so could be but when i first created this i had no visual issues it matched UO client to the T.

PR #1872, opened 2026-03-13
fix: wide item render being overwritten by southern statics

jackuoll:
Fixes this issue in the screenshot which has been present since at least 2021 and the fix matches it with the OSI view.

image

Testing involved walking around the world looking for incorrect renders, such as stairs, multis, boats, clumped trees and house placements. No visible problems.


DragonSlayer62:
I added you fix and if you run up to the bulletin board it puts the char behind it now.

jackuoll:
Will come back to this, closing for now.

PR #1882, opened 2026-03-23
Update GridLoot hueVector calculation for item rendering

canerksk:
The issue of items in GridLoot not displaying their colors and the mouse cursor completely painting the grid has been fixed.

before;
normal:
image

hover:
image

after;

normal;
image

hover;
image


andreakarasho (approved):

andreakarasho:
ty!

PR #1888, opened 2026-04-23
Fixes rendering cropped text. Partially fixes regression in gumps.

kamronbatman:
## Summary

Fixes blank `CroppedText` labels whose string exceeds `maxWidth` (e.g. long
property names in ModernUO's `[props` gump — `LongTermMurderExpiration`,
`AllianceMessageHue`, `AutoRenewInsurance` rendered as empty space). The EA
client renders the same server packets correctly; repro is any server gump
with a `{ croppedtext ... }` element whose text is longer than its column.

Two distinct defects are fixed. A third change rewrites the non-atlas gump
queue to eliminate per-frame closure allocations uncovered during the
investigation.

### Bug 1: atlas renderer draws from `_info.Data`, but `_info` wasn't carrying the cropped/ellipsis text

Commit 119108d3 ("replace per-string text textures with shared glyph atlas")
moved drawing from a per-string `Texture2D` to per-glyph quads pulled from a
shared atlas. `RenderedText.DrawGlyphs` iterates `_info.Data` (a
`MultilinesFontInfo` populated by `GetInfoUnicode` / `GetInfoASCII`) and places
a quad for each character.

Before that commit the texture-based path used `GenerateUnicode`, which calls
`GetTextByWidthUnicode` internally when text exceeds `maxWidth` — that helper
reserves space for `"..."`, truncates to fit, and appends the three dots into
the rasterised texture. `_info` was not on the draw path, so nothing noticed
that `_info` didn't carry the cropped string.

After 119108d3, `_info` IS on the draw path, but the `Text` setter still
passes the *full* string to `GetInfoUnicode`. Cropping happens implicitly
inside `GetInfo*` via the `UOFONT_CROPPED` branch — and that branch doesn't
append an ellipsis.

### Bug 2: `GetInfo*` truncates `Data.Length` to 0 for single-word overflow

Deeper in `GetInfoUnicode` / `GetInfoASCII` / `GetInfoHTML`, the wrap/break
handler does:

```csharp
i = lastSpace + 1;
// ...
ptr.Data.Length = ptr.CharCount;
if (isFixed || isCropped) break;
```

The local `charCount` counter is only rolled up into `ptr.CharCount` at space
characters. For a text with no whitespace (property names, usernames, a lot of
UO labels) `ptr.CharCount` is still `0` when we hit the overflow. Setting
`Data.Length = 0` drops every character that was pushed into `ptr.Data` during
the loop.

The pre-atlas renderer didn't care because it drew from a texture built out of
the final cropped string via `GeneratePixelsUnicode`, not from `_info.Data`.
The atlas renderer iterates `_info.Data` and ends up with an empty loop → the
label renders blank.

Both defects compound: even if Bug 2 were fixed in isolation, the label would
render its truncated content without the `"..."` marker.

### 1. Pre-crop in `RenderedText.Text` setter (visible fix)

`src/ClassicUO.Client/Game/GameObjects/RenderedText.cs`

When `FontStyle.Cropped` is set and `MaxWidth > 0`, pre-compute the cropped
string via the existing `GetTextByWidthUnicode` / `GetTextByWidthASCII` helper
and hand that to `GetInfoUnicode` / `GetInfoASCII`. Restores the pre-atlas
invariant that the layout info carries exactly what will be drawn, including
the `"..."` ellipsis. Short text that doesn't need cropping takes the cheap
`realWidth > MaxWidth` check and skips the extra work.

### 2. Commit pending chars before `Data.Length = CharCount` in `GetInfo*`

`src/ClassicUO.Assets/FontsLoader.cs` — three call sites (`GetInfoUnicode`,
`GetInfoASCII`, `GetInfoHTML`)

Defence-in-depth for Bug 2. When the overflow handler takes the
non-wrap-to-next-line branch with `isCropped && !isFixed`, roll
`readWidth`/`charCount` into `ptr.Width`/`ptr.CharCount` before the
`Data.Length = CharCount` truncation. Without this, any future caller that
sets `FontStyle.Cropped` without pre-cropping would hit the same blank-label
symptom. Two lines per site:

```csharp
else if (isCropped)
{
    ptr.Width += readWidth;
    ptr.CharCount += charCount;
}
```

### 3. Replace closure queue in `RenderLists` with typed struct commands

`src/ClassicUO.Client/Game/Scenes/RenderLists.cs`,
`src/ClassicUO.Client/Game/UI/Controls/CroppedText.cs`,
`src/ClassicUO.Client/Game/UI/Controls/Label.cs`,
`src/ClassicUO.Client/Game/UI/Controls/Button.cs`

Orthogonal to the render-correctness fix but came out of the same
investigation. `AddGumpNoAtlas` previously took a
`Func<UltimaBatcher2D, bool>`; every non-atlas draw allocated a closure per
frame. The existing code had a candid self-assessment in the file:

> This is an intermediate, crappy solution. Rewriting gump rendering would be
> way too much at this point.

This PR adds a tagged-union `NoAtlasGumpCommand` struct (either a
`RenderedText`-based draw or a `Func<>` fallback for compound ops like
clipping / nested render lists) and rewrites `AddGumpNoAtlas` plus
`DrawRenderListNoAtlas` around it. Insertion order is preserved across mixed
entries. The `Func` overload is kept for callers that need arbitrary draws
(`HtmlControl`, `StbTextBox`, `LoginGump` inner textbox, `HoveredLabel`,
`Line`, `ScissorControl`, etc.) — they work unchanged via the fallback path.

Text-drawing controls converted to the struct overload: `CroppedText`,
`Label`, and `Button` captions.

**Allocation impact** (informal): closure path was ~40+ bytes per
`AddGumpNoAtlas` call. For a PropsGump-style layout (~30 `CroppedText`
instances) at 60 FPS that's ~72 KB/s of Gen0 churn from one gump. Struct path
allocates nothing per call once the backing list has grown to capacity.

Draw-time includes an explicit `HasContent` guard so a destroyed/recycled
`RenderedText` (pool recycling via `QueuedPool<RenderedText>`) is skipped
rather than drawn as stale state.

### Before
Before

### After
After

## Tests

New: `tests/ClassicUO.UnitTests/Game/Scenes/RenderListsTests.cs` (10 tests).

- Null-text / null-func guards (both paths no-op cleanly).
- Insertion-order preservation across mixed entries.
- `Clear()` empties the queue.
- Struct constructor field coverage (both text and callback ctors).
- **Allocation-budget regression guard**: after warm-up, 1000
  `AddGumpNoAtlas` calls must allocate less than 1 KB. Fails immediately if
  closures are reintroduced into the queue.

Infrastructure note: `RenderedText.Create` and `UltimaBatcher2D.Draw` both need
graphics init, so the flush path itself isn't exercised directly. The queue /
ordering / allocation-budget behaviour is covered at the boundary. Internal
test accessors (`GumpTextsCount`, `PeekGumpText`) let the tests inspect the
queue without a live graphics device.

Full test suite: 172 pass (162 pre-existing + 10 new).

## Test plan

- [x] `dotnet build` clean (no new warnings).
- [x] `dotnet test` green (172/172).
- [x] `[props` on a PlayerMobile on ModernUO shows every row with both name
      and value, including long cropped names like `LongTermMurderExpiration`
      now rendering as `LongTermMurder...` within the column.
- [x] World rendering untouched — tiles, statics, mobiles all draw normally;
      only the gump-text path was modified.
- [x] Reviewer sanity: other gumps using `CroppedText` (skills, vendor menus,
      party list) render correctly.
- [ ] Reviewer sanity: non-text `AddGumpNoAtlas` callers (`HtmlControl`
      clipping, `HoveredLabel` hover rect, `Line`, scissor controls) behave
      unchanged.

## Related

- Commit 119108d3 (`replace per-string text textures with shared glyph atlas`)
  — the refactor that exposed the cropped-text defect.
- Commit dcd8d3fa (`Refactored the rendering stack`) — earlier refactor that
  introduced the closure-based `_gumpTexts` queue this PR replaces.

### DIFF
--- a/src/ClassicUO.Assets/FontsLoader.cs
+++ b/src/ClassicUO.Assets/FontsLoader.cs
@@ -10,7 +10,6 @@ using System.Globalization;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using System.Threading.Tasks;
 
 namespace ClassicUO.Assets
 {
@@ -539,7 +538,7 @@ namespace ClassicUO.Assets
 
                 if (size > 0)
                 {
-                    sb.Append(str.Substring(0, size));
+                    sb.Append(str.AsSpan(0, size));
                     str = str.Substring(str.Length - strLen, strLen);
 
                     if (GetWidthASCII(font, str) < width)
@@ -946,6 +945,17 @@ namespace ClassicUO.Assets
                             }
 
                             charCount++;
+                            ptr.Width += readWidth;
+                            ptr.CharCount += charCount;
+                        }
+                        else if (isCropped)
+                        {
+                            // Commit the characters accumulated so far. Without this, a
+                            // cropped single-word string (no spaces) would exit with
+                            // CharCount == 0 and the subsequent `Data.Length = CharCount`
+                            // would truncate every processed char out of ptr.Data — the
+                            // glyph-atlas draw path then has nothing to iterate and the
+                            // label renders blank.
                             ptr.Width += readWidth;
                             ptr.CharCount += charCount;
                         }
@@ -1506,6 +1516,17 @@ namespace ClassicUO.Assets
                             }
 
                             charCount++;
+                            ptr.Width += readWidth;
+                            ptr.CharCount += charCount;
+                        }
+                        else if (isCropped)
+                        {
+                            // Commit the characters accumulated so far. Without this, a
+                            // cropped single-word string (no spaces) would exit with
+                            // CharCount == 0 and the subsequent `Data.Length = CharCount`
+                            // would truncate every processed char out of ptr.Data — the
+                            // glyph-atlas draw path then has nothing to iterate and the
+                            // label renders blank.
                             ptr.Width += readWidth;
                             ptr.CharCount += charCount;
                         }
@@ -2289,10 +2310,7 @@ namespace ClassicUO.Assets
             bool isFixed = (flags & UOFONT_FIXED) != 0;
             bool isCropped = (flags & UOFONT_CROPPED) != 0;
 
-            if (len != 0)
-            {
-                ptr.Align = htmlData[0].Align;
-            }
+            ptr.Align = htmlData[0].Align;
 
             for (int i = 0; i < len; i++)
             {
@@ -2414,6 +2432,17 @@ namespace ClassicUO.Assets
                             readWidth += charWidth;
                             ptr.MaxHeight = MAX_HTML_TEXT_HEIGHT;
                             charCount++;
+                            ptr.Width += readWidth;
+                            ptr.CharCount += charCount;
+                        }
+                        else if (isCropped)
+                        {
+                            // Commit the characters accumulated so far. Without this, a
+                            // cropped single-word string (no spaces) would exit with
+                            // CharCount == 0 and the subsequent `Data.Length = CharCount`
+                            // would truncate every processed char out of ptr.Data — the
+                            // glyph-atlas draw path then has nothing to iterate and the
+                            // label renders blank.
                             ptr.Width += readWidth;
                             ptr.CharCount += charCount;
                         }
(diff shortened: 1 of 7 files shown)


kamronbatman:
After this is merged, I have a full-overhaul to handle the performance regressions.

andreakarasho (approved):
LGTM, thank you

PR #1892, opened 2026-04-25
Removes FastList<T> (not needed anymore)

kamronbatman:
## Summary

Removes `FastList<T>` and migrates the five consumers to modern .NET equivalents (`List<T>` + `CollectionsMarshal.AsSpan`, plain arrays, or a new ArrayPool-backed ref struct).

`FastList<T>` is a 2018-era thin wrapper over `T[]` with public mutable `Length` and `Buffer` fields. On net10 it offers no measurable perf advantage over `List<T>` + `CollectionsMarshal.AsSpan`, while being strictly worse:

- **Bug-prone API.** Public mutable `Length` lets callers silently violate `Length <= Buffer.Length` when assigned above current capacity. This is the bug fixed in #1891 (`Data.Length = CharCount` in `FontsLoader`'s wrap path triggering `IndexOutOfRangeException` in `RenderedText.DrawGlyphs`).
- **No version field** — concurrent mutation during iteration silently corrupts; `List<T>` throws.
- **No standard interfaces** — `IList<T>`/`IReadOnlyList<T>`/`IEnumerable<T>` not implemented.
- **Default capacity 5** — every empty `FastList` allocates a 5-element backing array even if never used.

## Migration strategy by consumer lifecycle

| Consumer | Pattern | Result |
|---|---|---|
| `AnimatedStaticsManager._staticInfos` | build-once at `Initialize`, iterate every `Process` forever | `StaticAnimationInfo[]` (built via `List` then `.ToArray()` once). Exact size, no spare-capacity overhead, ref iteration via array indexer. |
| `BoatMovingManager._items` | long-lived per-boat lists with mutation | `List<ItemInside>` + ref iteration via `CollectionsMarshal.AsSpan`. |
| `MultilinesFontInfo.Data` (`FontsLoader`) | hot-path glyph buffer, parsed and drawn frequently | `List<MultilinesFontData>` + ref iteration via `CollectionsMarshal.AsSpan`. The six `Data.Length = X` assignments now go through a private `SetDataCount` helper that grows with default-initialized slots (preserves the `countspaces` synthetic-+1 semantic) or shrinks via `CollectionsMarshal.SetCount`. **Absorbs and obsoletes #1891's `FastList.Resize` fix** — the bug sites are rewritten end-to-end. |
| `FontInfo.Links` (`FontsLoader`) | built once during HTML/Unicode parse, read-only thereafter | `WebLinkRect[]` (built locally via `List` then `.ToArray()`, or `null` if empty). |
| HTML parser `stack` (`FontsLoader.GetInfoHTML` local) | method-scoped temporary | New `PooledList<T>` ref struct that rents from `ArrayPool<T>.Shared` and returns on `Dispose`. Zero allocation per HTML parse after pool warm-up. `GetCurrentHTMLInfo` signature switched to `ReadOnlySpan<HTMLDataInfo>` since it only reads. |

## Why each shape was chosen

The user's framing in the design discussion: build-once → array; long-lived mutation → `List<T>` + `AsSpan`; method-scoped temp → ArrayPool.

- **Arrays for build-once data** keep the iteration tight with no spare-capacity overhead and let consumers continue to use `arr[i]` and `ref arr[i]` exactly as they did before.
- **`List<T>` + `CollectionsMarshal.AsSpan`** is the modern equivalent of FastList's "ref-by-index" pattern: identical perf, proper encapsulation, version detection, ecosystem compatibility.
- **`PooledList<T>`** (new ref struct in `ClassicUO.Utility.Collections`) wraps `ArrayPool<T>.Shared` with the minimal API that `FastList<T>` consumers need (`Add`, `Clear`, `RemoveAt`, indexer, `AsSpan`, `Count`). Use with `using var list = new PooledList<T>(initialCapacity);` to ensure the rented array returns to the pool. Reusable for any future method-scoped temporary buffer.

## Test plan

- [x] `dotnet build src/ClassicUO.Client/ClassicUO.Client.csproj` — clean
- [x] `dotnet test tests/ClassicUO.UnitTests/ClassicUO.UnitTests.csproj` — 172/172 pass
- [ ] In-game verification: open `[props self` (Label/Tooltip pipeline), books, journal, HTML signs/scrolls, sail a boat with cargo, observe animated statics (e.g., torches/forges)
- [ ] Reproduce the original `IndexOutOfRangeException` in `RenderedText.DrawGlyphs` from #1891 — should not occur

## Relationship to #1891

This PR is built on top of #1891 (`fix/fastlist-resize-invariant`). The merge order doesn't matter:

- **If #1891 lands first:** this PR's diff applies cleanly — `FastList.Resize` and the six `ptr.Data.Resize(...)` calls are deleted/replaced as part of the migration commit.
- **If this PR lands first:** #1891 becomes obsolete (its target file is gone) and can be closed.

Either way, the bug #1891 fixes is also fixed here, by construction — there is no public mutable `Length` to misuse.


andreakarasho (approved):

andreakarasho:
Thank you! There are conflict that need to be addressed before I can merge

kamronbatman:
@andreakarasho the merge conflicts are fixed.

PR #1899, opened 2026-05-22
Add 3-pixel forgiveness to backpack ItemGump pixel-pick

KonssnoK:
some items, like gm robes, dupre/lord british etc items, are currently not pickable when on the ground/in the backpack. this should address that.

### DIFF
--- a/src/ClassicUO.Client/Game/UI/Controls/ItemGump.cs
+++ b/src/ClassicUO.Client/Game/UI/Controls/ItemGump.cs
@@ -210,7 +210,10 @@ namespace ClassicUO.Game.UI.Controls
             }
             else
             {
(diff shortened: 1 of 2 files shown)


andreakarasho:
This will enlarge the radious of the picking area. I'm not sure will work good in containers full of items.
 I can pick the GM robe , maybe it's matter of pixel precision, but it's feasible
ClassicUO_sXdqNRAnBF
No issues with dupre suite.
ClassicUO_hK4BHypbYw

Tested on ModernUO.

I prefer to maintan the pixel perfect hit test honestly

KonssnoK:
can you try the same in the backpack ? 

i initially noticed this because i removed the gm robe and put it in the backpack

andreakarasho:
cuo_JSktdLrCdG

The picking still working fine in containers too

PR #1915, opened 2026-06-12
Revert "[statics] Fix same z rendering order"

andreakarasho:
Reverts ClassicUO/ClassicUO#1908

From discord:
image
image
image


KonssnoK:
ok that's strange, and those are clearly not OSI statics.
I guess those statics were made with tools, so the tools do not follow OSI statics rules

otherwise please give me the 
map position and i will check

PR #1930, opened 2026-08-29
Drain queued NPC steps at the server's cadence

kamronbatman:
### Summary

* **Steps drain at the server's measured cadence.** NPC steps are animated over a fixed nominal time selected by the run flag (400/200 ms on foot, 200/100 mounted). A server that steps on a different cadence renders badly either way: faster than nominal backs the step queue up until `EnqueueStep` drops packets and the mobile snaps forward; slower than nominal (e.g. a 300 ms cadence with the run flag) moves the mobile for 200 ms and stalls it for 100 ms, a visible stutter. This records the interval between consecutive *position-changing* step packets per mobile and animates each step over that interval whenever it lies in the band the client can follow — from a mounted run (100 ms) up to the walk time. Outside the band the nominal time for the flag applies, so slow wanderers keep the step-then-stand look. The player's own movement is untouched.
* **The backlog sync patch no longer fires on queued facing-turns.** The legacy compression heuristic keys on `Steps.Count > 1`, but a turn packet queued behind an in-flight step inflates the depth without any movement backlog — the arrival step of a chase (server turns the mobile to face its target) rendered in ~100 ms as a dart. Only entries that change position relative to their predecessor count as backlog now, and the cadence band is bounded by the mobile's real mounted/speed state rather than the one the heuristic fakes, so a measured cadence always outranks the guess. Genuine packet bursts still compress: their measured interval is itself small.
* **Animal (LOW group) bodies without a run animation group** fell back to their **stand** frames while moving, so they slid. They now use the walk group.

All changes are client-only: no protocol changes, no server-specific behavior.

### Testing

Verified with paired UTC-aligned traces (server step emission vs client `EnqueueStep`/`ProcessSteps`) plus frame-by-frame video tracking of a ModernUO lich chase (300 ms cadence):

* Before: the 200 ms-move / 100 ms-stall stutter mid-chase; ~100 ms dart on the arrival step whenever a facing-turn queued behind it.
* After: 158 of 159 rendered steps animate at the measured 250–350 ms cadence; the single fast render is the first step ever seen for the mobile (no interval measured yet, nominal-by-flag). Pets at ~100 ms and creatures at 125/300/450 ms cadences verified in-game.

### DIFF
--- a/src/ClassicUO.Client/Game/Data/MovementSpeed.cs
+++ b/src/ClassicUO.Client/Game/Data/MovementSpeed.cs
@@ -11,6 +11,12 @@ namespace ClassicUO.Game.Data
         public const int STEP_DELAY_RUN = 200;
         public const int STEP_DELAY_WALK = 400;
 
+        // Floor for a measured server cadence. Below this a step is a teleport, and a
+        // jitter burst (two packets a frame apart) must not collapse the steps it touches
+        // into one; above it every legitimate creature pace we know of fits (ModernUO
+        // floors creatures at 50 ms, UOX3 tunes to 80 ms, POL to 125 ms, Sphere to 100 ms).
+        public const int STEP_DELAY_MIN = 50;
+
         public static bool FastRotation;
         public static int TurnDelay => FastRotation ? Constants.TURN_DELAY_FAST : Constants.TURN_DELAY;
 
--- a/src/ClassicUO.Client/Game/Data/MovementSpeed.cs
+++ b/src/ClassicUO.Client/Game/Data/MovementSpeed.cs
@@ -11,10 +11,7 @@ namespace ClassicUO.Game.Data
         public const int STEP_DELAY_RUN = 200;
         public const int STEP_DELAY_WALK = 400;
 
-        // Floor for a measured server cadence. Below this a step is a teleport, and a
-        // jitter burst (two packets a frame apart) must not collapse the steps it touches
-        // into one; above it every legitimate creature pace we know of fits (ModernUO
-        // floors creatures at 50 ms, UOX3 tunes to 80 ms, POL to 125 ms, Sphere to 100 ms).
+        // Floor for a measured server cadence; below this a step reads as a teleport.
         public const int STEP_DELAY_MIN = 50;
 
         public static bool FastRotation;
--- a/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
+++ b/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
@@ -299,7 +299,7 @@ namespace ClassicUO.Game.GameObjects
             }
 
             Direction moveDir = DirectionHelper.CalculateDirection(endX, endY, x, y);
-            Step step = new Step();
+            Step step = new Step { Arrival = Time.Ticks };
 
             if (Serial != World.Player)
             {
@@ -729,10 +729,11 @@ namespace ClassicUO.Game.GameObjects
             LastAnimationChangeTime = Time.Ticks + currentDelay;
         }
 
-        // True when at least two position-changing steps are queued, counted from the
-        // tile the mobile stands on: the in-flight step plus a real move behind it.
-        // Facing-turns share their predecessor's tile and occupy depth without one.
-        private bool HasMoveBacklog()
+        // Milliseconds the first real move queued behind the in-flight step has waited,
+        // or -1 when there is no movement backlog. Counted from the tile the mobile
+        // stands on: the in-flight step plus a real move behind it. Facing-turns share
+        // their predecessor's tile and occupy depth without one.
+        private int QueuedMoveWait()
         {
             int moves = 0;
             int prevX = X;
@@ -746,7 +747,7 @@ namespace ClassicUO.Game.GameObjects
                 {
                     if (++moves > 1)
                     {
-                        return true;
+                        return (int)(Time.Ticks - s.Arrival);
                     }
 
                     prevX = s.X;
@@ -754,7 +755,7 @@ namespace ClassicUO.Game.GameObjects
                 }
             }
 
-            return false;
+            return -1;
         }
 
         public void ProcessSteps(out byte dir, bool evalutate = false)
@@ -788,13 +789,15 @@ namespace ClassicUO.Game.GameObjects
                     bool mounted = actuallyMounted;
                     bool run = step.Run;
 
+                    int backlogWait = Serial != World.Player && Steps.Count > 1 ? QueuedMoveWait() : -1;
+
                     // Client auto movements sync.
                     // When server sends more than 1 packet in an amount of time less than 100ms if mounted (or 200ms if walking mount)
                     // we need to remove the "teleport" effect.
                     // When delay == 0 means that we received multiple movement packets in a single frame, so the patch becomes quite useless.
                     // Only a genuine movement backlog may compress; a queued facing-turn
                     // occupies depth without one.
-                    if (!mounted && Serial != World.Player && Steps.Count > 1 && delay > 0 && HasMoveBacklog())
+                    if (!mounted && delay > 0 && backlogWait >= 0)
                     {
                         mounted =
                             delay
@@ -819,7 +822,11 @@ namespace ClassicUO.Game.GameObjects
                         && StepInterval <= MovementSpeed.TimeToCompleteMovement(false, actuallyMounted)
                     )
                     {
-                        stepTime = Math.Max(StepInterval, MovementSpeed.STEP_DELAY_MIN);
+                        // A move waiting behind this step is a tile the server has that we have
+                        // not shown yet. Shorten the step by the wait so the queued move starts
+                        // on time: a packet a few ms early costs a few ms, a full packet of lag
+                        // runs the tail at double speed and halves the deficit every step.
+                        stepTime = Math.Max(StepInterval - Math.Max(backlogWait, 0), MovementSpeed.STEP_DELAY_MIN);
                     }
 
                     int maxDelay = Math.Max(stepTime - (int)Client.Game.FrameDelay[1], 1);
@@ -1165,6 +1172,7 @@ namespace ClassicUO.Game.GameObjects
             public sbyte Z;
             public byte Direction;
             public bool Run;
+            public uint Arrival;
         }
     }
 }
--- a/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
+++ b/src/ClassicUO.Client/Game/GameObjects/Mobile.cs
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: BSD-2-Clause
+﻿// SPDX-License-Identifier: BSD-2-Clause
 
 using ClassicUO.Assets;
 using ClassicUO.Configuration;
@@ -126,6 +126,7 @@ namespace ClassicUO.Game.GameObjects
         // Lets queued steps drain at the server's cadence instead of the nominal step time.
         public int StepInterval { get; private set; }
         private uint _lastStepArrival;
+        private bool _turnSinceLastStep;
         public bool IsParalyzed => (Flags & Flags.Frozen) != 0;
         public bool IsYellowHits => (Flags & Flags.YellowBar) != 0;
         public bool IsPoisoned =>
@@ -300,20 +301,38 @@ namespace ClassicUO.Game.GameObjects
             Direction moveDir = DirectionHelper.CalculateDirection(endX, endY, x, y);
             Step step = new Step();
 
-            if (moveDir != Direction.NONE)
+            if (Serial != World.Player)
             {
-                // Only packets that move the mobile define its cadence; a turn packet
-                // followed by a step would otherwise read as a near-zero interval.
-                if (Serial != World.Player)
+                if (moveDir == Direction.NONE)
+                {
+                    // A facing-turn is not a step: it neither defines the cadence nor
+                    // resets the clock the next step is measured against.
+                    _turnSinceLastStep = true;
+                }
+                else
                 {
                     if (_lastStepArrival != 0)
                     {
-                        StepInterval = (int)(Time.Ticks - _lastStepArrival);
+                        int elapsed = (int)(Time.Ticks - _lastStepArrival);
+
+                        // Servers that emit turns as their own packet pace the step after a
+                        // turn by the turn latency, not the step period (the client itself
+                        // steps 80 ms after a turn; some emulators charge a full period), so a
+                        // pair that spans a turn keeps the last estimate while it is still
+                        // renderable. Beyond a walk, measure: the mobile actually paused.
+                        if (!_turnSinceLastStep || elapsed > MovementSpeed.STEP_DELAY_WALK)
+                        {
+                            StepInterval = elapsed;
+                        }
                     }
 
                     _lastStepArrival = Time.Ticks;
+                    _turnSinceLastStep = false;
                 }
+            }
 
+            if (moveDir != Direction.NONE)
+            {
                 if (moveDir != endDir)
                 {
                     step.X = endX;
@@ -710,22 +729,29 @@ namespace ClassicUO.Game.GameObjects
             LastAnimationChangeTime = Time.Ticks + currentDelay;
         }
 
-        // True when any entry queued behind the in-flight front step is an actual move.
+        // True when at least two position-changing steps are queued, counted from the
+        // tile the mobile stands on: the in-flight step plus a real move behind it.
         // Facing-turns share their predecessor's tile and occupy depth without one.
         private bool HasMoveBacklog()
         {
-            Step prev = Steps[0];
+            int moves = 0;
+            int prevX = X;
+            int prevY = Y;
 
-            for (int i = 1; i < Steps.Count; i++)
+            for (int i = 0; i < Steps.Count; i++)
             {
                 Step s = Steps[i];
 
-                if (s.X != prev.X || s.Y != prev.Y)
+                if (s.X != prevX || s.Y != prevY)
                 {
-                    return true;
-                }
+                    if (++moves > 1)
+                    {
+                        return true;
+                    }
 
-                prev = s;
+                    prevX = s.X;
+                    prevY = s.Y;
+                }
             }
 
             return false;
@@ -781,18 +807,19 @@ namespace ClassicUO.Game.GameObjects
 
                     int stepTime = MovementSpeed.TimeToCompleteMovement(run, mounted);
 
-                    // Animate at the server's cadence while it is renderable (a mounted run
-                    // up to a walk): faster than nominal backs the queue up until steps drop
-                    // and the mobile snaps; slower than nominal stalls between steps. Beyond
-                    // a walk, keep the usual step-then-stand look. Bounded by the real
-                    // mounted state so the measured cadence outranks the backlog heuristic.
+                    // Animate at the server's cadence while it is renderable (the fastest
+                    // tuned creatures up to a walk): faster than nominal backs the queue up
+                    // until steps drop and the mobile snaps; slower than nominal stalls
+                    // between steps. Beyond a walk, keep the usual step-then-stand look.
+                    // Bounded by the real mounted state so the measured cadence outranks
+                    // the backlog heuristic.
                     if (
                         Serial != World.Player
                         && StepInterval > 0
                         && StepInterval <= MovementSpeed.TimeToCompleteMovement(false, actuallyMounted)
                     )
                     {
-                        stepTime = Math.Max(StepInterval, MovementSpeed.STEP_DELAY_MOUNT_RUN);
+                        stepTime = Math.Max(StepInterval, MovementSpeed.STEP_DELAY_MIN);
                     }
 
                     int maxDelay = Math.Max(stepTime - (int)Client.Game.FrameDelay[1], 1);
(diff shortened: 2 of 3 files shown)


Xoduz:
## Suggested Refinements to PR #1930: Movement Cadence and Queue Handling

After extensive testing of this PR on UOX3 and after evaluating it (with some help) across the source repos for UOX3, ModernUO, SphereServer and POL, I've documented three specific issues encountered, along with the changes made in `Mobile.cs` to resolve them. Only UOX3 has been specifically tested with/without the changes.

---

### 1. Cadence Measurement on Turn Packets
* **Problem:** In PR 1930, step interval tracking was placed inside `if (moveDir != Direction.NONE)`:
  ```csharp
  // Only packets that move the mobile define its cadence; a turn packet
  // followed by a step would otherwise read as a near-zero interval.
  if (moveDir != Direction.NONE)
  ```
  In ModernUO (`AIMovement.cs:132-140`), direction changes and movement steps occur in the same server tick (`Mobile.Direction = d;` followed by `TryMove(d)`), so the direction packet and step packet arrive at the client in the same network batch (`delta == 0`).

  In **UOX3**, **SphereServer**, and **POL**, observed player direction changes (and NPC turns in UOX3 pathfinding) occur on separate ticks (`delta > 0`). Because turn packets were ignored by `moveDir != Direction.NONE`, the subsequent movement step measured elapsed time across both the turn and the step (~350–500ms). This inflated value caused an artificial slow-motion stride after turning, or exceeded `STEP_DELAY_WALK` (400ms), dropping cadence tracking entirely.

* **Fix:** Move cadence tracking outside `if (moveDir != Direction.NONE)` and filter out same-frame packet arrivals (`delta > 0`):
  ```csharp
  if (Serial != World.Player)
  {
      if (_lastStepArrival != 0)
      {
          int delta = (int)(Time.Ticks - _lastStepArrival);

          if (delta > 0)
          {
              if (delta <= MovementSpeed.STEP_DELAY_WALK)
              {
                  StepInterval = delta;
              }

              _lastStepArrival = Time.Ticks;
          }
      }
      else
      {
          _lastStepArrival = Time.Ticks;
      }
  }

  if (moveDir != Direction.NONE)
  {
  ```
  * In **ModernUO**, movement packets arriving in the same frame as turn packets have `delta == 0` and do not overwrite `StepInterval`.
  * In **UOX3**, **SphereServer**, and **POL**, turn packets arriving on separate ticks (`delta > 0`) update `_lastStepArrival`, allowing the subsequent movement step to measure its true cadence without post-turn slow-motion distortion.

---

### 2. False Movement Backlog from In-Place Turns (`HasMoveBacklog`)
* **Problem:** To prevent facing-turns from being mistaken for a lag-induced packet queue, PR 1930 introduced `HasMoveBacklog()` to ensure only entries queued behind an in-flight front step that are actual moves count as backlog. The implementation initializes its coordinate check against `prev = Steps[0]` and returns `true` as soon as any subsequent step differs in position from its predecessor.

  While this check works when a facing-turn is queued behind a movement step (`[Move, Turn]`, such as an NPC's chase arrival step facing the target), it fails whenever an entity turns **before** moving (`[Turn, Move]`):
  * `Steps[0]` is the in-flight **facing-turn** at coordinate `(X, Y)`.
  * `Steps[1]` is the **first and only** actual movement step to `(X+dx, Y+dy)`.

  Because `Steps[1]` has different coordinates than `Steps[0]`, `HasMoveBacklog()` immediately returns `true` on index 1, incorrectly classifying a single move step preceded by a facing-turn as a backlog.

  This occurs whenever an entity changes direction before stepping:
  * In **ModernUO**, when an observed remote player turns and then steps, or an entity changes facing before stepping, queuing the turn at `Steps[0]` and the step at `Steps[1]`.
  * In **UOX3**, when an observed remote player turns and steps around a corner, or an NPC in pathfinding (`WT_PATHFIND` in `movement.cpp:2627-2633`) executes a direction change before taking a step.
  * In **SphereServer**, when an NPC interacts with a door (`CCharNPCAct.cpp:988-998`), emitting a turn packet followed by a step packet in the same tick (`delta == 0`).
  * In **POL**, when AI scripts call `TurnToward()` followed immediately by `RunToward()`.

* **Consequence:** 
  * Because `HasMoveBacklog()` returns `true` when only a single movement step follows a turn, it falsely engages mounted speed compression (`mounted = true`), halving step duration and causing an unnatural acceleration burst whenever an observed player or creature changes direction.

* **Fix:** Track position changes relative to the entity's current world coordinates `(X, Y)`:
  ```csharp
  // True when there is an actual movement backlog (at least two position-changing steps queued).
  // Facing-turns share their predecessor's tile and must not be counted as a movement backlog.
  private bool HasMoveBacklog()
  {
      int moveCount = 0;
      int prevX = X;
      int prevY = Y;

      for (int i = 0; i < Steps.Count; i++)
      {
          Step s = Steps[i];

          if (s.X != prevX || s.Y != prevY)
          {
              moveCount++;
              prevX = s.X;
              prevY = s.Y;
          }
      }

      return moveCount > 1;
  }
  ```
  * In-place turns share `(X, Y)` with current entity coordinates and do not increment `moveCount`.
  * Backlog acceleration only engages when at least two actual coordinate-changing steps are queued, eliminating false turn sprints across ModernUO, UOX3, Sphere, and POL.

---

### 3. Minimum Step Time Clamping on Fast Speeds
* **Problem:** `stepTime` was clamped to a minimum of `MovementSpeed.STEP_DELAY_MOUNT_RUN` (100ms).
* **Consequence:** 
  * In **POL**, default mounted running delay is 80ms (`PKTIN_02_MOUNT_RUN 80`).
  * In **UOX3**, observed mounted running delay is ~90ms, and custom NPC running speeds can be tuned under 100ms (e.g. 80ms, or follow-boosted pets).
  * Clamping `stepTime` to 100ms forced the client to animate slower than the server was sending steps (80–90ms). The step queue accumulated a deficit on every step, leading to dropped steps and visible teleport snaps.
* **Fix:** Clamp `stepTime` to `Client.Game.FrameDelay[1]`:
  ```csharp
  // Animate at the server's cadence while it is renderable (from fast tuned speeds
  // up to a walk): faster than nominal backs the queue up until steps drop
  // and the mobile snaps; slower than nominal stalls between steps. Beyond
  // a walk, keep the usual step-then-stand look.
  if (
      Serial != World.Player
      && StepInterval > 0
      && StepInterval <= MovementSpeed.STEP_DELAY_WALK
  )
  {
      stepTime = Math.Max(StepInterval, (int)Client.Game.FrameDelay[1]);
  }
  ```
  * Clamping to `Client.Game.FrameDelay[1]` (~8–16ms) cleanly accommodates POL's 80ms mounted run, UOX3's 90ms mounted run, and fast-tuned NPCs across all emulators without queue backlogs or teleport snaps, while ensuring `maxDelay = Math.Max(stepTime - (int)Client.Game.FrameDelay[1], 1);` cannot underflow below zero.

---

## Cross-Emulator Impact Assessment

> [!NOTE]
> All references to "players" in these assessments refer specifically to **observed remote players** on screen (`Serial != World.Player`). The local player (`Serial == World.Player`) uses client-side prediction in `PlayerMobile.cs` / `Walker.cs`, which is unaffected by `StepInterval` and `HasMoveBacklog()`.

To ensure these refinements are server-agnostic and do not regress behavior across different UO server implementations, the movement and packet timing engines of **SphereServer (Source-X)**, **POL**, and **UOX3** were analyzed and evaluated against Vanilla ClassicUO, PR 1930 as proposed, and our proposed amendment.

---

<details>
    <summary>Expand to reveal Detailed Assessment</summary>

  ### SphereServer (Source-X Branch)

  #### Movement Architecture
  * **Clock & Timing:** Operates on 100ms tick slices (`TICKS_PER_SEC = 10`, `MSECS_PER_TICK = 100`, `MSECS_PER_TENTH = 100`).
  * **NPC Movement Speeds (`CCharNPCAct.cpp:610-691`):**
    * **Walking:** `1000ms + rand((100 - DEX)/3) * 100ms` (1000ms at 100 DEX, ~1750ms at 50 DEX). Sphere NPCs walk significantly slower than standard OSI (1.0s+ per step).
    * **Running:** `250ms + rand((100 - DEX)/5) * 100ms` (250ms at 100 DEX, ~450ms at 50 DEX).
    * **Clamps:** Hard-clamped between 100ms (`MSECS_PER_TENTH`) and 5000ms in modern style; old-style/scripted movers can exceed 100ms.
    * **Wander Pauses:** In `NPC_Act_Wander()`, `uiRand % (7 + DEX/30)` triggers `NPC_Act_Idle()`, pausing for 1000ms to 3000ms (`_SetTimeoutS(1 + rand(2))`).
  * **Packet Emission (0x77 `PacketCharacterMove`):**
    * **NPC Steps:** `NPC_WalkToPoint()` directly updates `m_dirFace = Dir` and moves `MoveToChar()`, emitting only the movement packet.
    * **NPC Door Opening (`CCharNPCAct.cpp:988-998`):** Calls `UpdateDir(door)` followed immediately by `NPC_WalkToPoint()` in the **same tick** (`delta == 0`).
    * **Observed Player Movement (`CClientEvent.cpp:958-965`):** Direction change emits 0x77 with current coordinates (`moveDir == NONE`), followed by step 0x77 (`moveDir != NONE`) when the next walk request arrives.

  #### Comparative Assessment
  * **Vanilla ClassicUO:**
    * Walking NPCs (1000ms+) take a 400ms step and pause for 600ms+ (classic stop-and-go).
    * Running NPCs (250ms at 100 DEX) animate at fixed 200ms, creating a slight 50ms stutter between steps.
    * In-place turns (door opening, observed player turns) falsely trigger mounted speed compression if `Steps.Count > 1`.
  * **PR 1930 As Proposed:**
    * Walking NPCs correctly fall back to nominal 400ms (1000ms exceeds 400ms cadence band).
    * Running NPCs at 250ms smoothly scale to 250ms, eliminating the 50ms stutter.
    * **Distorted Cadence on Turns:** Because cadence tracking is inside `moveDir != NONE`, turn packets (`moveDir == NONE`) are ignored. When an observed player or NPC turns and then runs, `delta` measures across both the pre-turn step and the turn, inflating to ~350ms slow-motion or exceeding 400ms and dropping cadence matching.
    * **False Acceleration on Door Opening:** An NPC opening a door emits turn (0x77) + step (0x77) in the same tick. `HasMoveBacklog()` checks `Steps[1]` vs `Steps[0]`, sees different coordinates, and falsely accelerates the in-place turn at mounted speed.
    * **100ms Min Clamp:** Clamping to 100ms causes fast-tuned or jittered packets (70–90ms) to outpace animation, backing up the queue and causing teleport snaps.
  * **With Our Amendment (Points 1–3):**
    * Point 1 tracks cadence outside `moveDir != NONE` with `delta > 0`: turn arrival timestamps are recorded, so post-turn steps measure their true 200–250ms delta without slow-motion distortion. Same-frame packets (`delta == 0`) from door opening are cleanly ignored.
    * Point 2 evaluates `HasMoveBacklog()` relative to current entity `(X, Y)`: in-place turns do not move away from `(X, Y)`, preventing false acceleration on door opening.
    * Point 3 clamps to `FrameDelay[1]` (~8–16ms): cleanly accommodates fast-tuned or jittered sub-100ms arrivals without queue backlog.

  ---

  ### POL Server

  #### Movement Architecture
  * **Clock & Timing Defaults (`pktdef.h:70-74`):**
    * Observed mounted run delay: **80ms** (`#define PKTIN_02_MOUNT_RUN 80`)
    * Observed mounted walk delay: **180ms**
    * Observed foot run delay: **180ms**
    * Observed foot walk delay: **380ms**
  * **NPC Movement Speeds (`npcmod.cpp:226-231`):**
    * `delay = std::max(1000 - npc.run_speed * 3, ssopt.npc_minimum_movement_delay)` (minimum delay defaults to 250ms).
    * `os_module->SleepForMs(run ? (sleep / 2) : sleep)`.
    * **Walking:** 700ms to 925ms for typical Dex (25–100).
    * **Running:** 350ms to 462ms for typical Dex (25–100); fast/boss NPCs can drop to 125ms or lower.
  * **Packet Emission (0x77 `MoveChrPkt`):**
    * **NPC Steps:** `Character::move()` updates facing and position simultaneously for NPCs (`script_isa(POLCLASS_NPC)` is true), emitting a single 0x77 packet.
    * **AI Script Turns (`TurnToward` / `Face`):** Dispatches 0x77 with current `(X, Y)` via `on_facing_changed()` and returns immediately without sleeping. When followed by `RunToward()`, both turn and step packets leave the server in the **exact same frame** (`delta == 0`).
    * **Observed Player Movement (`movement.cpp:170-218`):** Player direction changes emit 0x77 at current `(X, Y)` (`moveDir == NONE`), followed by step 0x77 (`moveDir != NONE`) at server cadence (~80ms on mounts, ~180ms on foot).

  #### Comparative Assessment
  * **Vanilla ClassicUO:**
    * Walking NPCs (700–925ms) step at 400ms and pause for 300–525ms.
    * Running NPCs (350ms at Dex 100) animate at 200ms, creating a prominent 150ms stutter between every running step.
    * Observed mounted running (80ms): Vanilla mounted run is 100ms; queue can accumulate.
    * In-place turns: `Steps.Count > 1` triggers false mounted compression in vanilla.
  * **PR 1930 As Proposed:**
    * Walking NPCs fall back to nominal 400ms (700ms+ exceeds 400ms cadence band).
    * Running NPCs at 350ms smoothly scale to 350ms, eliminating the 150ms stutter.
    * **Critical Regression on Observed Mounted Run (80ms):** POL sets mounted running cadence to 80ms. PR 1930 hardcodes `stepTime = Math.Max(StepInterval, 100)`. Packets arrive every 80ms, but animations take 100ms. The client falls behind by 20ms every step, the queue backs up, and the observed character teleports / snaps!
    * **Distorted Cadence on Observed Player Turns:** Player turn (0x77) is skipped by `moveDir != NONE`. The subsequent step measures `(pre-turn step + 80ms turn + 180ms step)` = ~360–440ms, causing an artificial slow-motion stride or breaking cadence entirely.
    * **False Acceleration on `TurnToward` + `RunToward`:** AI script turns followed by steps emit turn + step in the same frame. PR 1930's `HasMoveBacklog()` checks `Steps[1].X != Steps[0].X` (`true`), falsely accelerating the in-place turn at mounted speed.
  * **With Our Amendment (Points 1–3):**
    * Point 1 records turn timestamps, ensuring post-turn steps measure true 180–200ms intervals without slow-motion post-turn strides, and ignores same-frame (`delta == 0`) bursts.
    * Point 2 evaluates backlog relative to `(X, Y)`, preventing false acceleration on `TurnToward` in AI scripts.
    * Point 3 clamps to `FrameDelay[1]` (~8–16ms), natively accommodating POL's 80ms mounted run without queue backlog or teleport snaps.

  ---

  ### UOX3 Movement & Turning Assessment

  #### Movement Architecture
  * **Observed Player Movement Cadence (`movement.cpp:78-80`):**
    * The server processes and broadcasts player movement at nominal cadences:
      * Foot walk: ~390ms
      * Foot run: ~190ms
      * Mounted run: ~90ms
  * **Observed Player Turning (`movement.cpp:265, 462-469`):**
    * Turning in place (`amTurning == ((dir & 7) != c->GetDir())`) updates direction only, emitting 0x77 at current `(X, Y)` (`moveDir == NONE`).
    * The subsequent step emits 0x77 at new `(X+dx, Y+dy)` (`moveDir != NONE`) at `delta > 0`.
  * **Dynamic / Customizable NPC Movement (`movement.cpp:2264, 2950-3002`):**
    * NPC speeds are entirely dynamic and customizable per creature:
      * Driven by `mChar.GetWalkingSpeed()`, `mChar.GetRunningSpeed()`, `GetMountedRunningSpeed()`, etc., setting the `tNPC_MOVETIME` timer.
      * Default walk speed is 0.6s (600ms); default run speed is 0.2s (200ms).
      * Following pets/escorts dynamically increase speed (`GetRunningSpeed() / 1.5`, e.g. 133ms or 80ms).
      * Custom speeds configured in DFNs or scripts can set arbitrary rates (e.g. 0.08s = 80ms for fast mounts/monsters).
      * During pathfinding (`WT_PATHFIND`, lines 2627-2633), turns push the direction back onto the queue, consuming a dedicated `tNPC_MOVETIME` tick before the step.

  #### Comparative Assessment
  * **Vanilla ClassicUO:**
    * Straight-line running for observed players and 200ms NPCs animates at nominal 200ms.
    * Fast-tuned NPCs (< 200ms) or follow-boosted pets cannot keep up unless vanilla's multi-step compression engages.
    * In-place turns immediately followed by a step trigger vanilla's `Steps.Count > 1` check, falsely applying mounted speed compression (`mounted = true`) to unmounted characters.
  * **PR 1930 As Proposed:**
    * Straight-line running at 190–200ms tracks smoothly.
    * **Distorted Cadence on Turn-Then-Run:** When an observed player turns and runs, the turn packet (0x77) is ignored because `moveDir == NONE`. The subsequent step measures `delta = (pre-turn step + turn delay + step time)` = ~380–500ms. The observed running player suddenly takes a 380ms slow-motion stride upon turning, or drops cadence tracking if delta exceeds 400ms.
    * **False Backlog on Turns:** When an in-place turn and step are queued, `HasMoveBacklog()` checks `Steps[1].X != Steps[0].X` (`true`), falsely compressing the turn at mounted speed.
    * **Fast NPCs & Observed Mounted Gallop Deficit (< 100ms):**
      * Both observed mounted players (~90ms) and fast-tuned NPCs (e.g. 80ms or follow-accelerated creatures) arrive faster than 100ms.
      * PR 1930 clamps `stepTime = Math.Max(StepInterval, 100)`.
      * Animations take 100ms while packets arrive every 80–90ms. The client accumulates a deficit on every step until the entity teleports / snaps!
  * **With Our Amendment (Points 1–3):**
    * Point 1 tracks cadence on all packets where `delta > 0`. The turn timestamp is recorded, so the post-turn step accurately measures ~190ms. No slow-motion distortion around corners.
    * Point 2 checks backlog relative to `(X, Y)`. In-place turns do not move away from `(X, Y)` and never trigger false mounted acceleration.
    * Point 3 clamps to `FrameDelay[1]`, seamlessly handling both observed mounted player runs (90ms) and fast-tuned/follow-boosted NPCs (e.g. 80ms) without queue backlog or teleport snaps.

</details>

kamronbatman:
Went through all three points against the POL, Source-X and UOX3 sources, plus what the client itself does when it turns. Pushed refinements to the branch; details per point below.

### 1. Cadence Measurement on Turn Packets
* **Agree on the symptom, not the cause or the fix.** A step measured across a turn does read the period plus the turn latency. But this is not server-specific, and ModernUO is not the exception: the client's own walker sends the step after a turn on the turn delay (`Constants.TURN_DELAY`, 80 ms; 45 ms with fast rotation), and ModernUO acks a turn at zero cost and broadcasts it immediately. So an observed *player* produces `[turn, step 80 ms later]` on every server, ModernUO included. The `delta == 0` case only applies to ModernUO NPCs, where the direction delta is a duplicate packet `EnqueueStep` already drops.
* **Why the proposed fix regresses:** recording the turn's arrival and measuring the next step against it makes the 80 ms turn-to-step gap the cadence. Every post-turn step for an observed player would render as an 80 ms dart, on UOX3 too. It looked right in NPC testing because a UOX3 pathfinding NPC is charged a full `tNPC_MOVETIME` period for the turn, so turn-to-step there happens to equal the period.
* **What's on the branch:** a pair of steps that spans a turn keeps the previous estimate while the elapsed time is still within the walk band. Beyond a walk it measures, so a mobile that genuinely paused still resumes step-then-stand. Covers observed players everywhere, ModernUO NPC facing-turns, UOX3 pathfinding turns, the Sphere door case and the POL `TurnToward` + `RunToward` case.

### 2. False Movement Backlog from In-Place Turns
* **This scenario can't trigger.** A facing-turn at the front of the queue is removed unconditionally (`directionChange`), and `ProcessSteps` recurses in the same frame, so the backlog check is re-run with the turn gone and only one move queued. On the turn's own frame the `mounted` result is computed but never used: removal is forced and the offset math is skipped.
* **Took the rewrite anyway.** Counting from the mobile's tile is behaviorally equivalent and makes the comment true without relying on the same-frame pop. It now also returns how long the queued move has waited, which point 4 uses.

### 3. Minimum Step Time Clamping on Fast Speeds
* **The 80/90 ms figures are not cadences.** POL's `PKTIN_02_MOUNT_RUN 80` is `SpeedHack_MountRunDelay`, the minimum accepted interval on an incoming 0x02; UOX3's figures are the same kind of tolerance. A legitimate client still steps at 100/200/400, so only a speed-hacked client would be observed at 80 ms. Observed mounted players are not a regression case.
* **NPCs are a real case.** UOX3 DFN speeds and ModernUO's creature budget (floored at 50 ms) do legitimately step under 100 ms, so the floor was too high for them.
* **Why not `FrameDelay`:** two packets arriving one frame apart from network jitter would collapse the in-flight step and the next into an instant two-tile teleport. The floor is now `STEP_DELAY_MIN = 50`: below that a step is a teleport anyway, and every legitimate creature pace fits above it.

### 4. Draining a Backlog (new)
* **Problem:** with the measured cadence outranking the legacy sync patch, any step that once rendered longer than the server's period left the queue one packet behind for the rest of the run. The client already shows a mobile one step behind the server by construction; a second packet puts it a full tile behind, which is a visible range/LOS mismatch for anyone watching.
* **Fix:** each queued entry carries its arrival time. While a real move waits behind the in-flight step, the step is shortened by that wait so the queued move starts on time. A packet a few ms early costs a few ms; a full packet of lag runs the tail at double speed and the deficit halves every step until it is gone.

kamronbatman:
@andreakarasho - what do we need to do to get this over the finish line?
@Xoduz - I think you had some more suggestions? Waiting on you for this still?

Xoduz:
> @Xoduz - I think you had some more suggestions? Waiting on you for this still?

I do. I'm trying to finalize that suggestion as we speak. Sorry for the delay, bunch of stuff going on.

PR #1931, opened 2026-09-05
Read the map patch counts in the order servers send them

Pactor:
The 0xBF 0x18 packet carries, per facet, a count of patched statics blocks and a count of patched land blocks. This read land first. RunUO, ServUO and ModernUO all write statics first - ModernUO in
OutgoingMapPackets.SendMapPatches - so the two counts arrive swapped.

The failure is silent, which is why it can sit here unnoticed. A shard that patches land and not statics sends (0, n). Read the wrong way round that is "no land patches" plus n statics patches, and those n then clamp to zero against an empty stadifl. Nothing throws, nothing is logged, and the ground stays exactly as it was while the server is certain it changed - so the server and the client disagree about the world with no symptom beyond players walking through scenery that is not there, or into scenery that is.

Found by patching a facet and watching the server report 52 patched land blocks while the client drew none of them.

### DIFF
--- a/src/ClassicUO.Assets/MapLoader.cs
+++ b/src/ClassicUO.Assets/MapLoader.cs
@@ -461,10 +461,23 @@ namespace ClassicUO.Assets
                     continue;
                 }
 
-                int mapPatchesCount = (int)reader.ReadUInt32BE();
-                MapPatchCount[i] = mapPatchesCount;
+                // Statics first, then land.
+                //
+                // This was the other way round, and it silently disabled land patching against
+                // every server that sends this packet. RunUO, ServUO and ModernUO all write the
+                // statics count first - ModernUO in OutgoingMapPackets.SendMapPatches - so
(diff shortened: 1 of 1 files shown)


andreakarasho:
Oh that's a good catch! Maybe worth to fix this  too https://docs.polserver.com/packets/index.php?Packet=0xBF

andreakarasho (approved):

Pactor:
I have written https://github.com/Pactor/ModernUOIDE which needs all 5 of
my PR's updated to work

Had a look through the rest of 0xBF against what ModernUO actually sends —
0x10, 0x19, 0x1B and 0x22 all read correctly, so 0x18 seems to have been
the only one. Two small things if you want them: Send_CastSpell reads
GetPacketLength(ID) when it may be sending 0x12 (no effect today, both are
-1), and the polserver page has 0x22's first field as BYTE[2] when it's one
byte.

PR #1933, opened 2026-09-06
Size a tiled button to the area it tiles, and centre its item

Pactor:
Two bugs in ButtonTileArt, both about the last two arguments of the
gump command - the width and height of the region the art covers.

Size. Button's constructor sizes itself from its gump art, which is
right for a button whose art is its face and wrong for this one: the
whole point of buttontileart is that a small piece of art covers a
large region. Left at the art's size, ContainsByBounds tests a box the
size of one tile, so a list row two hundred pixels wide made from an
eight pixel strip is clickable for eight pixels of it.

The rest is not merely dead - it falls through to the gump behind, and
a gump is draggable, so pressing the middle of a row or a slider picks
the window up and moves it. That is how this was found.

Placement. The item was drawn at x + _tileX, y + _tileY - offset by the
button's own size. In a grid of cells that puts every item in the next
cell down and to the right, and the last column's items outside the
panel altogether. An item drawn at its container's size is drawn past
its far corner, so this cannot have been meant.

To see it: send a buttontileart with a width and height much larger
than its art, and click the far side of it.

### DIFF
--- a/src/ClassicUO.Client/Game/UI/Controls/ButtonTileArt.cs
+++ b/src/ClassicUO.Client/Game/UI/Controls/ButtonTileArt.cs
@@ -4,6 +4,7 @@ using ClassicUO.Game.Scenes;
 using ClassicUO.Renderer;
 using ClassicUO.Utility;
 using Microsoft.Xna.Framework;
+using System;
 using System.Collections.Generic;
 
 namespace ClassicUO.Game.UI.Controls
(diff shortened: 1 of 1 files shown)


andreakarasho:
Could you do a comparision between CUO and CC?

Pactor:
I am going to assume you mean classic client, login encryption is a stream
xor, if the official client connects unencrypted, a proxy can view it all
in plantext, and do you mean 0xBF for classic ? I can set up a logger and
run it, but not sure how much info that will give me

andreakarasho:
You can use any 3th party assistant to connect to any server and check graphically the classic client (CC)

Pactor:
well, this took awhile, Had a proper look at 0xBF. Two things.

The docs are wrong on 0x22. The page says the first field is BYTE[2]; it's
one byte. ModernUO writes (ushort)11 as the length then seven bytes of
payload — two bytes there would make 12. ClassicUO's Skip(1) is correct, so
anyone "fixing" it to match the page would break damage numbers.

And a real difference between CUO and CC. The Classic Client sends 0xBF
0x24 as a heartbeat — 6 bytes, one byte of payload, every ~3.6s (2.6–5.1s
range), independent of what the player is doing. Measured over 158 seconds
through a proxy. ClassicUO never sends it and has no case 0x24, so an
inbound one hits Log.Warn. That's the subcommand the page lists as "Unknown
(UOSE introduced)" — it can now say what it is.

Nothing observably broke without it, so whether it matters is your call.
But it was invisible from either codebase.

andreakarasho (approved):

Issue #1771, opened 2025-03-02
[Feature Request] Walking through people and mobs, and verdata.mul


Shelushun:
1. Please add a feature that allows to pass through mobs and NPCs. 
2. Allow the modified verdata.mul to be used on the original client

bittiez:
> 1. Please add a feature that allows to pass through mobs and NPCs.
> 2. Allow the modified verdata.mul to be used on the original client

1. Is a server side request, not client side

2. I think you can use verdata, you need to enable it in your `settings.json` file I believe

Shelushun:
1. Not quite. If you use Yoko Injection, then there is a setting that allows you to pass through people and mobs.

![Image](https://github.com/user-attachments/assets/33e4cd8b-de4e-4c0a-a50a-5d003cd1c7c7)

2. I'll check it out and come back with an answer.
upd
I found the "use_verdata" setting: true
And what command can I use to specify the path to the verdict file?

bittiez:
> 1. Not quite. If you use Yoko Injection, then there is a setting that allows you to pass through people and mobs.
> 
> ![Image](https://github.com/user-attachments/assets/33e4cd8b-de4e-4c0a-a50a-5d003cd1c7c7)
> 
> 2. I'll check it out and come back with an answer.
>    upd
>    I found the "use_verdata" setting: true
>    And what command can I use to specify the path to the verdict file?

I forgot yoko injection exists, and I forgot I think there is something in CUO that checks your stamina and stops you from walking through people if your low on stamina or something like that, I don't think there is an option built in though

2. I think it should just read the verdata inside the same location as the rest of your uo data files

jabinb:
Seems more like an assistant feature (intercepting player flags to add `IgnoreMobiles`). 

Personally I think it's not very useful as you'll just get rubber-banding due to the shard rejecting the move (like ServUO).
Unsure what the context is but if you're the shard owner you can just set all players to have `IgnoreMobiles`.

Shelushun:
> I forgot yoko injection exists, and I forgot I think there is something in CUO that checks your stamina and stops you from walking through people if your low on stamina or something like that, I don't think there is an option built in though
> 
> 2. I think it should just read the verdata inside the same location as the rest of your uo data files

1. All right. If your STM is <DEX, then you won't be able to pass through mobs. I want to fix it.
2. If I run the game through Yoko Injection or the client, then everything is readable. If I run it through ClassicUO, then there is no reading

[...]

PR #1938, opened 2026-09-21
Apply verdata.mul animation patches (FileID 6)

sharabdin1988:
# Apply `verdata.mul` animation patches (FileID 6)

Fixes the animation half of #1771: a shard can ship custom animations (mounts,
creatures, clothing) **only** as `verdata.mul` patches, and ClassicUO renders
those mobiles without them — while the original client (and any client that
reads `verdata.mul`) shows them normally.

## Why it does not work today

`UOFileManager` applies verdata patches for FileID `0, 2, 4, 12, 14, 16, 30, 32`
and explicitly drops the two animation ones:

```csharp
// src/ClassicUO.Assets/UOFileManager.cs
else if (vh.FileID != 5 && vh.FileID != 6)
{
    Log.Warn($"Unused verdata block\tFileID: {vh.FileID}\tBlockID: {vh.BlockID}");
}
```

FileID `6` is a patch of `anim.mul` (as `4` is a patch of `art.mul`), and
`AnimationsLoader.GetIndices()` always builds its index array from `anim*.idx`
and reads frames from `_files[fileIndex]`. So a block that has no bytes in
`anim.mul`/`anim.idx` — exactly the case a shard patches in — stays empty.

The intent to support this is already in the code: `AnimationDirection` in
`ClassicUO.Renderer` has

```csharp
public bool IsVerdata;   // src/ClassicUO.Renderer/Animations/AnimationDirection.cs
```

and `git grep IsVerdata` on current `main` returns **that single line** — the
field is never assigned and never read. This PR wires it up.

## What the patch does

`src/ClassicUO.Assets/AnimationsLoader.cs`

* Builds a lazy `Dictionary<uint, UOFileIndex5D>` of the FileID `6` patches
  (block index → position/length inside `verdata.mul`), invalidated in `Load()`.
* After the MUL indices are read in `GetIndices()`, applies the patches for
  `fileIndex == 0` (`anim.mul`). The `anim2..anim5` files have their own block
  numbering and are never patched through `verdata.mul`, so nothing else is
  touched.
* A patch with `length == 0` is the "deleted block" marker (Verdata entries can
  carry `offset = -1, length = 0`); such a block is turned into an empty one and
  `ReadMULAnimationFrames` returns no frames for it — the block is really gone.
* `ReadMULAnimationFrames(int fileIndex, AnimationDirection index, bool isVerdata = false)`
  reads from `FileManager.Verdata.File` when the block came from a patch, since
  `anim.mul` contains no bytes for it at all. The `fileIndex` bounds check is
  skipped in that case (the block does not live in `_files[fileIndex]`).
* `AnimationDirection` gained `public bool IsVerdata;`.

`src/ClassicUO.Renderer/Animations/Animation.cs`

* Copies `IsVerdata` from the loader index into `index.Groups[i].Direction[d]`.
* Passes `groupObj.Direction[dir].IsVerdata` to `ReadMULAnimationFrames`.

## Tested

* `dotnet build src/ClassicUO.Renderer/ClassicUO.Renderer.csproj -c Release`
  (.NET SDK 10.0.401, Linux, all submodules initialised) — build succeeded,
  0 errors, no new warnings from the touched files.
* Against a live shard (`middle-earth.ru`, 652 verdata patches of which 122 are
  FileID `6`): bodies `442`, `443`, `558` (custom mounts and a "wings" cloak
  whose `AnimID`s exist only in `verdata.mul`) now animate. Before the fix those
  blocks are missing in `anim.mul`, so the mobiles are drawn with no animation
  at all. Patch #35 of that file is a `length = 0` marker, which is the case
  handled above.
* Same fix has been running in the Android/Unity port of ClassicUO
  (MobileUO — https://github.com/MobileUO/MobileUO/pull/110) on the same shard.

## Notes / out of scope

* FileID `5` (patches of the `anim.idx` file itself) is still ignored, as
  before. I have not seen a shard use it; handling it would mean patching the
  `.idx` entries themselves rather than one animation block.
* UOP animations are not affected in any way.
* With no FileID `6` patches in `verdata.mul`, behaviour is unchanged (the
  patch map is empty and no index field is modified).


sharabdin1988:
Small logistic note: the `Build-Test` workflow on this PR is paused at `action_required` (first-time contributor to this repository), so it needs a maintainer to approve the run before CI can compile the change. The patch was verified locally against this exact `main` (`007ef8c`) with .NET SDK 10.0.401 — `dotnet build src/ClassicUO.Renderer/ClassicUO.Renderer.csproj -c Release` succeeds with 0 errors. Happy to adjust anything once the run goes through.

andreakarasho (approved):

